PR #5, opened 2019-11-13
Refactor Solver class and rewrite ConvTasNet/train.py accordingly

mpariente:
Refactor `Solver` to enable easy overwriting of sub-methods and give more freedom to the user.

### DIFF
--- a/asteroid/engine/solver.py
+++ b/asteroid/engine/solver.py
@@ -14,8 +14,12 @@ from ..utils import to_cuda
 _defaults = {
     # Training defaults
     'epoch': 1,
+    'min_prct_impv': 0.1,  # In percentage
     'half_lr': True,
+    'half_lr_patience': 3,
+    'lr_multiplier': 0.5,
     'early_stop': True,
+    'early_stop_patience': 8,
     'max_norm': 5.,
     # Model defaults
     'checkpoint': False,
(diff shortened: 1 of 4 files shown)


asteroid/engine/solver.py
```diff
@@ -52,37 +66,38 @@ def __init__(self, data, model, loss_class, optimizer, training_conf):
...
             start = time.time()
-            tr_avg_loss = self._run_one_epoch(epoch)
+            tr_avg_loss = self.run_one_epoch(epoch)
             print('-' * 85)
             print('Train Summary | End of Epoch {0} | Time {1:.2f}s | '
                   'Train Loss {2:.3f}'.format(epoch + 1, time.time() - start,
                                               tr_avg_loss))
             print('-' * 85)
-            if self.checkpoint:
+            if self.tr_conf['checkpoint']:
```
sunits: Should we use get function here? Just to be consistent

asteroid/engine/solver.py
```diff
@@ -95,69 +110,92 @@ def train(self):
...
                 self.early_val_no_impv = 0
             # Adjust learning rate (halving)
-            if self.half_lr and self.val_no_impv >= 3:
+            if self.tr_conf['half_lr'] and self.val_no_impv >= 3:
                 self.multiply_learning_rate(self.optimizer, 0.5)
                 self.ep_half_lr.append(epoch)
                 self.val_no_impv = 0
             # Early stopping
-            if self.early_val_no_impv >= 8 and self.early_stop:
+            if self.early_val_no_impv >= 8 and self.tr_conf['early_stop']:
```
sunits: Remove constants. Put it in default dictionary.

sunits: Same for 1.001 in line 106

asteroid/engine/solver.py
```diff
@@ -95,69 +110,92 @@ def train(self):
...
             self.tr_loss[epoch] = tr_avg_loss
             self.cv_loss[epoch] = val_loss
             if val_loss < self.best_val_loss:
                 self.best_val_loss = val_loss
-                file_path = os.path.join(self.save_folder, self.model_path)
-                self.save_model(file_path, epoch)
+                self.save_model(self.model_path, epoch)
 
-    def _run_one_epoch(self, epoch, validation=False):
+    def run_one_epoch(self, epoch, validation=False):
```
sunits: Let us discuss about the validation flag.

asteroid/engine/solver.py
```diff
@@ -95,69 +110,92 @@ def train(self):
...
+
+    def compute_loss(self, targets, est_targets, infos=None):
+        """ Computes the loss to be backpropred.
+        `infos` is a dictionary that is passed to the loss function.
+        """
+        return self.loss_class.compute(targets, est_targets, infos=infos)
+
+    def loss_backward_and_step(self, loss, validation=False):
+        """ Backpropagation of the loss and gradient descend step."""
+        if not validation and not torch.isnan(loss):
```
sunits: What happens if loss is NaN? May be do that check in compute_loss. 
How about making the some of these functions private? Should they be accessible outside the class?

asteroid/engine/solver.py
mpariente: `get` is called when the key might not be here. In this case, the key is here.

asteroid/engine/solver.py
mpariente: You're right, thanks. Callbacks will make this so much better

asteroid/engine/solver.py
mpariente: If loss is NaN, optimizer.step() makes all the weights NaN, and the network is unrecoverable.
The good approach will be to find the reason and prevent it. See [the anomaly detection tool from pytorch](https://pytorch.org/docs/master/autograd.html#anomaly-detection). I need to take time to set it up.

sunits: Put a Warning  for now.

PR #8, opened 2019-11-14
dprnn first commit

popcornell:
Added the DPRNN architecture. In Luo's TAC Conv2D is used for mask net. I have only tested this version which instead is more similar to plain ConvTasNet. 
Had to include also a 2D LayerNorm layer. 
Probably there is a better way and the vanilla _LayerNorm can be made compatible with 2D inputs

### DIFF
--- a/asteroid/masknn/blocks.py
+++ b/asteroid/masknn/blocks.py
@@ -277,9 +277,7 @@ class DPRNN(SubModule):
             net += [DPRNNBlock(bn_chan, hid_size, norm_type, bidirectional, rnn_type, num_layers, dropout)]
         self.net = nn.Sequential(*net)
 
-        # TODO Luo's code uses Conv2D in TAC not clear in [1] if it is also used Conv2D.
-        # This version with Conv1D however has been able to reach same results as in [1].
-        mask_conv = nn.Conv1d(bn_chan, n_src*out_chan, 1)
+        mask_conv = nn.Conv2d(bn_chan, n_src*out_chan, 1)
         self.mask_net = nn.Sequential(nn.PReLU(), mask_conv)
 
         # Get activation function. For softmax, feed the source dimension.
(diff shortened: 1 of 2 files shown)


mpariente:
You can now use `GlobLN` for 4D inputs thanks to [this commit](https://github.com/mpariente/AsSteroid/commit/6fba41c7c1e1f35fb7fee5e1d17cd7f2882e1ec5).

asteroid/masknn/blocks.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, in_chan, n_src, out_chan=None, n_blocks=8, n_repeats=3,
 
         layer_norm = norms.get(norm_type)(in_chan)
         bottleneck_conv = nn.Conv1d(in_chan, bn_chan, 1)
+        self.bottleneck = nn.Sequential(layer_norm, bottleneck_conv)
```
popcornell: I moved the self.bottleneck from the end here as it makes more sense to have it here.

popcornell:
Sorry forgot to do rebase before push. 
I can make a recipe for the DPRNN but right now it is not that clear to me what is the recipe template. 
Is the ConvTasNet recipe ready ? I am confused about the arguments passed by the bash script and the other fetched form the conf.yaml file.

asteroid/masknn/blocks.py
mpariente: That's clearer this way, you're right. Thank you

asteroid/masknn/blocks.py
```diff
@@ -140,3 +141,196 @@ def get_config(self):
...
+        self.n_repeats = n_repeats
+        self.n_src = n_src
+        self.norm_type = norm_type
+        self.mask_act = mask_act
+        self.bidirectional = bidirectional
+        self.rnn_type = rnn_type
+        self.num_layers = num_layers
+        self.dropout = dropout
+
+        layer_norm = norms.get('cLN')(in_chan)
```
mpariente: Shouldn't this be `layer_norm = norms.get(norm_type)(in_chan)` ?

mpariente:
> Is the ConvTasNet recipe ready ? I am confused about the arguments passed by the bash script and the other fetched form the conf.yaml file.

It's almost ready, but still there are some things to test. I should write a doc about the way it works but basically : 
- The YAML configuration is read into a dictionary in train.py
- train.py creates a parser containing all the keys which were present in the YAML file (and other if added manually with `parser.add_argument`)
- And all these arguments can be parsed from the command line.

So, anything in the YAML, or manually defined in train.py can be used in the bash script and if it is not passed by the bash scripts, it defaults to the value in conf.yml.

mpariente:
Thanks again for the valuable contribution ! 
Please next time, think about `git pull --rebase`, that'll make a cleaner history :-)

popcornell:
Ok i see you changed the run.sh 3 days ago now it makes more sense to me. Before it did stop at training unless you edited the right path in the conf.yaml file. 
I am also glad you adopted finally the yaml dictionary approach too ;).

mpariente:
Hey Samu, does the `DPRNN` runs for `n_src != 1` ? I think there is a simple reshaping problem at [this line](https://github.com/mpariente/AsSteroid/blob/master/asteroid/masknn/blocks.py#L344), am I mistaken? 

I rewriting the `__init__` to have more keyword arguments, 8 positional argument is too much actually, I should have spotted that in the first place.

Can you check this please? 
```python
import torch
from asteroid.masknn.blocks import DPRNN
model = DPRNN(128, 128, 128, 128, chunk_size=100,
              hop_size=50, n_repears=6, n_src=2)
u = torch.randn(8, 128, 1000)
v = model(u)
```
Adding the source dimension in the reshaping fixes the problem but I want to make sure that it looks ok to you before PR.
Thx

popcornell:
Good catch sorry !!! Yes it is broken it should be:


```python
to_unfold = n_filters * self.chunk_size
        output = fold(output.reshape(batch*self.n_src, to_unfold, n_chunks),
                      (n_frames, 1), kernel_size=(self.chunk_size, 1),
                      padding=(self.chunk_size, 0),
                      stride=(self.hop_size, 1))
```

We should add unittest indeed. Do you have any resource/other repo were we I can find how to set up the tests ? I am not so practical with unittest actually but I am eager to learn.

Issue #14, opened 2019-11-22
Use pytorch-lightning for training? 


mpariente:
This would solve #6 and #3 . 
- Check flexibility with respect to : 
  - Model saving and loading (need to instantiate or not?)
  - Optimizers and callbacks
  - config.yml files in egs
  - Loss classes to handle PIT nicely
  - Unified dataset and dataloader structure? 

Any feedback on that from anybody?

sunits:
Any estimate on the changes required?

mpariente:
Surprisingly not much ! 
Example training scripts for the ConvTasnet recipe can be found [here](https://github.com/mpariente/AsSteroid/tree/light_ignite_trial/egs/wham/ConvTasNet).
I tried `pytorch-lightning` and `ignite`.
`ignite` is soon going to be 2 years old and `pytorch-lightning` will soon be a year old. Even so, the interface of `pytorch-lightning` is already a bit better (even if less stable and will need refactoring IMO) and more ambitious.
At this stage  I would vote for `pytorch-lightning`. 
Any thoughts  @popcornell @sunits @hbredin ?

popcornell:
I am not familiar with either of two but I like the `pytorch-lightning` interface more.
I think is more clear. Plus its support for multi-node training seems to be better than the one in `ignite`.

hbredin:
There is also one more candidate: https://docs.fast.ai/train.html
However, I've never used any of them.

mpariente:
It's less easy to interface to native pytorch with `fastai`, so that's a little bit less flexible. Plus the dependencies are also quite big. On the other side, it's much more stable and very used, I'll look into it more.

popcornell:
It seems to me that the only advantage of `fast.ai` over the other two (aside of much more wider community and better support) is that it supports some training scheduling tricks which I have to say are quite effective. 
But these can be easily ported and implemented via callbacks. 
`fastai` seems also much more focused on CV tasks so much of the features it offers will not be used.

mpariente:
Solved by #19

PR #15, opened 2019-11-23
fixed loading bug

popcornell:
There is a bug 🐛 that prevents to resume training from a checkpoint. 
Might as well being useless this PR if we decide to move to either to `ignite` or `pytorch-lightning` as loading will be handled by those,
see [#14](https://github.com/mpariente/AsSteroid/issues/14#issue-527157100).

### DIFF
--- a/asteroid/engine/solver.py
+++ b/asteroid/engine/solver.py
@@ -48,14 +48,14 @@ class Solver(object):
 
     """
     def __init__(self, loaders, model, loss_class, optimizer, model_path,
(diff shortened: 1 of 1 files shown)


mpariente:
Actually, we will still build the `pack` in `asteroid` and save what we want in it (See [here](https://github.com/mpariente/AsSteroid/blob/light_ignite_trial/egs/wham/ConvTasNet/lightning_train.py#L83)). We will have to handle the loading as well so that's actually an important catch, thanks !

PR #24, opened 2019-12-26
New API for PIT and non PIT losses 

mpariente:
I'll take more time to describe what it does soon but this is ready for merging. 
Anyone wanting to review it is welcome !

Cheers !

### DIFF
--- a/asteroid/engine/system.py
+++ b/asteroid/engine/system.py
@@ -9,32 +9,33 @@ import pytorch_lightning as pl
 
 class System(pl.LightningModule):
     """ Base class for deep learning systems.
-    Subclass of pytorch_lightning.LightningModule.
-    Contains a model, an optimizer, a loss_class and training and validation
-    loaders and learning rate scheduler.
+    Contains a model, an optimizer, a loss function, training and validation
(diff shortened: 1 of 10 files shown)


popcornell:
You added the `nosrc` functions for better efficiency in Speech Enhancement (1 speaker) applications ? 
I think we can merge and then i can pull request changes regarding sd-sdr later.

mpariente:
> You added the `nosrc` functions for better efficiency in Speech Enhancement (1 speaker) applications ?

No actually, there are three ways to compute permutation invariant losses and the `nosrc` functions can be used for multisource separation as well. It's  a bit explained in the docstrings of the `PITLossWrapper` but I have to make a notebook to explain this clearly. 

`nosrc` functions expects signals without source dimensions, for example with shape `(batch, time)`.

popcornell:
:+1: gotcha now. 
So `wo_src` and `w_src` are added for easier integration with loss functions code which is not pairwise right ? 

BTW The tests are a beautiful addition !

mpariente:
Still didn't take time to make the notebook about the `PITLossWrapper` but I will eventually in the next few weeks.

PR #25, opened 2020-01-30
added sd-sdr

popcornell:
Lots of boiler-plate code, probably a better way is to use an if onto si-sdr for distinguishing between SI-SDR and SD-SDR. Like instead of scale_invariant=True --> scale_dependent=False 
What do you think ? 

pull requesting only to foster discussion


mpariente:
Thanks for the PR !


> Lots of boiler-plate code, probably a better way is to use an if onto si-sdr for distinguishing between SI-SDR and SD-SDR. Like instead of scale_invariant=True --> scale_dependent=False
> What do you think ?

It's a good question, but I think if we do that, we will need something like
```python
from functools import partial
 from asteroid.losses import PITLossWrapper
loss_func = PITLossWrapper(partial(pairwise_neg_sisdr,
                                   scale_dependent=True),
                           mode='pairwise')
```
This is not so nice.. 
Or, we will need to define in the training script something like 
```python
def my_loss(a, b):
    return pairwise_neg_sisdr(a, b, scale_dependent=True)
loss_func = PITLossWrapper(my_loss, mode="pairwise")
```

Or maybe, I should add a `loss_kwargs` argument into `PITLossWrapper`. 
What do you think about that? Maybe it will make things simpler.

I do think it's nicer to have a separate function for si-sdr.. 

Do you think you can work on the tests? And import the new functions in `losses/__init__.py`?

popcornell:
But maybe we can make one common function which serves as SI-SDR, SD-SDR and SNR. 
Actually three functions one for pairwise one for nopit and one for nosrc. 

Then we can make wrappers to those functions to handle each loss function. 
In this way, we have little duplication for code and we keep the user interface as simple as possible: 

```python
def pairwise_si_sdr(a, b):
    return pairwise_sdr_base(a, b, type="sisdr" )
```

mpariente:
Sounds great ! 
Do you want to do it ? :smiley:

popcornell:
Ok I will give it a try this weekend. 
Does it make sense for you the zero_mean argument also ? 

```python
if zero_mean:
        mean_source = torch.mean(targets, dim=2, keepdim=True)
        mean_estimate = torch.mean(est_targets, dim=2, keepdim=True)
        targets = targets - mean_source
        est_targets = est_targets - mean_estimate
        # Step 2. Pair-wise SI-SDR. (Reshape to use broadcast)
```

mpariente:
> Ok I will give it a try this weekend.

Cool thanks !

> Does it make sense for you the zero_mean argument also ?

Yes it does, we can keep it and keep the default to `True`

popcornell:
Closing, added a new branch on master [pit_losses](https://github.com/mpariente/AsSteroid/tree/pit_losses) with new losses API. Still needs tests.

PR #26, opened 2020-01-31
fixed bug in DPRNN which occurs if in_chan != bn_chan

popcornell:


mpariente:
Thanks for catching this. 
Adding unit tests for the architecture is on the todo list !

PR #27, opened 2020-02-01
[src] DPRNN signature change

mpariente:
- Fixes shape mismatch from #8
- `DPRNN` is globally importable from `masknn` now.
- Change signature in `DPRNN` to have only two positional arguments (same as `TDConvNet`).

`DPRNN` can simply be called like this now
```python
import torch
from asteroid.masknn import DPRNN
in_channels, n_src = 128, 2
model = DPRNN(in_channels, n_src)
inp = torch.randn(1, 128, 500)
out = model(inp)
```

### DIFF
--- a/asteroid/masknn/__init__.py
+++ b/asteroid/masknn/__init__.py
@@ -1,1 +1,2 @@
 from .blocks import TDConvNet
+from .blocks import DPRNN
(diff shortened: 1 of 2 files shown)


popcornell:
Looks good ! Thanks for fixing it.

PR #34, opened 2020-02-13
Deep clustering  Implementation

sunits:
-> deep clustering loss
    -> Chimera ++ network
    -> dataloader for WSJ 2 and 3 mix

Things pending:
    -> Complete recipe with speech separation metrics
    -> PIT loss for the mask estimation head


egs/wsj0-mix/DeepClustering/train.py
```diff
@@ -0,0 +1,123 @@
...
+parser.add_argument('--exp_dir', default='exp/tmp',
+                    help='Full path to save best validation model')
+
+class DcSystem(System):
+    """ System for deep clustering implementation. 
+        Overwrites the common_step.
+    """
+    def __init__(self, model, optimizer, loss_func, train_loader,
+                 val_loader=None, scheduler=None, config=None):
+        System.__init__(self, model, optimizer, loss_func, train_loader,
```
mpariente: Can we use `super` here please? With python3 style `super`, no need to name the `System` class so it's easier to refactor if needed.

egs/wsj0-mix/DeepClustering/train.py
```diff
@@ -0,0 +1,123 @@
...
+        new_sources = sources.view(-1, n_sample).unsqueeze(1)
+        src_mag_spec = take_mag(self.enc(new_sources))
+        fft_dim = src_mag_spec.shape[1]
+        src_mag_spec = src_mag_spec.view(n_batch, n_src, fft_dim, -1)
+        src_sum = src_mag_spec.sum(1).unsqueeze(1) + EPS
+        real_mask = src_mag_spec/src_sum
+        # Get the src idx having the maximum energy
+        binary_mask = real_mask.argmax(1)
+        infos = {}
+        infos['real_mask'] = real_mask
```
mpariente: This is now obsolete I guess.

asteroid/data/wsj0_mix.py
```diff
@@ -0,0 +1,144 @@
...
+        print("{}% file dropped".format(100*(1-self.len/len(id_list))))
+
+    def identity(self, *kargs):
+        return kargs
+
+    def __len__(self):
+        return self.len
+
+    def __getitem__(self, idx):
+        raise NotImplementedError
```
mpariente: Already inherited from `Dataset`.

asteroid/data/wav.py
```diff
@@ -79,6 +82,47 @@ def data(self):
             self.__wav_data = wav_data
```
mpariente: I know I should have checked sooner but why double underscore here? Instead of single? Double underscore have a special function in Python and I don't see the use of it. Can you explain please?

asteroid/data/wsj0_mix.py
```diff
@@ -0,0 +1,144 @@
+"""
+wsj-*mix Dataset class: A dataset interface to process wsj-{2,3}mix dataset
+@author : Sunit Sivasankaran, Inria-Nancy
+"""
+
```
mpariente: No more headers if possible. Git will handle ownership

egs/wsj0-mix/DeepClustering/model.py
```diff
@@ -0,0 +1,123 @@
...
+        out = self.dropout(out)
+        projection = self.embedding_layer(out)
+        projection = self.non_linearity(projection)
+        projection = projection.view(batches, -1, self.embedding_dim)
+        proj_norm = th.norm(projection, p=2, dim=-1, keepdim=True) + \
+                th.finfo(th.float32).eps
+        projection_final =  projection/proj_norm
+        mask_out = self.mask_layer(out)
+        mask_out = mask_out.view(batches, self.n_src, self.input_dim, seq_cnt)
+        return projection_final, mask_out
```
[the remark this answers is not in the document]
popcornell: May be better to organize the .masknn module in a more hierarchical way and not put everything into blocks.py

egs/wsj0-mix/DeepClustering/train.py
sunits: Agree

asteroid/data/wsj0_mix.py
sunits: Yup. Removing it.

asteroid/data/wav.py
sunits: Double underscore marks the variable as   "protected" and makes it harder to change it. Make sense to keep it.

asteroid/data/wsj0_mix.py
sunits: Agreed.

egs/wsj0-mix/DeepClustering/model.py
```diff
@@ -0,0 +1,123 @@
...
+                                         dropout=dropout,
+                                         batch_first=True,
+                                         bidirectional=bool(bidirectional))
+        self.rnn_type = rnn_type
+
+    def forward(self, inp):
+        """ Input shape [batch, seq, feats] """
+        output = inp
+        rnn_output, _ = self.rnn(output)
+        return rnn_output
```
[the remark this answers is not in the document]
sunits: It is the same. Removing it from the file and using the one from the core.

egs/wsj0-mix/DeepClustering/model.py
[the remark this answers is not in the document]
mpariente: I agree, it's not scalable in this way. You can always propose changes in a PR if you feel like ;-)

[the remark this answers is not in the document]
sunits: It perhaps makes sense to differentiate between blocks and models. For example SingleRNN should be inside blocks since it is used across different models and Chimerapp, DPRNN et al are models which are unlikely to be used across difference egs. Should you put models in the core or inside egs is debatable.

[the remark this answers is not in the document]
mpariente: The models can be used in recipes on different datasets, they have to be globally importable, they belong to core IMO.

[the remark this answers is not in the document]
popcornell: Okay to leave at least some models on masknn but to differentiate models and blocks like SingleRNN I think is reasonable. we can have masknn.blocks module folder and then in masknn all the models for example.

mpariente:
I guess you rebased and force pushed, we cannot see clearly what belongs to the PR. Let's open a new one, only with commits related to this PR. 

There is no need to rebase master if their are no conflicts and if there are conflicts, merging from master is a better a option.

mpariente:
Merging it, thanks !
We still need the recipe, the evaluation script and a readme presenting the dataset

PR #36, opened 2020-02-17
DPRNN recipe

popcornell:
Added recipe for DPRNN (adapted the Conv-TasNet one). 
Also added relative path for env installation of requirements. 
The env setup still fails because of pb-bss but we already knew that.

### DIFF
--- a/egs/wham/ConvTasNet/utils/prepare_python_env.sh
+++ b/egs/wham/ConvTasNet/utils/prepare_python_env.sh
@@ -2,7 +2,7 @@
 # Usage ./utils/install_env.sh --install_dir A --asteroid_root B --pip_requires C
 install_dir=~
 asteroid_root=../../../../
-pip_requires=  # Expects a requirement.txt
+pip_requires=../../../requirements.txt  # Expects a requirement.txt
 
 . utils/parse_options.sh || exit 1
 
--- a/egs/wham/DPRNN/conf.yml
+++ b/egs/wham/DPRNN/conf.yml
@@ -1,0 +1,40 @@
+# Filterbank config
+filterbank:
+  n_filters: 64
+  kernel_size: 2
+  stride: 1
+# Network config
+masknet:
+  in_chan: 64
+  n_src: 2
+  out_chan: 64
+  bn_chan: 128
+  hid_size: 128
+  chunk_size: 250
+  hop_size: 125
+  n_repeats: 6
+  mask_act: 'sigmoid'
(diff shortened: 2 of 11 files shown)


popcornell:
BTW I do not like that some hyperparameters (like batch size) are both on` conf.yml` and on `run.sh` script as argument. 
It may happen you change them on `conf.yml` but they don't change cos they are also on the `run.sh` (happened to me now cos it is late and i keep forgetting this). 
I like that this makes the framework flexible but it can be confusing at first and recipes should be as simple as possible IMO.

popcornell:
Yeah I should probably go to bed wrongly clicked close and comment

mpariente:
Thank you very much for the PR, I'll get to the review asap

mpariente:
> BTW I do not like that some hyperparameters (like batch size) are both on` conf.yml` and on `run.sh` script as argument.
> It may happen you change them on `conf.yml` but they don't change cos they are also on the `run.sh` (happened to me now cos it is late and i keep forgetting this).
> I like that this makes the framework flexible but it can be confusing at first and recipes should be as simple as possible IMO.

I agree with that.
Is it different than default values in the parser definition overwritten by the bash file / command line? Like in the *original* conv-tasnet repo?

I've been thinking about a bash function that retrieves the value for a given key in the yaml file and you can use this value if none were overriden in the bash file / from the command line. What would you think about this? 
While I like the idea, I think it makes the whole things to complicated and maybe just moving the yaml file to `local` would be enough. I seriously don't know, any thoughts on this are welcome. Thanks for pointing it out !

popcornell:
I am for retaining both but for recipes use either only the yaml or the bash. 
In this way a new user can tinker with the hyper-parameters without some "bad surprises" (you reduce batch size on the yaml but you don't know it will be overwritten in the run.sh and after 2 days of conv tasnet running you realize it has been overwritten !).

mpariente:
The way I see it :
- In the yaml file, you define all the arguments your experiment (train.py)
expects, with sensible default values.
- In the bash file, you parse and pass to train.py all the arguments you
want to experiment with.

This means that you should almost never change the conf.yml file. If you
want to experiment with the batch size, it has to be an entry in run.sh and
so on..

With this in mind, I think moving conf.yml into the local folder makes even
more sense.

Any thoughts on that?

popcornell:
Ok now is more clear and makes more sense. 
Yes moving the conf.yaml in the local folder could also prevent some confusion.

PR #37, opened 2020-03-01
Evaluation for the DNS Challenge

mpariente:
This PR adds evaluation scripts and denoising scripts to the baseline recipe. 

There are still a few TODOs, especially things that could be more general and that got here first just by convenience. 

Also `preprocess_dns.py` should be modified to follow [this](https://github.com/mpariente/asteroid/compare/dns_eval?expand=1#diff-54fd0e82eb83ad730e6508404a61a0c4R78) preprocessing which is 3x faster.

### DIFF
--- a/asteroid/torch_utils.py
+++ b/asteroid/torch_utils.py
@@ -1,0 +1,20 @@
+import torch
+from torch import nn
+
+
+def pad_x_to_y(x, y, axis=-1):
+    """  Pad first argument to have same size as second argument
+
+    Args:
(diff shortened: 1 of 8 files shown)


mpariente:
Fixed all the TODOs and changed preprocessing to be 3x faster. 
Things to do (in another PR): 
- Use the padding from utils in the other recipes. 
- Improve (a lot) the padding function.
- Evaluate microsoft's baseline with the same metrics.

Merging.

PR #39, opened 2020-03-04
DPRNN bug

popcornell:


mpariente:
Rebased and force pushed to see only the relevant commit in the PR, thanks for that !

PR #42, opened 2020-03-06
Add support for multichannel convolutions

mpariente:
Need feedback on output shapes @popcornell @sunits @JorisCos please !
### What is this PR about 
This adds support to multichannel 1D convolutions and transposed convolutions. It's useful for multichannel or multi-source STFT and iSTFT in particular. 
We also add support for convolution of 1D and 2D signals. Previously only 3D signals of shape `(batch, 1, time)` were supported in `Encoder.forward`. No need for batch or empty channel dimension anymore (but this is still supported) !

```python
import torch
from asteroid.filterbanks import STFTFB, Encoder, Decoder

batch=3
encoder = Encoder(STFTFB(512, 512, 256))
decoder = Decoder(STFTFB(512, 512, 256))

# 1D case : this now works
inp = torch.randn(32000,)
stft = encoder(inp)  # (freq, conv_time)
out =  decoder(stft)  # (1, time)

# 2D case : this also works
inp = torch.randn(batch, 32000)
stft = encoder(inp)   # (batch, freq, conv_time)
out =  decoder(stft)  # (batch, 1, time)

# 3D case with channels : this also works
inp = torch.randn(batch, 3, 32000)
stft = encoder(inp)   # (batch, 3, freq, conv_time)
out =  decoder(stft)  # (batch, 3, time)
```

This is achieved with grouping the non-convoluted dimensions into the batch so we keep `torch`'s efficiency.
All the expected behavior are unit tested.

### Why this choice for the output shapes?
The STFT gives something like this 
```
(time) --> (freq, conv_time)
(1, time) --> (1, freq, conv_time)  # We could squeeze here but I think that's a bad idea.
(1, 1, time) --> (1, freq, conv_time)  # Same as a regular Conv1D
(2, time) --> (2, freq, conv_time)
(2, 1, time) --> (2, freq, conv_time)  # Same as a regular Conv1D
(2, 2, time) --> (2, 2, freq, conv_time)  # Multichannel starts here
(18, 32, 1) --> (18, 32, 1, freq, conv_time)
```
Because `(batch, time)` and `(batch, 1, time)` match to the same stft shape `(batch, freq, conv_time)`, the `Decoder` cannot know from which shape it comes from. What do you think about this? Which choice for the `Decoder`'s output shape? Should we add an `output_dim` argument to the decoder?  

### What remains to be done 
- Take a decision for the iSTFT.
- Update the docs to reflect on the changes (comes on the next PR on filterbanks, with some other cleanups)

### DIFF
--- a/asteroid/filterbanks/enc_dec.py
+++ b/asteroid/filterbanks/enc_dec.py
@@ -1,3 +1,4 @@
+import warnings
 import torch
 from torch import nn
 from torch.nn import functional as F
@@ -119,13 +120,15 @@ class Encoder(_EncDec):
                 -  ``'comp'`` or ``'complex'`` corresponds to a complex mask:
                    the input and the mask are point-wise multiplied in the
                    complex sense.
+        as_conv1d (bool): Doc to come.
 
     """
     def __init__(self, filterbank, inp_mode='reim', mask_mode='reim',
-                 is_pinv=False):
+                 is_pinv=False, as_conv1d=True):
         super(Encoder, self).__init__(filterbank, is_pinv=is_pinv)
         self.inp_mode = inp_mode
         self.mask_mode = mask_mode
+        self.as_conv1d = as_conv1d
 
         self.inp_func, self.in_chan_mul = _inputs[self.inp_mode]
         self.mask_func, self.out_chan_mul = _masks[self.mask_mode]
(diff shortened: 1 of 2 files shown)


mpariente:
What we propose (thanks @sunits) : 
We can add a `as_conv1d=True` argument to the `Encoder`, we will have 
```
Encoder
(time) --> (freq, conv_time)
if as_conv_1D (default):
    (1, 1, time) --> (1, freq, conv_time)  # Same as a regular Conv1D
    (1, 2, time) --> (1, 2, freq, conv_time)  # Same as a regular Conv1D
    (2, 1, time) --> (2, freq, conv_time)
else:
    (1, 1, time) --> (1, 1 freq, conv_time)  # Same as a regular Conv1D
    (2, 1, time) --> (2, 1, freq, conv_time)
(2, 2, time) --> (2, 2, freq, conv_time)  # Multichannel starts here
```
and
```
Decoder
(freq, conv_time) --> (time)
(B, freq, conv_time) --> (B, 1, time)
(batch, chan, freq, conv_time) --> (batch, chan, src, time)
```
Additionally, we will add a warning that 2D tensors will be mapped to 3D in the STFT, but iSTFT will map them back to 3D `(B, T) --> (B, F, CT) --> (B, 1, T)`

popcornell:
Another option is to do not have Encoder mapping to the same shape for both `(batch, time)` and `(batch, 1, time)`. But keeping it consistent: `(batch, 1, time) --> (batch, 1, freq, frames)` But I guess this will require some changes in all recipes. The option proposed by @sunits  is a good compromise.

mpariente:
> Another option is to do not have Encoder mapping to the same shape for both `(batch, time)` and `(batch, 1, time)`. But keeping it consistent: `(batch, 1, time) --> (batch, 1, freq, frames)` But I guess this will require some changes in all recipes. The option proposed by @sunits is a good compromise.

The `as_conv1d` argument in the Encoder, will do that to be able to keep consistency if we want to play with number of sources. But the thought was that, by default, Encoder should match the behaviour of `nn.Conv1D` so that it's not confusing for new users.

PR #45, opened 2020-03-06
Fix evaluation bugs

mpariente:
Models where not loaded correctly because of `strict=False` and keys mismatch, my bad. 
- Model loading from a trained `System` is now easy with `torch_utils.load_state_dict_in`. This tries loading the state_dict as is, if it fails, it removes the string subparts corresponding to the upper module. 
Please read the docs and the unit tests, this should be quite clear and if it's not, tell me. 
- Evaluation scripts are updated accordingly to load model correctly 
- load_best_model function got to model.py file in the recipes because it makes more sense IMO
- Also added a `are_models_equal` in torch_utils.py to check if two models have the same weights or not (also unit tested).

### DIFF
--- a/asteroid/torch_utils.py
+++ b/asteroid/torch_utils.py
@@ -1,5 +1,6 @@
 import torch
 from torch import nn
+from collections import OrderedDict
 
 
 def pad_x_to_y(x, y, axis=-1):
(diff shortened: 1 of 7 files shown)


popcornell:
I tested it right now with DPRNN and it works like a charm thanks !!

PR #48, opened 2020-03-08
Filterbank small refactor

mpariente:
I've been wanting to do this refactor for a while now. 

- `inputs_and_masks` didn't contain only input and mask transforms, this now goes to transforms (Future warning for those who still use `inputs_and_masks`)
- Encoder doesn't have `post_process_inputs` and `apply_mask` methods anymore. The methods where not useful and things are much clearer this way IMO.

Sorry but rebase master seems mandatory for most branches and forks after merging this PR.. @popcornell @sunits @JorisCos


popcornell:
So basically masking should be handled in each model/recipe by hand now using `transforms.py` e.g. `take_mag` ?

mpariente:
> So basically masking should be handled in each model/recipe by hand now using `transforms.py` e.g. `take_mag` ?

Correct. 
You canimagine several post_processing on the same TF representation, which wouldn't be handled by the current `Encoder`. 
I think using `take_mag` is straightforward and clear, there is no need to attach it nowhere.

sunits:
We will have to start thinking of writing test cases to the recipes as well. It is quite simple to break things when we modify the core part of the asteroid package. One way to do that would be to create a small set of dummy  wav files using np.random and run the recipe with it!

mpariente:
Couldn't agree more about the testing of the recipes, I tested them by hand now and it's definitely not scalable.
We shouldn't not lengthen CI builds too much though, do you want to give it a try?

sunits:
Yup. Will have to create dataset specific folder structure and  save some wav files. Do you know of any specific limits (memory or otherwise) in CI?

mpariente:
7.5GB memory apparently ([source](https://docs.travis-ci.com/user/reference/overview/)).
How about symlinking the recipes into the test folder and testing model definition, foward / loss / backward with a dummy dataset, and a most importantly a very small model?

PR #50, opened 2020-03-08
Renaming in PITLossWrapper (backward compatible)

mpariente:
We've been thinking about the names in the `PITLossWrapper`for the `mode`. 

We change the argument name from `mode` (uninformative) to `pit_from` and the provided string tells what the `PITLossWrapper` expects (from what it computes the PIT losses). Valid values are 
- `'pw_mtx'` for pairwise matrix, the loss_func returns the pairwise matrix firectly
-  `'pw_pt'` for pairwise point, the loss func returns one point of the pairwise matrix, between one source and its estimate.
-  `'perm_avg'` for permutation average, the loss func returns the loss average along a given sources/estimates average. 

The tests are passing with the old mode but with `DeprecationWarning`.

### Remains to be done
- Adjust the tests
- Update the notebook

Need feedback please @popcornell @sunits @JorisCos @mhu-coder 
More importantly, find out names for the corresponding loss functions. For the one computing the pairwise losses, the name will stay `pairwise_*_loss` but what about the others? How about `est2ref*` and `ests2refs` which has two advantages : it clarifies the function behavior (intended) but also the order of the arguments. Any suggestions on this will be appreciated ! The current names are just bad...


mpariente:
Opted for `Pairwise*`, `SingleSrc*` and `MultiSrc*` for the loss function names.

PR #52, opened 2020-03-10
Fix for GPU usage

JorisCos:
### Content
Fix the behavior of the `SingleSrcMultiScaleSpectral` loss function when dealing with data coming from a GPU.

### What is wrong now

You need to specify:
`loss_func=PITLossWrapper(SingleSrcMultiScaleSpectral(),pit_from='pw_pt').to(device)`
with `device` a `cuda.device` object matching your targets and estimates `device`. If you don't, encoders in the loss function will be generated with `device= torch.device('cpu')` and your code will break.

### Intended behavior

You should just call `PITLossWrapper(SingleSrcMultiScaleSpectral(),pit_from='pw_pt')`
and the function will choose the `device` according to your targets and estimates.

### DIFF
--- a/asteroid/losses/multi_scale_spectral.py
+++ b/asteroid/losses/multi_scale_spectral.py
@@ -72,6 +72,7 @@ class SingleSrcMultiScaleSpectral(_Loss):
             in range(len(self.n_filters)))
 
     def forward(self, est_target, target):
+        self.encoders.to(est_target.device)
         batch_size = est_target.shape[0]
         est_target = est_target.unsqueeze(1)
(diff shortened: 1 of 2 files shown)


asteroid/losses/multi_scale_spectral.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, n_filters=None, windows_size=None,
             in range(len(self.n_filters)))
 
     def forward(self, est_target, target):
+        self.encoders.to(est_target.device)
```
mpariente: This is not needed anymore right?

PR #60, opened 2020-03-14
added aliases for Pytorch BatchNorm1D and 2D

popcornell:
as the title... felt we should add this in order to have existing blocks support batchnorm.

### DIFF
--- a/asteroid/masknn/norms.py
+++ b/asteroid/masknn/norms.py
@@ -80,9 +80,8 @@ class CumLN(_LayerNorm):
 
 class BatchNorm(_BatchNorm):
     """Wrapper class for pytorch BatchNorm1D and BatchNorm2D"""
-
     def _check_input_dim(self, input):
-        if input.dim() < 3 or input.dim() > 4:
+        if input.dim() < 2 or input.dim() > 4:
             raise ValueError('expected 4D or 3D input (got {}D input)'
                              .format(input.dim()))
 
--- a/asteroid/masknn/norms.py
+++ b/asteroid/masknn/norms.py
@@ -1,8 +1,6 @@
 import torch
 from torch import nn
-from torch.nn.parameter import Parameter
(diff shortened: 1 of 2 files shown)


mpariente:
Oh cool, sure! 
Can you just add them in the test file please, as strings?

mpariente:
The test is failing on BatchNorm2d because it expects 4D inputs, not 3D. 
Making a separate test for 4D inputs would be useful, even for the other classes

popcornell:
I modified _BatchNorm base class in order to have just one module to handle 1D and 2D. 
Actually it would have been easier (less duplicate code) with newer torch cos they have a _NormBase class now.

asteroid/masknn/norms.py
```diff
@@ -76,10 +78,19 @@ def forward(self, x):
         return self.apply_gain_and_bias((x - cum_mean) / (cum_var + EPS).sqrt())
 
 
+class BatchNorm(_BatchNorm):
+    """Wrapper class for pytorch BatchNorm1D and BatchNorm2D"""
+
+    def _check_input_dim(self, input):
+        if input.dim() < 3 or input.dim() > 4:
```
mpariente: BatchNorm1D supports 2D tensors as well. Can we have `input.dim()<2` instead?

mpariente:
So `BatchNorm2D` expects the channel to be in the second dim. This is not compatible with the (batch, n_src, chan, time) format that we have sometime. 

For now, this wrapper is fine. If we have the use case, we might make a `BatchNorm1d` wrapper that supports any input dimension, like in the STFT. Let's just keep it in mind.

 Merging as soon as the shape check is resolved.

popcornell:
Agree. Now the only use case where we could use the 2D batchnorm is in fact the DPRNN.

mpariente:
Great thanks !

PR #63, opened 2020-03-17
A new dataset and a recipe 

JorisCos:
### **Content**
Add a new dataset derived from LibriSpeech ASR corpus http://www.openslr.org/12/
The recipe uses ConvTasNet with the MSS loss as a loss function


egs/libri_2_mix/ConvTasNet/run.sh
```diff
@@ -0,0 +1,116 @@
...
+set -e  # Exit on error
+# Main storage directory. You'll need disk space to dump the raw LibriSpeech flac files
+# and generate the sources and mixtures
+storage_dir=/srv/storage/talc@talc-data.nancy/multispeech/calcul/users/jcosentino/DATA/
+
+# If you haven't downloaded and extracted the LibriSpeech dataset start from stage 1
+
+# If you already have dowloaded and extracted LibriSpeech,
+# specify the path to the directory here and start from stage 2
+LibriSpeech_dir=${storage_dir}
```
mpariente: Create a Librispeech directory inside storage_dir (at stage 1), do all the downloads there. LibriSpeech_dir would default to ${storage_dir}/LibriSpeech. 
Also, no caps in variables please : LibriSpeech_dir --> librispeech_dir

egs/libri_2_mix/ConvTasNet/local/prepare_data.sh
```diff
@@ -0,0 +1,41 @@
...
+
+. ../utils/parse_options.sh
+
+## Download WHAM noises
+mkdir -p $out_dir
+mkdir -p $storage_dir
+echo "Download LibiSpeech_train_100_hours_clean dataset into $storage_dir"
+# If downloading stalls for more than 20s, relaunch from previous state.
+wget -c --tries=0 --read-timeout=20 http://www.openslr.org/resources/12/train-clean-360.tar.gz -P $storage_dir
+tar -xzvf $storage_dir/train-clean-100.tar.gz -C $storage_dir
```
mpariente: Download 360 and untar 100, fix to 360.

egs/libri_2_mix/ConvTasNet/local/create_dataset_from_metadata.py
```diff
@@ -0,0 +1,240 @@
...
+import pyloudnorm as pyln
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--librispeech_root_path', type=str, default=None,
+                    help='Path to librispeech root directory')
+parser.add_argument('--dataset_root_path', type=str, default=None,
+                    help='Path to the desired dataset root directory')
+parser.add_argument('--freqs', nargs='+', default=['8K'],
+                    help='--freqs 16K 8K will create 2 directories 8K '
+                         'and 16K')
```
mpariente: Make the argument case-insensitive, maybe in `main()`, call upper() or lower().

asteroid/data/librimix_dataset.py
```diff
@@ -0,0 +1,56 @@
...
+        for i in range(self.n_src):
+            source_path = row[f'source_{i+1}_path']
+            s, _ = sf.read(source_path, dtype='float32')
+            if self.desired_length is not None:
+                s = s[:self.desired_frames]
+            sources_list.append(s)
+
+        mixture, _ = sf.read(mixture_path, dtype='float32')
+        if self.desired_length is not None:
+            mixture = mixture[:self.desired_frames]
```
mpariente, line 49: I rather have a random start each time if the desired_length is specified. 
Rather than always starting from the beginning.

asteroid/data/librimix_dataset.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Args:
+        metadata_path (str): The path to the metatdata file in
+        dataset/metadata
+        desired_length (int) : The desired sources and mixtures length in s
+        sample_rate (int) : The sample rate of the sources and mixtures
+        n_src (int) : The number of sources in the mixture
+    """
+
+    def __init__(self, metadata_path, desired_length=None, sample_rate=16000,
```
mpariente, line 19: Let's keep the same interface a the other datasets and have a `segment` argument instead of `desired_length`.

asteroid/data/librimix_dataset.py
```diff
@@ -0,0 +1,56 @@
...
+                 n_src=2):
+        self.metadata_path = metadata_path
+        self.desired_length = desired_length
+        self.sample_rate = sample_rate
+        # Open csv file
+        self.df = pd.read_csv(metadata_path)
+        if self.desired_length is not None:
+            self.desired_frames = self.desired_length * self.sample_rate
```
mpariente, line 27: This is a number of samples, not of frames.

asteroid/data/librimix_dataset.py
```diff
@@ -0,0 +1,65 @@
...
+        sources_list = []
+        # Read sources
+        for source_ID in source_ID_list:
+            s, _ = sf.read(os.path.join(self.sources_directory_path,
+                                        source_ID + '.wav'))
+            sources_list.append(s)
+
+        mixture, _ = sf.read(mixture_path)
+
+        mixture = torch.from_numpy(mixture).type(
```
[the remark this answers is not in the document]
mpariente: What is the role of the `type`?

egs/libri_2_mix/ConvTasNet/eval.py
```diff
@@ -0,0 +1,149 @@
+import argparse
```
mpariente, line 1: The template eval script has changed a little bit, could you adapt to the newer versions please?

egs/libri_2_mix/ConvTasNet/local/conf.yml
```diff
@@ -0,0 +1,38 @@
...
+  early_stop: yes
+# Optim config
+optim:
+  optimizer: adam
+  lr: 0.001
+  weight_decay: 0.
+# Data config
+data:
+  metadata_train_path : "/home/jcosentino/libri/libri2mix/8K/min/metadata/mixture_train-clean-100.csv"
+  metadata_val_path : "/home/jcosentino/libri/libri2mix/8K/min/metadata/mixture_dev-clean.csv"
```
mpariente, line 34: We're trying that only `run.sh` needs to be modified when running the recipe. We could create a local `./data` folder and copy things there? 
Let's try to avoid needing modifs in the conf file

egs/libri_2_mix/ConvTasNet/local/conf.yml
```diff
@@ -0,0 +1,38 @@
...
+  desired_length: 4
```
mpariente, line 35: Can we change all the segment like the other recipes? 
If the name should be changed, we can make another PR about it.

egs/libri_2_mix/ConvTasNet/local/create_dataset_from_metadata.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+def main(arguments):
+    # Get librispeech root path
+    librispeech_root_path = arguments.librispeech_root_path
+    # Get dataset root path
+    dataset_root_path = arguments.dataset_root_path
+    # Get the desired frequencies
+    freqs = arguments.freqs
+    freqs = [freq.upper() for freq in freqs]
```
mpariente, line 31: `freqs = [freq.upper() for freq in arguments.freqs]`

egs/libri_2_mix/ConvTasNet/local/create_dataset_from_metadata.py
```diff
@@ -0,0 +1,357 @@
...
+    # Get the desired modes
+    modes = arguments.modes
+    modes = [mode.lower() for mode in modes]
```
mpariente, line 34: Same here.

egs/libri_2_mix/ConvTasNet/local/create_metadata.py
```diff
@@ -0,0 +1,440 @@
...
+    # If you wish to ignore some metadata files add their name here
+    # Example : to_be_ignored = ['dev-other.csv']
+    to_be_ignored = []
+    # Check if the metadata files in LibriSpeech already have been used
+    already_generated = os.listdir(mixtures_metadata_directory_path)
+    for generated in already_generated:
+        if generated.startswith("generating_mixture"):
+            to_be_ignored.append(generated.replace("generating_mixture_", ""))
+            print(f"{generated} already exists in "
+                  f"mixtures_metadata_directory_path it won't be overwritten")
```
mpariente, line 198: Missing `{}`?

egs/libri_2_mix/ConvTasNet/model.py
```diff
@@ -0,0 +1,83 @@
...
+        est_masks = self.masker(tf_rep)
+        masked_tf_rep = est_masks * tf_rep.unsqueeze(1)
+        return self.pad_output_to_inp(self.decoder(masked_tf_rep), x)
+
+    @staticmethod
+    def pad_output_to_inp(output, inp):
+        """ Pad first argument to have same size as second argument"""
+        inp_len = inp.size(-1)
+        output_len = output.size(-1)
+        return nn.functional.pad(output, [0, inp_len - output_len])
```
mpariente, line 32: This also changed since then.

egs/librimix/ConvTasNet/eval.py
```diff
@@ -0,0 +1,149 @@
...
+    model = load_state_dict_in(state_copy, model)
+
+    # Handle device placement
+    if conf['use_gpu']:
+        model.cuda()
+    model_device = next(model.parameters()).device
+
+    test_set = LibriMix(conf['test_dir'], None,
+                        conf['sample_rate'],
+                        conf['train_conf']['data']['n_src'])
```
mpariente, line 59: Can you add the keywords in the call please? It makes it much easier to read. In particular, it's not obvious that `None` means, use all segment length, which is crucial !

asteroid/data/librimix_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+        metadata_path (str): The path to the metatdata file
+        sample_rate (int) : The sample rate of the sources and mixtures
+        n_src (int) : The number of sources in the mixture
+        segment (int) : The desired sources and mixtures length in s
+    """
+
+    def __init__(self, metadata_path, sample_rate=16000,
+                 n_src=2, segment='full'):
```
mpariente, line 20: Can we keep the same behaviour as the other loaders : `segment=None` means full utterances

egs/librimix/ConvTasNet/eval.py
```diff
@@ -0,0 +1,149 @@
...
+
+        mix_np = mix.cpu().data.numpy()
+        sources_np = sources.squeeze().cpu().data.numpy()
+        est_sources_np = reordered_sources.squeeze().cpu().data.numpy()
+        est_sources_np[0, :] = est_sources_np[0, :] / \
+                               np.max(est_sources_np, axis=1)[0] * \
+                               np.max(sources_np, axis=1)[0]
+        est_sources_np[1, :] = est_sources_np[1, :] / \
+                               np.max(est_sources_np, axis=1)[1] * \
+                               np.max(sources_np, axis=1)[1]
```
mpariente, line 87: We'll need to get rid of this. Estimate normalization cannot depend directly on the true source.

egs/librimix/ConvTasNet/local/conf.yml
```diff
@@ -0,0 +1,37 @@
+encoder:
+  enc_name : free
+decoder:
+  dec_name : free
+
```
mpariente, line 5: How about sticking to ConvTasnet for a baseline recipe? So we assume encoder/decoder use free filterbanks.

egs/librimix/ConvTasNet/local/conf.yml
```diff
@@ -0,0 +1,37 @@
...
+  early_stop: yes
+# Optim config
+optim:
+  optimizer: adam
+  lr: 0.001
+  weight_decay: 0.
+# Data config
+data:
+  metadata_train_path : "/home/jcosentino/jcosentino/libri/small_snr/8K/min/metadata/mixture_train-clean-360.csv"
+  metadata_val_path : "/home/jcosentino/jcosentino/libri/small_snr/8K/min/metadata/mixture_dev-clean.csv"
```
mpariente, line 34: We need a local path here, not an absolute one.
So the csv files need to be stored in data/wav8k/min/tr for example

egs/librimix/ConvTasNet/local/prepare_data.sh
```diff
@@ -0,0 +1,32 @@
...
+tar -xzf $storage_dir/test-clean.tar.gz -C $storage_dir
+rm -rf $storage_dir/test-clean.tar.gz
+
+echo "Download LibiSpeech/train-clean-100 into $storage_dir"
+# If downloading stalls for more than 20s, relaunch from previous state.
+wget -c --tries=0 --read-timeout=20 http://www.openslr.org/resources/12/train-clean-100.tar.gz -P $storage_dir
+tar -xzf $storage_dir/train-clean-100.tar.gz -C $storage_dir
+rm -rf $storage_dir/train-clean-100.tar.gz
+
+echo "Download LibiSpeech/train-clean-360 into storage_dir"
```
mpariente, line 28: Miss the `$` to storage_dir to print it.

egs/librimix/ConvTasNet/train.py
```diff
@@ -0,0 +1,119 @@
...
+def main(conf):
+    train_set = LibriMix(conf['data']['metadata_train_path'],
+                         conf['data']['sample_rate'],
+                         conf['data']['n_src'],
+                         conf['data']['segment'])
+
+    val_set = LibriMix(conf['data']['metadata_val_path'],
+                       conf['data']['sample_rate'],
+                       conf['data']['n_src'],
+                       conf['data']['segment'])
```
mpariente, line 38: Same here, please us keyword arguments.

egs/librimix/ConvTasNet/eval.py
JorisCos: Sure and you made me realize that my arguments don't match the dataloader arguments ...

asteroid/data/librimix_dataset.py
JorisCos: Ok, but I had a problem with that when I tried to set `segment =None' in the conf file when you load the .yml  None becomes a str "None" and the dataloader breaks. How do you think we should deal with that ?

egs/librimix/ConvTasNet/local/conf.yml
JorisCos: Sure

egs/librimix/ConvTasNet/eval.py
JorisCos: Yeah that's bad. How do you usually deal with normalization ? I guess I could just avoid the clipping by dividing by 'max(est_source)'. Is relu better than sigmoid for mask activation ?

egs/librimix/ConvTasNet/local/conf.yml
JorisCos: Sure, but can you explain why please ?

asteroid/data/librimix_dataset.py
mpariente: I think `null` is what you're looking for.

egs/librimix/ConvTasNet/eval.py
mpariente: This is one way, `src * max(abs(mix))/max(abs(src))` is also used but is not perfect. We could think about better. 
For the mask activation, I remember seeing better results for ReLU but cannot remember where. But, if better, it is marginal.

egs/librimix/ConvTasNet/local/conf.yml
mpariente: The user will modify one thing before running the recipe: `storage_dir` in run.sh. 
After that, the .csv files are stored in ./data and this path is given to the training script. 
This is one less thing for the user to do, so it is better.
If not stored, maybe symlinked.

JorisCos:
I made a mistake when I pushed my last changes. The PR is now really messy. I will close this one and open new one

PR #65, opened 2020-03-24
Integration PMSQE loss in asteroid

mdjuamart:
Two files are added here:

- An implementation of the PMSQE loss function in Pytorch. The loss is implemented as a class whose forward return the symmetric and asymmetric disturbances for a given batch. The batch is composed of padded sequences of TF spectrums (more info in the file). The returned losses should be combined and with additional logMSE function for the implementation of the original paper.

- Additional .mat file with the bark matrices for 16 kHz.


mpariente:
I'm going to make a few edits to fit with Asteroid's coding style, add a test and it'll be ready for merging.
Thank you very much for your PR, it looks great !

mpariente:
Thanks again for the PR, I made some changes to adhere to Asteroid's coding style (every call is the same as before) and added unit-tests. 
The only differences are : 
- Instead of returning both symetrical and asymetrical distorsions, I return the weighted sum (as in the paper and the TF implementation) so that it can be used with MSE directly. 
- I made the `pad_mask` optional in the forward, if it is not fed, we assume an all one mask (i.e no padding).   

#### What's left to do : 
- Support 8khz implementation
- Package the bark_matrix in the setup.py so that the loss can be used if `asteroid` is installed with `pip`.

I'll handle the second point for the next release and let the first point to the community. 

@mdjuamart Please tell me if this looks fine to you, I'll merge this afterwards.

mdjuamart:
The support for 8 kHz is done for Tensorflow and it is easy to integrate into the Pytorch code. You only need to define different variables in the init and use the bark matrix file for 8 kHz. This information can be found in http://sigmat.ugr.es/PMSQE/ (in the code, file perceptual_constants.py). The difference is that the class should allow to select between both modes.

Tell me if you can integrate it in an easy way.

mpariente:
Ended up doing it  :wink:

PR #75, opened 2020-04-15
Fixed two bugs in ConvTasNet

Ariel12321:
I fixed two bugs.
1. asteroid\egs\wham\ConvTasNet\run.sh
the mode variable was being overwritten
2. asteroid\egs\wham\ConvTasNet\train.py
the max_nb_epochs tag was deprecated


mpariente:
Good catch for `mode`, thank you very much ! 
I guess we didn't catch it because once stage 2 is done, we always start from stage 3 and `mode` is not changed, thanks again.

Would you like to correct those on the other wham recipes? That'd be very nice

PR #78, opened 2020-04-18
Improve Encoder and Decoder classes

mpariente:
- Add support for `padding` in `Encoder`
- Add support for `padding` and `output_padding` in  `Decoder`
- Improve the docs a bit


mpariente:
@etzinis, you should be able to use these class in to same way you used yours now. Tell me if it doesn't suits you.

PR #80, opened 2020-04-19
Dynamic mixing wham

popcornell:
Creating pull-request for Dynamic Mixing example. 
Also added a sanitize_configs to allow for lists and tuples on yaml config file.


egs/wham/DynamicMixing/README.md
```diff
@@ -0,0 +1,18 @@
...
+- we added speed perturbation with SoX via [python-audio-effects](https://github.com/carlthome/python-audio-effects).
+- this recipe comes with DPRNN as default model but the model can be swapped by modifying model.py and local/conf.yml. 
+- Original WSJ0 data is needed to run this recipe. 
+
+### Results:
+
+|   model   |   task    |kernel size|chunk size|batch size|SI-SNRi(dB) | SDRi(dB)|
+|:----:|:---------:|:---------:|:--------:|:--------:|:----------:|:-------:|
+| [DPRNN]() | sep_clean |    16     |     100  |     8    |    TODO    |  18.4   |
+
```
mpariente, line 13: Could you please add a line with the corresponding result without DynamicMixing maybe?

egs/wham/DynamicMixing/local/augmented_wham.py
```diff
@@ -0,0 +1,278 @@
+import os
```
mpariente, line 1: Let's keep it here for now and see if we move it to src later.

egs/wham/DynamicMixing/model.py
```diff
@@ -0,0 +1,76 @@
...
+        est_masks = self.masker(tf_rep)
+        masked_tf_rep = est_masks * tf_rep.unsqueeze(1)
+        return self.pad_output_to_inp(self.decoder(masked_tf_rep), x)
+
+    @staticmethod
+    def pad_output_to_inp(output, inp):
+        """ Pad first argument to have same size as second argument"""
+        inp_len = inp.size(-1)
+        output_len = output.size(-1)
+        return nn.functional.pad(output, [0, inp_len - output_len])
```
mpariente, line 32: Can you use the padding function that's used in the DPRNN recipe, it has changed

asteroid/engine/system.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, model, optimizer, loss_func, train_loader,
         self.val_loader = val_loader
         self.scheduler = scheduler
         config = {} if config is None else config
-        self.config = config
+        self.config = sanitize_configs(config)
```
mpariente, line 39: I think the config dict should not be modified. 
Instead, we replace self.none_to_string() by self.config_to_hparams(), and config2hparams contains flatten_dict, none_to_string, and the function you wrote to support tuples and lists. 
As all these functions are only related to hparams, SummaryWriter and System, we can attach it to `System`. 
Tell me if it's clear.

asteroid/utils.py
```diff
@@ -236,6 +236,22 @@ def flatten_dict(d, parent_key='', sep='_'):
     return dict(items)
 
 
+def sanitize_configs(conf_dict):
```
mpariente, line 239: Just to be safe, move this to `System` and modify only the config that's going to be logged to tensorboard (`hparams`)

egs/wham/DynamicMixing/local/augmented_wham.py
popcornell: Yes that was the idea, basically the class right now can do data-augmented wham alongside with plain wham and for orig_percentage==1 is just plain wham

asteroid/engine/system.py
popcornell: I think it is. So we make an unique function for this self.none_to_string(flatten_dict(config)) where we also convert lists and tuples to Tensors

mpariente: Yes.

mpariente:
Thanks !

PR #81, opened 2020-04-20
[src] Generalize ! Implement pairwise losses reduce function

mpariente:
Until now, pairwise losses were averaged over each permutation, then compared to find the best permutation. This averaging is what we call `perm_reduce`. 
This PR removes makes it possible to choose different `perm_reduce` than average (which stays the default).
The `perm_reduce` function can also receive arguments at runtime, we might need it so we made it possible. 

@popcornell Could you review please ?


mpariente:
Thanks ! 
Yes indeed, I'm going to write a unit-test and an example.

PR #82, opened 2020-04-22
No skip option to TDCN

mpariente:
This adds the possibility to make the same architecture as the v1 of the ConvTasnet paper, without skip connections. 

In particular, @etzinis used this version, can you confirm?

### DIFF
--- a/asteroid/masknn/blocks.py
+++ b/asteroid/masknn/blocks.py
@@ -14,6 +14,9 @@ class Conv1DBlock(nn.Module):
         hid_chan (int): Number of hidden channels in the depth-wise
             convolution.
         skip_out_chan (int): Number of channels in the skip convolution.
+            If 0 or None, `Conv1DBlock` won't have any skip connections.
+            Corresponds to the the block in v1 or the paper. The `forward`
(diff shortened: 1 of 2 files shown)


asteroid/masknn/blocks.py
```diff
@@ -40,12 +44,16 @@ def __init__(self, in_chan, hid_chan, skip_out_chan, kernel_size, padding,
...
+        if skip_out_chan:
+            self.skip_conv = nn.Conv1d(hid_chan, skip_out_chan, 1)
 
     def forward(self, x):
         """ Input shape [batch, feats, seq]"""
         shared_out = self.shared_block(x)
         res_out = self.res_conv(shared_out)
+        if not self.skip_out_chan:
+            # Return it twice to have consistent number of outputs
+            return res_out, res_out
```
mpariente, line 56: @popcornell Only thing I'm unsure about : whether I should have consistent outputs or rather output only res_out..

popcornell: No I think is better to return only one argument, in that way is easier to catch errors in the implementation.

mpariente: You're right, changing it

PR #86, opened 2020-04-24
Add FUSS data preparation files

michelolzam:
Data files to get the FUSS dataset.


mpariente:
Thanks a lot again ! 
Merging this :rocket:

PR #87, opened 2020-04-27
LibriMix recipe

JorisCos:
### About this PR
In this PR we introduce the LibriMix dataset. 
We give results on the Libri2Mix version using ConvTasNet. 

Some small changes were made to the eval script to make it easier to run cross dataset evaluation.


egs/librimix/ConvTasNet/eval.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--test_dir', type=str, required=True,
+                    help='Test directory including the json files')
+parser.add_argument('--task', type=str, required=True,
+                    help='One of `enh_single`, `enh_both`, '
+                         '`sep_clean` or `sep_noisy`')
+parser.add_argument('--out_dir', type=str, required=True,
+                    help='Directory in exp_dir where the va will be stored')
```
mpariente, line 29: the va?

egs/librimix/ConvTasNet/eval.py
```diff
@@ -0,0 +1,146 @@
...
+    # Load checkpoint
+    checkpoint = torch.load(best_model_path, map_location='cpu')
+    state = checkpoint['state_dict']
+    state_copy = state.copy()
+    # Remove unwanted keys
+    for keys, values in state.items():
+        if keys.startswith('loss'):
+            del state_copy[keys]
+            print(keys)
+    model = load_state_dict_in(state_copy, model)
```
mpariente, line 56: We should probably add this logic to the usual loading function as the loss func is never in the model. Keep in mind for a future PR

egs/librimix/ConvTasNet/eval.py
```diff
@@ -0,0 +1,146 @@
...
+        loss, reordered_sources = loss_func(est_sources, sources[None],
+                                            return_est=True)
+        mix_np = mix.cpu().data.numpy()
+        sources_np = sources.squeeze().cpu().data.numpy()
+        est_sources_np = reordered_sources.squeeze().cpu().data.numpy()
+        # Declipping the sources
+        est_sources_np[0, :] = est_sources_np[0, :] / \
+                               np.max(est_sources_np, axis=1)[0]
+        est_sources_np[1, :] = est_sources_np[1, :] / \
+                               np.max(est_sources_np, axis=1)[1]
```
mpariente, line 90: This is useful only before saving. 
All our metrics are invariant to scale here this won't always be the case. 
Unless there's a good reason, the normalization should be just before sf.write.

egs/librimix/ConvTasNet/local/conf.yml
```diff
@@ -0,0 +1,33 @@
...
+masknet:
+  n_blocks: 8
+  n_repeats: 3
+  mask_act: relu
+  bn_chan: 128
+  skip_chan: 128
+  hid_chan: 512
+# Training config
+training:
+  epochs: 8
```
mpariente, line 16: We might want to change that to something like 200

egs/librimix/ConvTasNet/local/conf.yml
```diff
@@ -0,0 +1,33 @@
...
+  batch_size: 24
```
mpariente, line 17: This will work only on 4 GPUs, a more reasonnable default of 6 would be better.

egs/librimix/ConvTasNet/run.sh
```diff
@@ -0,0 +1,95 @@
...
+
+exp_dir=exp/train_convtasnet_${tag}
+mkdir -p $exp_dir && echo $uuid >> $exp_dir/run_uuid.txt
+echo "Results from the following experiment will be stored in $exp_dir"
+
+
+if [[ $stage -le 1 ]]; then
+  echo "Stage 1: Training"
+  mkdir -p logs
+  $python_path train.py --exp_dir $exp_dir \
```
mpariente, line 71: Not sure but this will use all GPUs by default. Like this, there is no way to tell to use only two GPUs from the command line. 
Can we have the same `--id` logic as in ConvTasnet?

egs/librimix/ConvTasNet/run.sh
```diff
@@ -0,0 +1,95 @@
...
+  --lr $lr \
+  --weight_decay $weight_decay \
+  --train_dir $train_dir \
+  --valid_dir $valid_dir \
+  --sample_rate $sample_rate \
+  --n_src $n_src \
+  --segment $segment
+fi
+
+out_dir=librimix
```
mpariente, line 90: This should probably not be here but rather before stage 0, where things are read by the user. 
Or even before parse_options so that it can be specified from command line.

asteroid/data/librimix_dataset.py
```diff
@@ -0,0 +1,102 @@
+import numpy as np
```
mpariente, line 1: Make it importable in asteroid/data/__init__.py

egs/librimix/README.md
```diff
@@ -0,0 +1,6 @@
+### About the LibriMix dataset
+
+The LibriMix dataset is an open source dataset 
+derived from LibriSpeech dataset. It's meant as an alternative to [WHAM](./../wham/).
```
mpariente, line 4: [alternative] and a complement to [...]

egs/librimix/ConvTasNet/utils/prepare_python_env.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
mpariente, line 1: Utils folder should be soft linked to wham/ConvTasnet/utils for now to reduce the amount of code copies. 
From librimix/ConvTasnet, delete the utils folder and then run `ln -s ../../wham/ConvTasnet/utils .`

egs/librimix/ConvTasNet/eval.py
JorisCos: Should I remove it from the script and use the usual load_best_model_function ?

egs/librimix/ConvTasNet/eval.py
JorisCos: Ahah missed that thanks

egs/librimix/ConvTasNet/eval.py
JorisCos: Ok i'll make the change

egs/librimix/ConvTasNet/local/conf.yml
JorisCos: Yeah sounds better ahah

egs/librimix/ConvTasNet/run.sh
JorisCos: Yeah it uses all GPU by default I will add the ---id logic

egs/librimix/ConvTasNet/local/conf.yml
JorisCos: Sure

asteroid/data/librimix_dataset.py
JorisCos: Didn't think about that thanks

egs/librimix/ConvTasNet/eval.py
JorisCos: Also realized this syntax is bad. It only works  for 2 sources I will just remove it.

egs/librimix/ConvTasNet/eval.py
mpariente: No, I think it's fine for now. 
I just commented so that I'll remember

egs/librimix/ConvTasNet/utils/parse_options.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
mpariente, line 1: It doesn't look like a symlink but that's fine for now, I have to do it for all the recipes, it's not consistent for now.

mpariente:
Thanks again :rocket:

PR #91, opened 2020-05-04
Two step results

etzinis:


mpariente:
Looks cool, thanks a lot again!
We don't need the second decimal on the results though, I'll remove it.

PR #95, opened 2020-05-10
Source changes towards deep clustering recipe

mpariente:
#### Source changes
- Energy-based VAD which correctly computes VAD independently on batch dim (and any leading dimension)
- Normalize final cost based on VAD mask, independently on batch_dim
- Improve docs in `deep_clustering_loss`. 

#### Egs changes
- Start run.sh with steps and parameters
- Use WHAM-like wsj0-mix preprocessing
- Use WHAM-like data.Dataset (recipe-local for now) 

I'll push the changes to the training and eval script when I have final results, and make another PR for those.

### DIFF
--- a/asteroid/losses/cluster.py
+++ b/asteroid/losses/cluster.py
@@ -9,15 +9,16 @@ def batch_matrix_norm(matrix, norm_order=2):
     Returns:
         normed matrix: torch.Tensor. 
     """
-    return torch.norm(matrix, p=norm_order, dim=[1,2])
+    return torch.norm(matrix, p=norm_order) ** norm_order
 
-def deep_clustering_loss(embedding, tgt_index, spk_cnt=None):
+def deep_clustering_loss(embedding, tgt_index, spk_cnt=None, binary_mask=None):
     """ Compute the deep clustering loss defined in  
     Args:
         embedding: torch.Tensor. Expected shape  [batch, frame x frequency, embeddingDim]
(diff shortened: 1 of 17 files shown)


egs/wsj0-mix/DeepClustering/simple_train.py
```diff
@@ -0,0 +1,238 @@
+import os
```
sunits, line 1: You can remove this file, if the rest of it works

asteroid/losses/cluster.py
```diff
@@ -1,52 +1,76 @@
...
+    """ Normalize a matrix according to `norm_order`
 
+    Args:
+        matrix (torch.Tensor): Expected shape [batch, *]
+        norm_order (int): Norm order.
+
+    Returns:
+        torch.Tensor, normed matrix of shape [batch]
+    """
+    keep_batch = list(range(1, matrix.ndim))
```
sunits, line 75: This was the issue?

mpariente: No this was the not the issue. 
I just made it more general. Before it only worked for 3D tensors. 
For now it is here in the deep clustering loss but we might move it.

egs/wsj0-mix/DeepClustering/simple_train.py
mpariente: I keep it to have it in the master history but I'll remove it in the recipes's PR.

mpariente:
Thanks for the review @sunits !

PR #96, opened 2020-05-11
Deep clustering/Chimera recipe

mpariente:
Adding the Deep clustering / Chimera++ recipe on wsj2mix and wsj3mix.
Mainly based on @sunits initial work. 

Data prep, dataloader, training and evaluation script. 

Evaluatio script will use the mask-inference head whenever possible and DC head only if loss_alpha is equal to 1. 

First successful iteration on the DC head only gets 10.1 dB SDR improvement (not SI-SDR). 

#### Things left to do : 
- Upload some more results on DC alone, Chimera++ (maybe even 3mix)
- Try to use kmeans in native PyTorch instead of sklearn because evaluation is really slow.


egs/wsj0-mix/DeepClustering/model.py
```diff
@@ -32,34 +21,141 @@ def make_model_and_optimizer(conf):
...
+        # Common net
+        out = self.rnn(input_data.permute(0, 2, 1))
+        out = self.dropout(out)
+
+        # DC head
+        proj = self.embedding_layer(out)  # batch, time, freq * emb
+        proj = self.embedding_act(proj)
+        proj = proj.view(batch, n_frames, -1, self.embedding_dim).transpose(1, 2)
+        # (batch, freq * frames, emb)
+        proj = proj.reshape(batch, -1, self.embedding_dim)
```
mpariente, line 67: The bug was here. Without the transpose, the time bins where not aligned with each other and training was impossible. 
I added a note about it in the DC loss.

egs/wsj0-mix/DeepClustering/model.py
```diff
@@ -32,34 +21,141 @@ def make_model_and_optimizer(conf):
...
+    try:
+        # Last best model summary
+        with open(os.path.join(exp_dir, 'best_k_models.json'), "r") as f:
+            best_k = json.load(f)
+        best_model_path = min(best_k, key=best_k.get)
+    except FileNotFoundError:
+        # Get last checkpoint
+        all_ckpt = os.listdir(os.path.join(exp_dir, 'checkpoints/'))
+        all_ckpt.sort()
+        best_model_path = os.path.join(exp_dir, 'checkpoints', all_ckpt[-1])
```
mpariente, line 155: Also, this would be a way to bypass the `best_k_models.json`.

egs/wsj0-mix/DeepClustering/train.py
```diff
@@ -1,144 +1,199 @@
...
-                        )
+                         train_percent_check=1.0,  # Useful for fast experiment
+                         gradient_clip_val=200,)
     trainer.fit(system)
+
     with open(os.path.join(exp_dir, "best_k_models.json"), "w") as f:
         json.dump(checkpoint.best_k_models, f, indent=0)
-    #torch.save(system.model.state_dict(), os.path.join(exp_dir, 'final.pth'))
+    # Save last model for convenience
+    torch.save(system.model.state_dict(),
```
mpariente, line 80: And this would be another one. But if training didn't finish, this model is not saved..

PR #104, opened 2020-05-18
Helper for yaml/bash/python arg passing

mpariente:
#### Why? 
- We don't really like to look in two places for default values. Right now, we check in run.sh, and if it is not declared, in conf.yml to change default. This can be confusing and introduce small errors as well. 

- Writing the same variables three times (conf.yml, bash before parsing, training call after parsing) is useless when we know the args are common.

### Instead, what we want : 
- Writing the config once in conf.yml, enable parsing from CLI, directly into training.py

#### Source code modif
- Handle parsing lists as arguments with `nargs='*'` : `train.py --my_list 1 2` works now.
- Accept keys without values with `nargs='?'`. `train.py --my_arg --my_arg2 1` works. This is useful to deal with empty variables in bash more easily.

#### Utils modif
- yaml2bash.py can print variable assignements from a config (yml). This declares bash env varialbe (exported thanks to `set -a`)
- `parse_options.sh` can then modify the values just declared in this way, from CL agrs.
- yaml2bash.py makes an argument list with update values to be passed to the training 

Lists can be parsed from bash with
```bash
./run.sh --list "0 1"
or
./run.sh --list 0,1
```
Both will be converted to right argparse "code"

Any feedback appreciated

### DIFF
--- a/asteroid/utils.py
+++ b/asteroid/utils.py
@@ -103,8 +103,14 @@ def prepare_parser_from_dict(dic, parser=None):
         group = parser.add_argument_group(k)
         for kk in dic[k].keys():
             entry_type = standardized_entry_type(dic[k][kk])
-            group.add_argument('--' + kk, default=dic[k][kk],
-                               type=entry_type)
+            if entry_type is list:
+                # '*': 0 or more args into list
+                group.add_argument('--' + kk, default=dic[k][kk],
+                                   type=str_int_float, nargs='*')
+            else:
+                # '?' : 0 or 1 arg can be parsed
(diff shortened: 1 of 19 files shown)


popcornell:
That is great and very clever ! 
So all recipes will not have anymore arguments in run.sh unless passed by the user correct ?

mpariente:
Thanks !
We'll still have --id and --storage_dir etc.. 
But all the args from the conf.yml won't be in run.sh. I'll be automated. 

I'm looking for it, but for now, I don't see any downside to this..

mpariente:
Closing this for now, for visibility. 
But we might come back to it later.

Issue #53, opened 2020-03-10
Pretrained models


faroit:
Do you plan to add support for pretrained models (maybe through torchhub?) I think that would make a really nice addition. 

E.g. recipes could be different for training and inference.

mpariente:
> Do you plan to add support for pretrained models (maybe through torchhub?) I think that would make a really nice addition.

Yes, this is on the list of things we'd like to do. I haven't looked into it yet, but yes definitely. 
> 
> E.g. recipes could be different for training and inference.

I don't understand what you mean. Can you explain please?

faroit:
> >  E.g. recipes could be different for training and inference.

> I don't understand what you mean. Can you explain please?

you could maybe add recipes that only serve the purpose of using pretrained models. So that they could easily be deployed on cloud platforms for example

mpariente:
Some news about this. 
Models are derived from data. So models trained on data with restrictive licenses cannot be released. 
wsj0-mix, WHAM and WHAMR are based on WSJ0 which is not open, hence models cannot be released..

That would be another advantage of LibriMix.

faroit:
>  Models are derived from data. So models trained on data with restrictive licenses cannot be released.

I thought this is going to be changed in the EU and is already in place in France. Maybe good to ask inria administration to get the current status on this.

mpariente:
Do you have any resources regarding this change? A link to an article or anything?

mpariente:
Alright, we have the permission of LDC to share wsj0-based models, the first one has been shared. 
I think this can be closed !

PR #109, opened 2020-05-20
[WIP] Importable predefined models 

mpariente:
This is a first step towards sharing pretrained models. 

- Model classes should have three methods 
  - serialize
  - from_pretrained 
  - separate    

Serialize should output a dict that can easily be parsed to create the same model. 
This way, we can prepare all the models in the same way and sharing gets to be much easier. 

The plan here is not to have some kind of base class that handles all the cases. For similar models (TasNet, ConvTasNet and DPRNN for exampe), we can share a common class. If a model needs its own class, that's fine. 
We give priority to simple and understandable code rather than trying to fit all purposes with one class. 

This is the beginning, so any feedback is welcome regarding organization etc..


mpariente:
Just tagging this issue #53 as related

mpariente:
This implies nothing so far so I'll merge. The preparation of the model for publishing will come next

Issue #47, opened 2020-03-07
Music separation datasets and recipes


faroit:
A nice hello from the sigsep gang, 

This looks like a very nice and ambitious approach. Would love to contribute here. Would you be interested in adding music separation things such as

* The musdb dataset
* Music specific augmentations
* multichannel support throughout the package
* Open-unmix model

mpariente:
Wow that would be great! We'd be really glad to have your contributions obviously. Music separation was on the list but we didn't get to it yet. 



> A nice hello from the sigsep gang,
> 
> This looks like a very nice and ambitious approach. Would love to contribute here. Would you be interested in adding music separation things such as
> 
* > The musdb dataset

Yes please !
> 
* > Music specific augmentations

Yes please ! Can you link to a paper or an existing implementation? What do you have in mind for it? Implementation in numpy or torch? 


* > multichannel support throughout the package
> 
Yes please ! Again, what do you have in mind exactly? We are very interested.
The docs was not updated yet but you might be interested by #42, our STFT now supports arbitrary number of channels, the only convention is that time is last (like nn.Conv1D)
* > Open-unmix model

Sure, it would be great !

Feel free to [join the slack](https://join.slack.com/t/asteroid-dev/shared_invite/enQtOTM4NjEzOTI0MzQzLWMyODBmZjRiODAyOGZmNjQ0ZjVjZjM3NGM0NDIyOTc4ZjEyYjc0ZmI1NDI5N2I2YzE5OWU2ZGE1NmEyMjBlYTQ), it will make some things easier.  

I'm going to moderately refactor some things in the next days, maybe I can ping you when it's ready.
Thanks again for your message

mpariente:
@faroit Released 0.1.0 yesterday, should be ready to use now ! :rocket:

faroit:
Thats great. Will check it out soon. In the meantime, I wanted to ask if you are interested to work on a new evaluation tool that should serve the whole community and replace mir_eval.separation and museval. I can add you to the discussion if you are interested

groadabike:
Hi @faroit, I am testing a model (ConvTasNet) by using the current version of the musdb18 dataset and it works great.
I just did small modifications for my specific problem:
- Transform to mono  
- work with 16KH
- Target = vocals.
 
Can I ask you why the mixture is the addition of all sources instead of using the mixture.wav from each directory?
What can I expect in the second draft of MUSDB18 dataset?

Thank you

groadabike:
[...]

PR #110, opened 2020-05-21
First draft of MUSDB18 dataset

faroit:
this PR adds the MUSDB18 dataset and dataloaders as well as a smoke test recipe that includes optimal parameters for training MUSDB18 models.

This dataset is a variant of the [FixedSourcesTrackFolderDataset](https://github.com/sigsep/open-unmix-pytorch/blob/master/data.py#L414) from the Open-Unmix repository.

* it was stripped down to the most basic use cases
* dataloading is limited to soundfile to not increase the number of dependencies on asteroid

This means the dataset is only compatible with MUSDB18HQ and not work with the stems version of MUSDB18. Documentation and scripts to convert MUSDB18 to wav will be added later.

### DIFF
--- a/asteroid/data/musdb18_dataset.py
+++ b/asteroid/data/musdb18_dataset.py
@@ -7,90 +7,114 @@ import soundfile as sf
 
 
 class MUSDB18Dataset(torch.utils.data.Dataset):
-    """A dataset of that assumes audio sources to be stored
-    in track (subb)folder where each folder has a fixed number of sources.
-    For each track the users specifies a list of `source_files` and
-    a common `suffix`.
-    A linear mix is performed on the fly by summing the target and
-    the inferers up.
+    """MUSDB18 music separation dataset
+
+    The dataset consists of 150 full lengths music tracks (~10h duration) of
+    different genres along with their isolated stems:
+        `drums`, `bass`, `vocals` and `others`.
+
+    Out-of-the-box, asteroid does only support MUSDB18-HQ which comes as
+    uncompressed WAV files. To use the MUSDB18, please convert it to WAV first:
+
+    MUSDB18 HQ: https://zenodo.org/record/3338373   
+    MUSDB18     https://zenodo.org/record/1117372 
+
+    Note: The datasets are hosted on Zenodo and require that users
+          request access, since the tracks can only be used for
+          academic purposes. We manually check this requests.
+
+    This dataset asssumes music tracks in (sub)folders where each folder
+    has a fixed number of sources (defaults to 4). For each track, a list
+    of `sources` and a common `suffix` can be specified.
+    A linear mix is performed on the fly by summing up the sources
 
     Due to the fact that all tracks comprise the exact same set
-    of sources, random track mixing augmentation technique
-    can be used, where sources from different tracks are mixed
-    together.
+    of sources, random track mixing can be used can be used,
+    where sources from different tracks are mixed together.
 
-    Example
-    =======
-    train/1/vocals.wav ---------------\
-    train/1/drums.wav -----------------+--> input (mix), output[target]
-    train/1/bass.wav ------------------|
-    train/1/other.wav ----------------/
+    Folder Structure:
+        train/1/vocals.wav ---------------\
+        train/1/drums.wav -----------------+--> input (mix), output[target]
+        train/1/bass.wav ------------------|
+        train/1/other.wav ----------------/
+
+    Args:
+        root (str): Root path of dataset
+        sources (:obj:`list` of :obj:`str`, optional): List of source names.
+            Defaults to MUSDB18 4 stem scenario: `vocals`, `drums`, `bass`, `other`.
+        suffix (str, optional): Filename suffix, defaults to `.wav`.
+        split (str, optional): Dataset subfolder, defaults to `train`.
+        subset (:obj:`list` of :obj:`str`, optional): Selects a specific of
+            list of tracks to be loaded, defaults to `None` (loads all tracks).
+        seq_duration (float, optional): Duration of segments in seconds,
+            defaults to ``None`` which loads the full-length audio tracks.
+        samples_per_track (int, optional):
+            Number of samples yielded from each track, can be used to increase 
+            dataset size, defaults to `1`.
+        random_segments (boolean, optional): Enables random offset for track segments.
+        random_track_mix boolean: enables mixing of random sources from
+            different tracks to assemble mix.
+        source_augmentations (:obj:`list` of `obj`:`callable`):
+            list of augmentation function names,
+            defaults to no-op augmentations (input = output)
+        sample_rate (int, optional): Samplerate of files in dataset.
+
+    Attributes:
+        root (str): Root path of dataset
+        sources (:obj:`list` of :obj:`str`, optional): List of source names.
+            Defaults to MUSDB18 4 stem scenario: `vocals`, `drums`, `bass`, `other`.
+        suffix (str, optional): Filename suffix, defaults to `.wav`.
+        split (str, optional): Dataset subfolder, defaults to `train`.
+        subset (:obj:`list` of :obj:`str`, optional): Selects a specific of
+            list of tracks to be loaded, defaults to `None` (loads all tracks).
+        seq_duration (float, optional): Duration of segments in seconds,
+            defaults to ``None`` which loads the full-length audio tracks.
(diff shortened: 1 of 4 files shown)


faroit:
@mpariente for now, please have a look at the overall structure and the API. I will iterate on that then and will polish before it can "officially" be reviewed

asteroid/data/musdb18_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+                 source_files=['vocals', 'bass', 'drums', 'other'],
+                 suffix='.wav',
+                 seq_duration=None,
+                 samples_per_track=1,
+                 random_chunks=False,
+                 random_track_mix=False,
+                 source_augmentations=lambda audio: audio,
+                 sample_rate=44100):
+
+        """MUSDB18 torch.data.Dataset that samples from the MUSDB18 tracks
```
mpariente, line 43: We follow Google-style docstring style. 
And this should also go in the class docstring, before the `__init__` please

asteroid/data/musdb18_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+        root : str
+            root path of MUSDB18HQ, please download from zenodo
+        suffix : str
+            specify the filename suffice, defaults to `.wav`
+        split : str
+            select dataset subfolder, defaults to ``train``.
+        subsets : list(str)
+            selects a list of track folders, defaults to `None` 
+            which loads all tracks
+        seq_duration : float
```
mpariente, line 57: We use `segment` in the other datasets? Can we change it for consistency?

asteroid/data/musdb18_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+            # optionally select a random track for each source
+            if self.random_track_mix:
+                track_id = random.choice(range(len(self.tracks)))
+            else:
+                track_id = index // self.samples_per_track
+
+            track_path = self.tracks[track_id]['path']
+            if self.random_chunks:
+                min_duration = self.tracks[track_id]['min_duration']
+                start = random.uniform(0, min_duration - self.seq_duration)
```
mpariente, line 103: I didn't run the code but it looks like start is not defined from `random_chunks=False` (default)

asteroid/data/musdb18_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+            specify the filename suffice, defaults to `.wav`
+        split : str
+            select dataset subfolder, defaults to ``train``.
+        subsets : list(str)
+            selects a list of track folders, defaults to `None` 
+            which loads all tracks
+        seq_duration : float
+            training is performed in chunks of ``seq_duration`` (in seconds,
+            defaults to ``None`` which loads the full audio track
+        samples_per_track : int
```
mpariente, line 60: I don't understand the goal of this argument. 
The number of epochs is here for that right? 
Or maybe I don't understand

asteroid/data/musdb18_dataset.py
mpariente: Also, the start from different instruments will be different right? Is this intended?

egs/musdb18/OpenUnmix/test_dataloader.py
```diff
@@ -0,0 +1,134 @@
+from asteroid.data import MUSDB18Dataset
+import argparse
+import torch
+from pathlib import Path
+import tqdm
+
+
+class Compose(object):
```
mpariente, line 8: I've been thinking about adding this for a while.. We'll probably move it inside asteroid.

egs/musdb18/OpenUnmix/test_dataloader.py
```diff
@@ -0,0 +1,134 @@
...
+
+    args = parser.parse_args()
+
+    dataset_kwargs = {
+        'root': Path(args.root),
+    }
+
+    source_augmentations = Compose(
+        [globals()['_augment_' + aug] for aug in args.source_augmentations]
+    )
```
mpariente, line 76: I like that. We would probably opt for a `get` function for augmentations, as we have for norms, filterbanks etc..

asteroid/data/musdb18_dataset.py
mpariente: And `min_duration` is None if `seq_duration` is None, but `random_chunks` can still be True.

mpariente:
The overall structure is perfect! As we discussed, we can search the files only because the DB is very small, but this in fine in this case. 

We might want to integrate the transforms in the source code, but this can be done in a follow-up PR.

asteroid/data/musdb18_dataset.py
faroit: sure, will adopt

asteroid/data/musdb18_dataset.py
faroit: right

asteroid/data/musdb18_dataset.py
faroit: as long as this is not a `IterableDataset` this param will artificially extend one epoch. Without, one epoch would be quite short (100 tracks//batch_size).

egs/musdb18/OpenUnmix/test_dataloader.py
faroit: yes, please

asteroid/data/musdb18_dataset.py
faroit: > I didn't run the code but it looks like start is not defined from random_chunks=False (default)

good catch

faroit: > Also, the start from different instruments will be different right? Is this intended?

yes, it helps a lot. I might want to make this optional, indeed

faroit: | made this behavior optional and fixed the other issues

egs/musdb18/OpenUnmix/test_dataloader.py
faroit: I would change this later when you've added the Compose class to asteroid core

asteroid/data/musdb18_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+        # load sources
+        for source in self.sources:
+            # optionally select a random track for each source
+            if self.random_track_mix:
+                # load a different track
+                track_id = random.choice(range(len(self.tracks)))
+                if self.random_segments:
+                    start = random.uniform(
+                        0, self.tracks[track_id]['min_duration'] - self.segment
+                    )
```
mpariente, line 135: This makes sense to me, at least, when we don't mix track together, we keep things aligned

asteroid/data/musdb18_dataset.py
mpariente: Ok. I don't see the use case but we can keep it for now.

mpariente:
Looks great, thank you very much ! 
Cannot wait for the Open-Unmix recipe  :smiley:

faroit:
>  Cannot wait for the Open-Unmix recipe 😃

actually what about we start with pulling the model using `torch.hub.load('sigsep/open-unmix-pytorch', ...)` and train that in the recipe until there is a need to diverge from our sigsep repo? That would reduce confusion about the asteroid/openunmix vs sigsep/openunmix. And it would still allow us to use asteroid loss functions and asteroid optimizers/tricks...

mpariente:
Oh yes, I didn't think about that actually, it's much better !

faroit:
> Oh yes, I didn't think about that actually, it's much better !

Great, I can do that pretty quickly. I was about to update the training code to lighting anyway

faroit:
@mpariente we forgot to invite @adefossez to review the MUSDB dataset code. @adefossez please review this with respect to mean/std properties you need for demucs and lets add another PR if necessary

mpariente:
I assumed you converged beforehand. Yes, @adefossez feel free to review.

Also : addresses part of #47 .

adefossez:
Hey guys, sorry for replying quite late on this. Indeed I was doing mean/std normalization for training Demucs, at the track level, not segment level, but to be honest, I don't think this is really worth adding here, because it either means loading the entire dataset to compute it every training, or storing some state somewhere which I don't like (what if one of the file changes, etc.)
I think I would rather change slightly the normalization so that it can be performed on the fly, using a moving average of a given window size, with the window being equal or smaller to the training segment duration.

PR #114, opened 2020-05-24
[bin] Create binary folder with parse_options.sh and add it to setup.py

mpariente:
In this way, we can replace the utils which is symlinked everywhere by scripts into `bin` or `utils`.

Any bash (or python, perl etc..) script can be shared in this way and it gets in the PATH. 

After installing asteroid like this, we can directly call `parse_options.sh`

We'll probably move to this behaviour soon and remove the utils folder altogether


danpovey:
Thanks.  Piotr can decide..
my main concern is about how self-explanatory it is, and how discoverable.  But having to source a path.sh
is also not easy for everyone.

mpariente:
Let's close this for now, we probably won't use it.

PR #125, opened 2020-05-27
Publishing pretrained models !!

mpariente:
### :rocket:  New features :rocket: 
- We opened a asteroid-models community both [in Zenodo](https://zenodo.org/communities/asteroid-models) and [Zenodo's sandbox](https://sandbox.zenodo.org/communities/asteroid-models).
- New Zenodo interface (I prefered writing this instead of depending on yet another third party lib)
- New CLI `asteroid-upload` to upload pretrained models directly to the Zenodo community. 

### What's left to do 
- Tests & docs
- Same for other ConvTasNet recipes
- Same for DPRNN recipes
- Add other architectures in `asteroid.models`
- Handle the loading from Hub (should be easy).
- Licenses and infos on the other datasets.

### More about the CLI
The CLI is made with and `entry-point` in the setup.py so it is installed with pip and accessible everywhere.
[This record](https://sandbox.zenodo.org/record/578753#.Xs7leBZ8JC0) (in sandbox) is an example of deposit which was made only by CLI with this command
```bash
asteroid-upload exp/train_convtasnet/publish_dir --uploader "Manuel Pariente" --affiliation "INRIA"
```
The text is auto-generated as well as the attribution in the Licenses etc.. 

Instead of specifying the uploader and affiliation every time in the CLI, the user can create a `uploader_infos.yml` in the package root 
```yaml
uploader: Manuel Pariente
affiliation: INRIA
git_username: mpariente
token: XXX
force_upload: n
```
and the usage becomes `asteroid-upload publish_dir`

### Don't forget
- Zenodo is still in `sandbox` mode by default, we'll need to change that (and provide an arg in the CLI)

Tagging #53

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Project specific
+uploader_info.yml
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/asteroid/__init__.py
+++ b/asteroid/__init__.py
@@ -1,3 +1,5 @@
+import pathlib
 from .utils import deprecation_utils, torch_utils
 
+project_root = str(pathlib.Path(__file__).expanduser().absolute().parent.parent)
 __version__ = '0.2.1'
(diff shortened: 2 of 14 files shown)


mpariente:
The first _real_ pre-trained model is available [here](https://zenodo.org/communities/asteroid-models) with approval of LDC. 

Lots of things left to do, but this is a start, merging

PR #126, opened 2020-05-28
[src&tests] Pretrained model loading

mpariente:
`from_pretrained` supports three types of input:
- filename: 'tmp/model.path'
- Zenodo model name: 'mpariente/ConvTasNetWHAM!...' (find in community)
- Simple URL from Zenodo

We keep a hashtable with all the name: URL mappings for simplicity.
In both last cases, the models are downloaded and cached in `ASTEROID_CACHE` env variable (defaults to '~/.cache/torch/asteroid').

Loading in two lines works
```python
from asteroid.models import ConvTasNet
model = ConvTasNet.from_pretrained('mpariente/ConvTasNet_WHAM!_sepclean')
```

Loading from hub as well.
```python
model = torch.hub.load('mpariente/asteroid:load_pretrained', 'conv_tasnet', 'mpariente/ConvTasNet_WHAM!_sepclean') 
```

Tagging #53

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -4,6 +4,7 @@ import numpy as np
 
 from .. import torch_utils
 from .. import __version__ as asteroid_version
+from ..utils.hub_utils import cached_download
 
 
 class BaseTasNet(nn.Module):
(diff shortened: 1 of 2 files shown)


mpariente:
We will need to work more on the tests later on. The download slows down the build..

PR #127, opened 2020-05-29
Adding AVSpeech and Audio-Visual Speech Separation introduced in Looking to Listen at the Cocktail Party [2].

vitrioil:
Implementation of Audio-Visual Speech Separation introduced in https://arxiv.org/abs/1804.03619.


egs/avspeech/local/requirements.txt
```diff
@@ -0,0 +1,85 @@
...
+threadpoolctl==2.0.0
+tifffile==2020.5.11
+torch==1.5.0
+torchvision==0.6.0
+tqdm==4.46.0
+traitlets==4.3.3
+urllib3==1.25.9
+wcwidth==0.1.9
+Werkzeug==1.0.1
+zipp==3.1.0
```
mpariente, line 85:
```suggestion
catalyst==20.1
facenet-pytorch>=2.2.9
librosa>=0.7.2
mir-eval>=0.6
pysndfx>=0.3.6
threadpoolctl>=2.0.0
torchvision>=0.6.0
zipp>=3.1.0
```

egs/avspeech/local/loader/data.py
```diff
@@ -0,0 +1,125 @@
...
+
+    while (frame < frame_count  and ret):
+        ret, f = video.read()
+        buffer_video[frame] = cv2.cvtColor(f, cv2.COLOR_BGR2RGB)
+
+        frame += 1
+    video.release()
+    return buffer_video
+
+class Signal:
```
mpariente, line 27: Could you refactor this to be in the same file as the `AVSpeechDataset`? I don't see the point on passing the to the dataset? Is it meant to be changed or not?

egs/avspeech/local/loader/data.py
```diff
@@ -0,0 +1,125 @@
...
+        '''
+        self.video = augmenter(self.video, *args, **kwargs)
+
+    def _convert_video(self):
+        if self.embed_saved:
+            return
+        self.buffer_video = get_frames(self.video)
+
+    def _check_video_embed(self, embed_ext=".npy"):
+        from local.loader.constants import EMBED_DIR, SPEC_DIR
```
mpariente, line 79: Also, we shouldn't have these types of constants in this class I guess. We should have these constants in train.py, run.sh or a conf file and pass it to this class if needed.

egs/avspeech/local/loader/audio_feature_generator.py
```diff
@@ -0,0 +1,82 @@
+'''
+    This file deals with raw audio and converts them into
+    spectrogram.
+'''
+import librosa
+import numpy as np
+
+
+def convert_to_spectrogram(audio_raw: np.ndarray, sr=16_000, hann_length=400, hop_length=160, n_fft=512, p=0.3) -> np.ndarray:
```
mpariente, line 9: Also this is passed to the dataset, is this to support different types of features in an easier way?

vitrioil: No, both these files were in the same directory previously. Since, the AVSpeechDataset moved into a different directory, this was a temporary fix. Since we will be using STFT from asteroid itself, I guess keep it as it is for now? As it is going to change anyway.

egs/avspeech/local/loader/data.py
vitrioil: The whole file, right? Yeah, I guess it would be better if we just move it into `avspeech_dataset.py`. However `generate_video_embedding.py` needs `get_frames` as well. So, we can move that function to `asteroid.utils` and import it as `from asteroid.utils import get_frames`?

egs/avspeech/local/loader/data.py
vitrioil: Yeah, that sounds better. Pass it as `__init__` arg?

egs/avspeech/local/loader/data.py
mpariente: Because this is the only dataset that uses video stuff for now, I'd put all the related functions into `avspeech_dataset.py` and add optional requirements in the setup. The install is already pretty heavy and avspeech will need lots of stuff. Does it sound ok for you?

egs/avspeech/local/loader/audio_feature_generator.py
mpariente: I would change now to asteroid STFT actually ^^ Is it possible?

egs/avspeech/local/loader/data.py
mpariente: Yes please

mpariente:
In general, we want something like this 
```
from asteroid.data.avspeech_dataset import AVSpeechDataset
train_set = AVSpeechDataset(train_dir, **some_more_args)
```
where the `train_dir` was created during the recipe and contains text files (json, csv, anything) that contain absolute path to the files of interested (prepared by the recipe). 

So, we want:
- `AVSpeech` has a very simple interface based on text files in a local directory so we don't need to glob or find at the beginning of each training (you can assume the directory to have the structure you want)
- The dataprep part of the recipe which prepares the data and creates this directory with approriate text file.

asteroid/data/__init__.py
```diff
@@ -3,4 +3,4 @@
 from .dns_dataset import DNSDataset
 from .librimix_dataset import LibriMix
 from .wsj0_mix import Wsj0mixDataset
-from .avspeech_dataset import AVSpeechDataset
+from .avspeech_dataset import AVSpeechDataset, get_frames
```
mpariente, line 7: We won't be able to have this import because of all the dependencies in avspeech_dataset.py. 
Those will be extra_requirements and installed in the recipe.

asteroid/data/avspeech_dataset.py
```diff
@@ -1,25 +1,186 @@
...
+        self.embed_dir = embed_dir
         self.input_df = pd.read_csv(input_df_path.as_posix())
-        self.signal_constructor = signal_constructor
-        self.convert_to_spectrogram = convert_to_spectrogram
+
+    @staticmethod
+    def encode(x: np.ndarray, p=0.3, stft_encoder=None):
+        if stft_encoder is None:
+            stft_encoder = Encoder(STFTFB(n_filters=512, kernel_size=400,
+                                          stride=160))
```
mpariente, line 153: This will create an encoder everytime, it would be better to define these in `__init__`. 
In this case, you can either use self.stft_encoder, or keep themethod static and pass the encoder but why would you like to keep this static?

asteroid/data/avspeech_dataset.py
```diff
@@ -1,25 +1,186 @@
...
+
+        x = torch.from_numpy(x).float()
+
+        # time domain to time-frequency representation
+        tf_rep = stft_encoder(x).squeeze(0).unsqueeze(2)
+        # power law on complex numbers
+        tf_rep = (torch.abs(tf_rep) ** p) * torch.sign(tf_rep)
```
mpariente, line 160: I would add an EPS to `tf_rep` here as (**p) can have NaN gradient.

asteroid/data/avspeech_dataset.py
```diff
@@ -1,25 +1,186 @@
...
+        # (514, 298) -> (257, 298, 2)
+        tf_rep = torch.cat((tf_rep[:257], tf_rep[257:]), axis=2)
```
mpariente, line 162: We have helper functions `transforms.from_torchaudio`, `transforms.to_torchaudio`. This will avoid hardcoding the 257 frequencies.

asteroid/data/avspeech_dataset.py
vitrioil: We would still need both of these methods during inference. So, we will have to create the `AVSpeechDataset` object during inference as well.

asteroid/data/avspeech_dataset.py
vitrioil: Oh, okay. Will do that.

mpariente:
This looks nice !
I think we could try to make the `Signal` class a bit more generic to hold audio visual signals of "different" nature. The network needs to receive the features in a synchronized way and this depends on the FPS of the video, the sample rate of the audio and the STFT parameters. If we could have a class that automatically computes the interpolation factor that needs to be applied depending on this variables, it would make the class much more flexible

vitrioil:
Abstract class maybe? Implement more children according to the needs? Or just one huge class?

asteroid/data/avspeech_dataset.py
mpariente: Maybe, this is related to the general comment I made. We would like a class that handles feature computation and audio-video alignement (which is configurable as well). And this class would be used in the `Dataset` object, but this would also be the class that we could use at inference to feed the network with data. Would this make sense?

mpariente:
Let's start with a class that receives parameters in `__init__` (audio sampling rate, video FPS, stft parameters etc..) and has a method that takes an audio file and video file and computes aligned features. Does this make sense? It can be huge if needed, we'll see if we need to split it up or not.

vitrioil:
> Let's start with a class that receives parameters in `__init__` (audio sampling rate, video FPS, stft parameters etc..) and has a method that takes an audio file and video file and computes aligned features. Does this make sense? It can be huge if needed, we'll see if we need to split it up or not.

Sounds good.

asteroid/data/__init__.py
vitrioil: I see. Where should I put this file? Inside `egs/avspeech/`?

mpariente: No no, it can stay here. Just you'll have to import it as
`from asteroid.data.avspeech_dataset import ...`

asteroid/data/avspeech_dataset.py
```diff
@@ -12,7 +12,9 @@
 from typing import Callable, Tuple, List
 
 from asteroid.filterbanks import (Encoder, Decoder,
-                                  STFTFB)
+                                  STFTFB, transforms)
+
+EPS = 1e-6
```
mpariente, line 17:
```suggestion
EPS = 1e-8
```

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+
+        self._load(sr=sr)
+        self._check_video_embed()
+        self._convert_video()
+
+    def _load(self, sr: int):
+        self.audio = None
+        if self.load_spec and self.spec_path.is_file():
+            self.audio = np.load(self.spec_path)
+            self._is_spec = True
```
mpariente, line 75: In the current usage, self.spec_path.is_file() will return False right?

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+            audio = librosa.load(str(audio_path), sr=sr)[0]
+        return audio, spec_exists, spec_path
+
+
+class AVSpeechDataset(data.Dataset):
+
+    def __init__(self, input_df_path: Path,
+                 embed_dir: Path,
+                 input_audio_size=2):
+        """
```
mpariente, line 138: Docstring should be above init.

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+            spec_exists = True
+            audio = np.load(spec_path)
+        else:
+            audio = librosa.load(str(audio_path), sr=sr)[0]
+        return audio, spec_exists, spec_path
+
+
+class AVSpeechDataset(data.Dataset):
+
+    def __init__(self, input_df_path: Path,
```
mpariente, line 135: Input arguments should also accept strings, it's a simple check in the input (if str, convert to Path).

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def __len__(self):
+        return len(self.input_df)
+
+    def __getitem__(self, idx):
+        row = self.input_df.iloc[idx, :]
+        all_signals = []
+
+        for i in range(self.input_audio_size):
+            #get audio, video path from combination dataframe
```
mpariente, line 194: The combination dataset has `video 1, video 2, audio 1, audio 2` columns or something like this?
It would be clearer to have names instead of these indices IMO.

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        else:
+            audio = librosa.load(str(audio_path), sr=sr)[0]
+        return audio, spec_exists, spec_path
+
+
+class AVSpeechDataset(data.Dataset):
+
+    def __init__(self, input_df_path: Path,
+                 embed_dir: Path,
+                 input_audio_size=2):
```
mpariente, line 137: Is this the number of sources in the mixture or is it something else?

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+                spectrogram = self.encode(all_signals[i].get_audio(),
+                                          stft_encoder=self.stft_encoder)
+            #convert to tensor
+            audio_tensors.append(spectrogram)
+
+            #check if the embedding is saved
+            if all_signals[i].embed_is_saved() and all_signals[i].get_embed() is not None:
+                embeddings = torch.from_numpy(all_signals[i].get_embed())
+                video_tensors.append(embeddings)
+                continue
```
mpariente, line 231: Why `continue`?

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+            #audio to spectrogram
+            if all_signals[i].spec_path.is_file():
+                spectrogram =  all_signals[i].get_spec()
+            else:
+                spectrogram = self.encode(all_signals[i].get_audio(),
+                                          stft_encoder=self.stft_encoder)
+            #convert to tensor
+            audio_tensors.append(spectrogram)
+
+            #check if the embedding is saved
```
mpariente, line 227: In the future, it would be nice to be able to do this on the fly and to cache the computed embeddings. 
But that's not for now I guess

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+            if all_signals[i].embed_is_saved() and all_signals[i].get_embed() is not None:
+                embeddings = torch.from_numpy(all_signals[i].get_embed())
+                video_tensors.append(embeddings)
+                continue
+
+        mixed_signal_tensor = torch.transpose(mixed_signal,0,2) #shape (2,298,257)  , therefore , 2 channels , height = 298 , width = 257
+        audio_tensors = [i.transpose(0, 2) for i in audio_tensors]
+        audio_tensors = torch.stack(audio_tensors)
+        audio_tensors = audio_tensors.permute(1, 2, 3, 0)
```
mpariente, line 236: Is this `(channels, time, freq, sources)`?

egs/avspeech/local/requirements.txt
```diff
@@ -0,0 +1,8 @@
+catalyst==20.1
+facenet-pytorch>=2.2.9
+librosa>=0.7.2
+mir-eval>=0.6
+pysndfx>=0.3.6
+threadpoolctl>=2.0.0
+torchvision>=0.6.0
+zipp>=3.1.0
```
mpariente, line 8: Need to add open-cv here (my bad)

egs/avspeech/model.py
```diff
@@ -0,0 +1,381 @@
...
+
+    def get_padding(self,kernel_size,dilation):
+        padding = (((dilation[0])*(kernel_size[0]-1))//2,((dilation[1])*(kernel_size[1]-1))//2)
+        return padding
+
+    def forward(self,input_audio):
+        # input audio will be (2,298,257)
+
+        output_layer = F.relu(self.batch_norm1(self.conv1(input_audio)))
+        #print (output_layer.shape)
```
mpariente, line 159: Remove print statements.

egs/avspeech/model.py
```diff
@@ -0,0 +1,381 @@
...
+        output_layer = F.relu(self.batch_norm13(self.conv13(output_layer)))
+        output_layer = F.relu(self.batch_norm14(self.conv14(output_layer)))
+        output_layer = F.relu(self.batch_norm15(self.conv15(output_layer)))
+
+        # output_layer will be (N,8,298,257)
+        # we want it to be (N,8*257,298,1)
+        batch_size = output_layer.size(0) # N
+        height = output_layer.size(2)  # 298
+
+        output_layer = output_layer.view(batch_size,-1,height,1)
```
mpariente, line 191:
```
# output_layer will be (N,8,298,257)
# we want it to be (N,8*257,298,1)
```
This will mix the time and freq axis, you need to transpose before I guess. You should check Row and column major ordering. 
I might be wrong but I think there is a mistake here.

egs/avspeech/model.py
```diff
@@ -0,0 +1,381 @@
...
+class Audio_Visual_Fusion(nn.Module):
+
+    def __init__(self,num_person=2,device=None):
+        if isinstance(device, str):
+            device = torch.device(device)
+
+        self.device = device
+        super(Audio_Visual_Fusion, self).__init__()
+        self.num_person = num_person
+        self.input_dim = 8*257 + 256*self.num_person
```
mpariente, line 270: Do you want to try to get rid of the constants and depend on an `input_dim` param that can be computed from the STFT?

egs/avspeech/model.py
```diff
@@ -0,0 +1,381 @@
...
+import torch.nn.functional as F
+
+from pathlib import Path
+
+from asteroid import torch_utils
+from asteroid.engine.optimizers import make_optimizer
+
+
+# Reference: https://github.com/bill9800/speech_separation/blob/master/model/lib/utils.py
+def generate_cRM(Y,S):
```
mpariente, line 14: Is this in mask utils as well?

vitrioil:
Thank you for referring  https://github.com/psf/black. It is really good!

egs/avspeech/model.py
vitrioil: Oh, good point!! I did not notice that!

Referring to this [answer](https://discuss.pytorch.org/t/whats-the-best-way-to-concatenate-these-pytorch-dimensions/35737/2).

Something like this should solve:
`output_layer.permute(0, 1, 3, 2).contiguous().view(batch_size, -1, height, 1)`

What do you think?

mpariente: That would work. Personally I would use `transpose` and `reshape`. 
```python 
output_layer.transpose(-1, -2).reshape((batch_size, -1, height, 1)
```
Using `transpose` makes that you have to read two axis instead of 4, it's easier to see the intent IMO. 
`reshape` is `view` if the resulting array is contiguous, otherwise it's `.contiguous().view()`

vitrioil: Good to know, thanks!

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,234 @@
...
+    def get_embed(self):
+        return self.embed
+
+    def get_audio(self):
+        return self.audio
+
+    @staticmethod
+    def load_audio(audio_path: Union[str, Path], sr=16_000):
+        audio, _ = librosa.load(audio_path, sr=sr)
+        return audio
```
mpariente, line 119: This is used only once, remove it.

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,234 @@
...
+            signal = Signal(
+                video_path,
+                audio_path,
+                self.embed_dir,
+                video_start_length=video_length_idx,
+            )
+            all_signals.append(signal)
+
+        # input audio signal is the last column.
+        mixed_signal = Signal.load_audio(row.loc["mixed_audio"])
```
mpariente, line 209: And then load with librosa here. 
At least, we know we don't do anything fancy with the loading, without reading other lines of code.

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,234 @@
...
+            # get embed
+            embeddings = torch.from_numpy(all_signals[i].get_embed())
+            video_tensors.append(embeddings)
+
+        mixed_signal_tensor = torch.transpose(
+            mixed_signal, 0, 2
+        )  # shape (2,298,257)  , therefore , 2 channels , height = 298 , width = 257
+        audio_tensors = [i.transpose(0, 2) for i in audio_tensors]
+        audio_tensors = torch.stack(audio_tensors)
+        audio_tensors = audio_tensors.permute(1, 2, 3, 0)
```
mpariente, line 231: This could be written
```python
torch.stack(audio_tensors, dim=-1)
```
right?

asteroid/data/avspeech_dataset.py
mpariente: I don't think this should be resolved. 
This is a weird format for the spectrograms IMO

asteroid/data/avspeech_dataset.py
```diff
@@ -0,0 +1,234 @@
...
+            input_audio_size (int): total audio/video inputs.
+
+        References:
+            [1]: 'Looking to Listen at the Cocktail Party:
+            A Speaker-Independent Audio-Visual Model for Speech Separation' Ephrat et. al
+            https://arxiv.org/abs/1804.03619
+    """
+    dataset_name = "AVSpeech"
+
+    def __init__(self, input_df_path: Union[str, Path], embed_dir: Union[str, Path], input_audio_size=2):
```
mpariente, line 137: `input_audio_size` is the number of sources?

asteroid/data/avspeech_dataset.py
mpariente: If we return spectrograms, they should be returned as Asteroid usually returns them otherwise it's confusing. This shape is required by this architecture but not necessarily by future ones. 
A _normal_ shape would be (sources, frequencies, time), where frequencies contain both Re and Im. The spliting an reshaping would happen in the network, like we would start from the waveform in the network. Do you agree?

egs/avspeech/README.md
```diff
@@ -0,0 +1,14 @@
...
+- pysndfx
+- catalyst
+- mir_eval
+- alchemy: pip install alchemy-catalyst
+
+### Notes: 
+- Download Interruptible but starts from scratch if interrupted.
+- Need storage_dir variable even if not env variable : 
+    set it where needed.
+
```
mpariente, line 14:
```suggestion
```

mpariente: Removed my notes. 
We should add a small description of the dataset here and link to the paper.

egs/avspeech/local/plot_spec.py
```diff
@@ -0,0 +1,80 @@
+#TODO: make it compatible with current changes
```
mpariente, line 1: Is this still relevant?

egs/avspeech/local/loader/generate_video_embedding.py
```diff
@@ -0,0 +1,125 @@
+import cv2
+import torch
+import numpy as np
+import pandas as pd
+
```
mpariente, line 5: General imports rule. 
No space between them, until the local ones

egs/avspeech/local/predict.py
```diff
@@ -0,0 +1,163 @@
+#TODO: make it compatible with current changes
```
mpariente, line 1: Still relevant?

egs/avspeech/model.py
```diff
@@ -0,0 +1,533 @@
...
+from asteroid.engine.optimizers import make_optimizer
+
+
+# Reference: https://github.com/bill9800/speech_separation/blob/master/model/lib/utils.py
+def generate_cRM(Y, S):
+    """
+    :param Y: mixed/noisy stft  (N,H,W,C)
+    :param S: clean stft (N,H,W,C)
+    :return: structed cRM
+    """
```
mpariente, line 19: Google style docstring please

mpariente: In the whole file please.

egs/avspeech/train/callbacks.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    def __init__(
+        self,
+        input_key: str = "targets",
+        output_key: str = "logits",
+        prefix: str = "snr",
+        mixed_audio_key: str="input_audio"
+    ):
+        """
```
mpariente, line 23: Docs above init

egs/avspeech/train/callbacks.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    def __init__(
+        self,
+        input_key: str = "targets",
+        output_key: str = "logits",
+        prefix: str = "sdr",
+        mixed_audio_key: str="input_audio"
+    ):
+        """
```
mpariente, line 71: Same here

vitrioil:
So, currently output shape for spectrogram in dataloader is `(num_sources, 514, 298)`. I am converting that back to `(2, 298, 257, num_sources)` and reconverting it to `(num_sources, 514, 298)`. Let me know, if that is okay. Also, input to network being a spectrogram is fine? or should it be waveform and do the conversion inside as well?

egs/avspeech/local/predict.py
vitrioil: I removed it completely since it is not being used... That's okay, right?

mpariente: Yes.

mpariente:
Ah I just see now that you added it in the docstrings. 
If it is the number of sources, can we call it `n_src` like in other datasets? 
If it's something else, could you explain? 

Otherwise, LGTM, great job thanks again !

vitrioil:
Yeah, no particular reason. `n_src` is consistent. I'll do that in the morning.

PR #128, opened 2020-05-29
[egs] GPU `id` defaults to $CUDA_VISIBLE_DEVICES in all recipes

mpariente:
We've had feedback and this is the most common usecase so far

`find . -iname run.sh -exec sed -E -i 's/id=.?$/id=\$CUDA_VISIBLE_DEVICES/g' {} +`


mpariente:
Zenodo is down so this won't pass.
But these are _just_ egs, not tested. 
Merging

PR #129, opened 2020-05-29
[egs] set -e in all recipes to exit or errors

mpariente:
```bash
find . -iname run.sh -exec sed -i '/set -e.*/ d' {} \;
find . -iname run.sh -exec sed -i '2iset -e  # Exit on error' {} \;
```


mpariente:
This is better than going though evaluation after training has crashed

PR #132, opened 2020-05-30
[egs] Enable pretrained models sharing on LibriMix

mpariente:
Follow the example of wham/ConvTasNet and share on Librimix, pretrained models to come !

@JorisCos, can you review please?


JorisCos:
Looks great to me ! I can take care of the pretrained models. Do you think we should publish all of them ? ( I mean the 8 models  8k/16k min/max sep_clean/sep_noisy)

mpariente:
> Looks great to me ! I can take care of the pretrained models. Do you think we should publish all of them ? ( I mean the 8 models 8k/16k min/max sep_clean/sep_noisy)

Cool, merging. 
8k sep_clean and 8k sep_noisy would be great to start with !

PR #143, opened 2020-06-04
[src] Add targets argument to musdb18_dataset.py

mpariente:
The is to address the "issue" mentioned in #47.
We can add a `targets` argument to the dataset so that the return sources are in the form of a Tensor (with the required target sources stacked in the given order) instead of the dict.

```python
# Mix all sources and return a dict
MUSDB18Dataset(root)
# Mix only 'vocals', 'bass', 'drums' and return dict
MUSDB18Dataset(root, sources=['vocals', 'bass', 'drums'])
# Mix only 'vocals', 'bass', 'drums' and return stacked vocals and drums (sorry bass)
MUSDB18Dataset(root, sources=['vocals', 'bass', 'drums'], targets=['vocals', 'drums'])
```
The main use case is that we don't have to subclass `System` to use it, which makes sense!

@faroit Could you review please and tell me if you'd be fine with that
@groadabike, would this suit your needs?

### DIFF
--- a/asteroid/data/musdb18_dataset.py
+++ b/asteroid/data/musdb18_dataset.py
@@ -40,8 +40,13 @@ class MUSDB18Dataset(torch.utils.data.Dataset):
 
     Args:
         root (str): Root path of dataset
-        sources (:obj:`list` of :obj:`str`, optional): List of source names.
+        sources (:obj:`list` of :obj:`str`, optional): List of source names
+            that composes the mixture.
             Defaults to MUSDB18 4 stem scenario: `vocals`, `drums`, `bass`, `other`.
+        targets (list or None, optional): List of source names to be used as
+            targets. If None, a dict with the 4 stems is returned.
+             If e.g [`vocals`, `drums`], a tensor with stacked `vocals` and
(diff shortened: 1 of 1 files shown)


faroit:
Yes a list of target can make sense but I don't understand why this would need to be a stacked tensor instead of a dict. Couldn't users do the stacking inside their model or loss?

mpariente:
> Yes a list of target can make sense but I don't understand why this would need to be a stacked tensor instead of a dict. Couldn't users do the stacking inside their model or loss?

They could. But then it's almost the same as not having the targets argument.
If you use ConvTasNet with a L1 loss (for example) and MUSDB18 with targets arg, you would be able to use the default `System` ([see here](https://github.com/mpariente/asteroid/blob/master/asteroid/engine/system.py#L79) for the forward that would use the stacked tensors in the loss).

It's just a matter of ease-of-use IMO.

faroit:
> It's just a matter of ease-of-use IMO.

yes, ok. I agree then

mpariente:
Cool thanks !

PR #144, opened 2020-06-05
[src] LibriMix from download

mpariente:
Instantiate LibriMix from download in one line. 
Needs review and basic testing but it works on my side.


JorisCos:
Looks greats ! You can merge 🚀

PR #146, opened 2020-06-08
[src & tests] Bidirectional residual RNN

mpariente:
Finally got time to examine the output of bidir RNN.

@popcornell Could you review tell me if it makes sense for you? 
Thanks

### DIFF
--- a/asteroid/masknn/recurrent.py
+++ b/asteroid/masknn/recurrent.py
@@ -88,6 +88,63 @@ class StackedResidualRNN(nn.Module):
         return x
 
 
+class StackedResidualBiRNN(nn.Module):
+    """ Stacked Bidirectional RNN with builtin residual connection.
+    Residual connections are applied on both RNN directions.
+    Only supports bidiriectional RNNs.
+    See StackedResidualRNN for unidirectional ones.
+
+    Args:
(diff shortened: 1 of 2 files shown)


asteroid/masknn/recurrent.py
```diff
@@ -88,6 +88,63 @@ def forward(self, x):
...
+        self.dropout_layer = nn.Dropout(self.dropout)
+
+    def forward(self, x):
+        """ Builtin residual connections + dropout applied before residual.
+            Input shape : [batch, time_axis, feat_axis]
+        """
+        # First layer
+        rnn_out = self.first_layer(x)
+        dropped_out = self.dropout_layer(rnn_out)
+        x = torch.cat([x, x], dim=-1) + dropped_out
```
popcornell, line 141: so you just replicate the input and sum easy peasy

mpariente:
By the way, this is the way they do for non-residual bidirectional RNN. From the second layer, the layer has 2N input. 
And the current approach has less parameters I think, because ih (input to hidden) in 2NxN but the hh (hidden to hidden) is still NxN.

asteroid/masknn/recurrent.py
mpariente: Yes, residual in both directions. That would be my first guess. To be tried.

PR #147, opened 2020-06-09
Fix librimix

mhu-coder:
Hey guys,
I have made some minor modifications to the librimix recipe.
*  `eval.py`:
  * `soundfile` will clip any value larger than 1.0 to `2^15-1` when writing a float array to int16, which makes the enhanced data unusable when read from the audio file.
*  `local/prepare_data.sh`:
  * `generate_librimix.sh` never uses parameter `$2`. I don't really understand the purpose of `$n_src` in the script.
* `run.sh`:
  * Added the possible values of `task` so that people do not have look for it (in `asteroid/data/librimix_dataset.py`
  * Added the `task` flag to `train.py` call as the program would crash without it when the task is `enh_single`.

What do you think?

Cheers,
Mathieu


egs/librimix/ConvTasNet/eval.py
```diff
@@ -87,6 +88,7 @@ def main(conf):
                 sf.write(local_save_dir + "s{}.wav".format(src_idx), src,
                          conf['sample_rate'])
             for src_idx, est_src in enumerate(est_sources_np):
+                est_src /= np.max(np.abs(est_src))  # to avoid clipping
```
mpariente, line 91: Then, you can multiply by the largest value in the mix. Doesn't make sense to have all the output have max of 1.

egs/librimix/ConvTasNet/local/prepare_data.sh
```diff
@@ -6,11 +6,16 @@ python_path=python
...
 git clone https://github.com/JorisCos/LibriMix
 
 # Run generation script
+if [ -n $n_src ]; then
+  echo "DEPRECATION WARNING: n_src has no effect on the generated data and"
+  echo "will be removed from future release"
+fi
+
 cd LibriMix
 . generate_librimix.sh $storage_dir $n_src
```
mpariente, line 20: You can remove `$n_src` from here.

egs/librimix/ConvTasNet/local/prepare_data.sh
```diff
@@ -6,11 +6,16 @@ python_path=python
...
+current_dir=$(pwd)
 # Clone LibriMix repo
 git clone https://github.com/JorisCos/LibriMix
 
 # Run generation script
+if [ -n $n_src ]; then
+  echo "DEPRECATION WARNING: n_src has no effect on the generated data and"
+  echo "will be removed from future release"
+fi
+
```
mpariente, line 18: This can be removed.
And the `n_src=` in line 4 as well

egs/librimix/ConvTasNet/run.sh
```diff
@@ -43,7 +43,7 @@ test_dir=data/wav8k/min/test
 sample_rate=8000
 n_src=2
 segment=3
-task=sep_clean
+task=sep_clean  # one of 'enh_single', 'enh_both', 'sep_clean', 'sep_noisy'
 
 . utils/parse_options.sh
```
mpariente, line 48: And remove the `--n_src $n_src` in line 52 as well.

mhu-coder:
Oops, I didn't notice that none of the commit had an author. I gave a go at git subtrees in my project, where I use asteroid as a dependency but contributing back is not optimal... should have stuck with git submodules instead. I am closing to re-open a PR with a cleaner history, your suggestions and my name as the author.
Thanks for the comments!

PR #148, opened 2020-06-09
Fix librimix corrected

mhu-coder:
Hello again,
This time with a cleaner commit history and your suggestions.

Regarding the normalization before writing to disk, anything that avoids significant amount of clipping is possible. Matching the peak value to that of the mixture as you suggested is one possibility, so is matching the energy of the target source but I am not sure if any of these really matters.

Cheers,
Mathieu

### DIFF
--- a/egs/librimix/ConvTasNet/local/prepare_data.sh
+++ b/egs/librimix/ConvTasNet/local/prepare_data.sh
@@ -1,19 +1,18 @@
 #!/bin/bash
 
 storage_dir=
-n_src=
 python_path=python
 
 . ./utils/parse_options.sh
 
-current_dir= pwd
+current_dir=$(pwd)
 # Clone LibriMix repo
 git clone https://github.com/JorisCos/LibriMix
 
 # Run generation script
 cd LibriMix
-. generate_librimix.sh $storage_dir $n_src
+. generate_librimix.sh $storage_dir
 
 cd $current_dir
-$python_path local/create_local_metadata.py --librimix_dir $storage_dir/Libri$n_src"Mix"
+$python_path local/create_local_metadata.py --librimix_dir $storage_dir/LibriMix
 
--- a/egs/librimix/ConvTasNet/local/prepare_data.sh
+++ b/egs/librimix/ConvTasNet/local/prepare_data.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 storage_dir=
+n_src=
 python_path=python
 
 . ./utils/parse_options.sh
@@ -14,5 +15,5 @@ cd LibriMix
 . generate_librimix.sh $storage_dir
 
 cd $current_dir
-$python_path local/create_local_metadata.py --librimix_dir $storage_dir/LibriMix
+$python_path local/create_local_metadata.py --librimix_dir $storage_dir/Libri$n_src"Mix"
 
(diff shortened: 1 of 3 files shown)


mpariente:
Matching the energy of the target source is cheating, you shouldn't be able to know it ! ;-)

egs/librimix/ConvTasNet/local/prepare_data.sh
```diff
@@ -1,19 +1,18 @@
...
 git clone https://github.com/JorisCos/LibriMix
 
 # Run generation script
 cd LibriMix
-. generate_librimix.sh $storage_dir $n_src
+. generate_librimix.sh $storage_dir
 
 cd $current_dir
-$python_path local/create_local_metadata.py --librimix_dir $storage_dir/Libri$n_src"Mix"
+$python_path local/create_local_metadata.py --librimix_dir $storage_dir/LibriMix
```
mpariente, line 17: Hmm actually, I overlooked that part. 
Are you sure this will work with the rest of the recipe? 
@JorisCos would have a better overview than me about this.

mhu-coder: From my understanding, the audio data are dumped in `${storage_dir}/Librimix` then the `.csv` files are created using `create_local_metadata` from `local/prepare_data.sh`, which effectively copies the `.csv` files from `${storage_dir}/LibriMix` into appropriate directories under `data/` (that is created at the root of the librimix directory).
So the name `LibriMix` itself should not matter but I have to say, I struggled to understand some parts of the code so don't quote me on that. Any thoughts @JorisCos?

mpariente: My only concern is if the files have the same name of 2mix and 3mix,they would overwrite each other if it's the same folder. 
He'll review and tell,it'll be faster than me reading through the code again

JorisCos: Hello ! Thanks for the PR :)
I don't think this is correct. The path given to `create_local_metadata.py` is the path to the root of `Libri2Mix` or `Libri3Mix` that was created during the generation process in `storage_dir.` No `LibriMix` folder should be there.
The purpose of `create_local_metadata.py` is to collect and reorder all the metadata files that are stored in `storage_dir/Libri$n_srcMix` and to copy them in the recipe folder. 
In `storage_dir` you have `Libri2Mix` and `Libri3Mix` inside these folders the structure is the same  
`storage_dir/wav8k/min/` where you can change wav8k for wav16k and min for max. In the min folder you have folder called `metadata` this contains the metadata for all subsets that we need to collect and reorder.

mhu-coder: So we put back the `n_src` option in `prepare_data.sh` and use it as such:

```bash
n_src=2

# [...]

$python_path local/create_local_metadata.py --librimix_dir $storage_dir/Libri${n_src}Mix
```
and the change would be applied to `run.sh` as well?

JorisCos: Looks good to me ;)

mhu-coder: Ok, done.

JorisCos: Thanks again for the PR !

mpariente: Sorry, I overlooked that part in the first place. 
LGTM thans

PR #152, opened 2020-06-12
SEGAN recipe 

JorisCos:
# About this PR
This PR introduce a new pytorch-lightning module named GanSystem to handle GAN training.
It also reproduces SEGAN https://arxiv.org/pdf/1703.09452 a GAN for speech enhancement.
It adapts an already existing implementation of SEGAN on pytorch https://github.com/dansuh17/segan-pytorch

This PR is meant to open a discussion about the way that GanSystem should be implemented.

The recipe is not finished yet. Especially, SEGAN was implemented to handle signals with a very specific shape (16384 samples) and we need to figure out a proper way to handle larger signals.

For now the models have been trained on LibriMix train 360 with task enh_single. They also been tested on LibriMix. The results show an improvement of around 6 db in SNR against around 10 for ConvTasNet


mpariente:
Thanks ! 
I didn't look at the code yet but I like the principle. 
Could we train ConvTasNet in an adversarial way as well (just replacing the SEGAN archicture by the ConvTasNet one, omitting the z for now) to have a better idea of the role the adversarial training can play? 

That would be very informative ! 

Also, can we match the performance reported in the papers on the datasets used in the papers?

JorisCos:
I did some experiments on SEGAN.
I removed the random noise z that is concatenated to the encoded representation of the noisy signal : c. To keep the same architecture I concatenated c with itself. I obtained similar results with and without z. This suggests that the network learns to ignore z. This has already been reported in few papers about SEGAN.

mpariente:
Thanks for the feedback, we could also replace the supervised loss in SEGAN with whichever loss function, like SI-SNR right?

JorisCos:
Yes that's the plan.

egs/SEGAN/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Results
+ 
+All the models were trained on [LibriMix](../librimix) train-360 with task enh_single.
+
+|  Generator loss | SI-SNRi(dB) | SI-SNR(dB) | STOI | STOIi | PESQ | PESQi |
+|:---------------:|:-----------:|:----------:|:----:|:-----:|:----:|:-----:|
+| ConvTasnet      | 10.5        | 13.9        |0.92 |0.12   | 2.1  |  0.91 |
+| Adversarial + L1| 4.2         | 7.66        |0.83 |0.03   | 1.4  |  0.21 |
+|  L1             | 4.1         | 7.59        |0.83  |0.03  |1.39  |  0.23 |
```
jonashaag, line 30: Math seems wrong here :) PESQi should be 0.2

JorisCos: True ! Thanks, I made a wrong rounding.

mpariente:
I don't know what to do with those three PRs. #163 #152 #187 
Maybe they could go in a separate repo because I don't think it fits with the source code, not sure.

JorisCos:
Yes it probably doesn't fit and it remains mostly experimental. I could create a repo and upload the work there. Let me know : )

mpariente:
Cool, we agree. 
So, we won't merge this, but this is still interesting to have somewhere. Having it in a separate repo would make sense IMO, we can link to it from the docs.

JorisCos:
A question remains...
What will be the name of this mysterious repo ?

mpariente:
asteroid-gan-exps or something like that? 
I close the three PRs and we'll add a link here when the repo is open

PR #156, opened 2020-06-22
Slicer and an unslicer

JorisCos:
### About this PR
This PR introduce a slicer and an unslicer to slice signals, feed the sliced signals to a model and reconstruct the unsliced signal afterwards. This could be useful for very long signals that may cause code to break and also allows models (like SEGAN) that take a precise input length to handle longer signals.

### Proposed slicer
The proposed slicer uses hanning window and can handle overlap between 0 to 50 %. I did not find an elegant way to implement a slicer where more than two chunks are overlapping at the same time. Small numerical errors are introduced during the reconstruction phase they are around 1E-7.

### DIFF
--- a/asteroid/utils/torch_utils.py
+++ b/asteroid/utils/torch_utils.py
@@ -125,23 +125,23 @@ def are_models_equal(model1, model2):
     return True
 
 
-def slice(sources, w_size=16384, h_size=16384 // 2):
+def slice(sources, w_size=16384, stride=16384 // 2):
     """  Slice a signal in slices of size w_size
 
      Args:
          sources (torch.Tensor): sources of shape [nb_sources,channels,samples]
          w_size (int): size of the window
-         h_size (int): size of the stride
+         stride (int): size of the stride
      Returns:
          sliced signal (torch.Tensor): shape [nb_sources,nb_slices,channels,
                                               samples]
      """
     # stride must be shorter than window
-    if h_size < w_size / 2:
+    if stride < w_size / 2:
         raise ValueError(
             f"You can't have a stride smaller than half the window")
     # stride must be shorter than window
-    if h_size > w_size:
+    if stride > w_size:
         raise ValueError(
             f"You can't have a stride larger than the window")
     # Get source length
@@ -150,18 +150,18 @@ def slice(sources, w_size=16384, h_size=16384 // 2):
     if len_s > w_size:
         padding = 0
         # Automatic padding
-        if (len_s - w_size) % h_size != 0:
+        if (len_s - w_size) % stride != 0:
             padded = torch.zeros((sources.size(0), sources.size(1),
                                   sources.size(2) + w_size - (len_s - w_size) %
-                                  h_size))
+                                  stride))
             padded[:, :, :len_s] = sources
             sources = padded
             len_s = sources.size(-1)
-        nb_slices = int((len_s - w_size) // h_size) + 1
-        sliced = torch.zeros((sources.size(0),  nb_slices, sources.size(1),
+        nb_slices = int((len_s - w_size) // stride) + 1
+        sliced = torch.zeros((sources.size(0), nb_slices, sources.size(1),
                               w_size))
         for j in range(nb_slices):
-            sliced[:, j, :, :] = sources[:, :, j * h_size: j * h_size + w_size]
+            sliced[:, j, :, :] = sources[:, :, j * stride: j * stride + w_size]
         return sliced
     # If source is smaller then pad it to w_size
     if len_s < w_size:
@@ -173,7 +173,7 @@ def slice(sources, w_size=16384, h_size=16384 // 2):
     return sources.unsqueeze(1)
 
 
-def unslice(sources, original_shape, w_size=16384, h_size=16384 // 2):
+def unslice(sources, w_size=16384, stride=16384 // 2,original_shape=None):
     """  Reconstruct signal from sliced signal
 
      Args:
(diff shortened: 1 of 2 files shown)


tests/utils/torch_utils_test.py
```diff
@@ -36,9 +36,11 @@ def test_loader_module():
 def test_loader_submodule():
     class SuperModule(nn.Module):
         """ nn.Module subclass that holds a model under self.whoever """
+
```
mpariente, line 39: No need to rewrite parts of the code that the PR doesn't address.

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +123,107 @@ def are_models_equal(model1, model2):
...
+                                                    * w_up
+    # Cut unwanted padding
+    reconstructed = reconstructed[:, :, :original_shape[-1]]
+    return reconstructed
+
+import torch
+x = torch.randn(10, 1, 16384)
+sliced = slice(x)
+unsliced = unslice(sliced, x.shape)
+assert torch.mean(x-unsliced) < 1E-6
```
mpariente, line 229: Need new line please.

mpariente, line 229: This should be removed right?

tests/utils/torch_utils_test.py
```diff
@@ -53,3 +55,13 @@ def __init__(self, sub_model):
...
     model_2 = torch_utils.load_state_dict_in(state_dict, model_2)
     assert torch_utils.are_models_equal(model, model_2)
+
+
+def test_slicer():
+    x = torch.randn(10, 2, 16384)
+    sliced = torch_utils.slice(x)
+    unsliced = torch_utils.unslice(sliced, x.shape)
+    result = False
+    if torch.mean(x - unsliced) < 1E-6:
```
mpariente, line 65:
```python
from torch.testing import assert_allclose
assert_allclose(x, unsliced, atol=..., rtol=...)
```

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +123,107 @@ def are_models_equal(model1, model2):
...
     return True
+
+
+def slice(sources, w_size=16384, h_size=16384 // 2):
+    """  Slice a signal in slices of size w_size
+
+     Args:
+         sources (torch.Tensor): sources of shape [nb_sources,channels,samples]
+         w_size (int): size of the window
+         h_size (int): size of the stride
```
mpariente, line 134: I would call this stride though.

tests/utils/torch_utils_test.py
```diff
@@ -53,3 +55,13 @@ def __init__(self, sub_model):
     # Apply our workaround torch_utils.load_state_dict_in and assert True.
     model_2 = torch_utils.load_state_dict_in(state_dict, model_2)
     assert torch_utils.are_models_equal(model, model_2)
+
+
+def test_slicer():
+    x = torch.randn(10, 2, 16384)
```
mpariente, line 61: Hmm, this doesn't need to be sliced no? 
So it actually doesn't unit test, or I'm missing something? 
Can we test with other shapes

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +123,107 @@ def are_models_equal(model1, model2):
...
+    if len_s < w_size:
+        new_size = list(sources.size()[:-1]) + [w_size]
+        padded = torch.zeros(new_size)
+        padded[:, :, :len_s] = sources
+        return padded.unsqueeze(1)
+    # Else source equal window just unsqueeze
+    return sources.unsqueeze(1)
+
+
+def unslice(sources, original_shape, w_size=16384, h_size=16384 // 2):
```
mpariente, line 176: Because we pad in the slicing, we should have a target_len argument to remove the zeros we added at the end right?

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +123,107 @@ def are_models_equal(model1, model2):
...
+
+def slice(sources, w_size=16384, h_size=16384 // 2):
+    """  Slice a signal in slices of size w_size
+
+     Args:
+         sources (torch.Tensor): sources of shape [nb_sources,channels,samples]
+         w_size (int): size of the window
+         h_size (int): size of the stride
+     Returns:
+         sliced signal (torch.Tensor): shape [nb_sources,nb_slices,channels,
```
mpariente, line 136: At inference, we are expecting signals with shape (1, 1, wal_len), (1, wav_len) or (wav_len,) but we might not want to make the assumption the shape is always (*, *, time).

mpariente: A recursive solution would work. Or a class interface that handles the reshaping.

mpariente: If we want to support several input shapes, their inversion should be unique (unslice(slice(x)) should equal x).

asteroid/utils/torch_utils.py
JorisCos: Yes

asteroid/utils/torch_utils.py
JorisCos: Yes, that's what original_shape arg in unslice is. However, we can't remove the padding in the output of `slice() `otherwise the slicing method is unsuitable for models that only take a specific shape as input.

asteroid/utils/torch_utils.py
JorisCos: I think i'm going to try the class interface.

tests/utils/torch_utils_test.py
JorisCos: No you are right

asteroid/utils/torch_utils.py
mpariente: Oh yes, sorry. 
Then this should be a kwarg with None as default. 
If None, we don't remove any padding.

mpariente:
Closing in favor of #191

PR #160, opened 2020-06-29
Fix clipping problems when saving estimate wav file for Wham ConvTasNet

tux-coder:
- based on PR #148 made by Mathieu for librimix :
https://github.com/mpariente/asteroid/pull/148/files


mpariente:
Looks good thanks !

PR #163, opened 2020-07-06
Add gan_system.py

JorisCos:
This PR separates gan_system.py from the SEGAN recipe.
The code has been changed to provide a more general approach of GAN's training

### DIFF
--- a/asteroid/engine/gan_system.py
+++ b/asteroid/engine/gan_system.py
@@ -1,10 +1,5 @@
-from collections import OrderedDict
-import soundfile as sf
 import pytorch_lightning as pl
 import torch
-import os
-from torch_stoi import NegSTOILoss
-from asteroid.losses import PITLossWrapper
 
 
 class GanSystem(pl.LightningModule):
@@ -36,8 +31,8 @@ class GanSystem(pl.LightningModule):
 
     def __init__(self, discriminator, generator, opt_d, opt_g,
                  discriminator_loss, generator_loss, train_loader,
-                 validation_loss=PITLossWrapper(NegSTOILoss(), pit_from='pw_pt'),
-                 val_loader=None, scheduler_d=None, scheduler_g=None,
+                 validation_loss=None, val_loader=None,
+                 scheduler_d=None, scheduler_g=None,
                  conf=None):
 
         super(GanSystem, self).__init__()
@@ -53,49 +48,50 @@ class GanSystem(pl.LightningModule):
         self.scheduler_d = scheduler_d
         self.scheduler_g = scheduler_g
         self.conf = conf
-        self.estimates = None
 
-    def forward(self, z):
-        return self.generator(z)
+    def forward(self, *args, **kwargs):
+        return self.generator(*args, **kwargs)
 
     def training_step(self, batch, batch_nb, optimizer_idx):
-        # Get data from data_loader
-        inputs, targets = batch
-        # Forward inputs
-        self.estimates = self(inputs)
-        # Train discriminator
-        if optimizer_idx == 0:
-            # Compute D loss for targets
-            est_true_labels = self.discriminator(targets, inputs, targets)
-            true_loss = self.d_loss(inputs, targets, self.estimates,
-                                    est_true_labels, True)
-            # Compute D loss for self.estimates
-            est_false_labels = self.discriminator(self.estimates.detach(),
-                                                  inputs, targets)
-            fake_loss = self.d_loss(inputs, targets, self.estimates,
-                                    est_false_labels, False)
-            # Overall, the loss is the mean of these
-            d_loss = (true_loss + fake_loss) * 0.5
-            tqdm_dict = {'d_loss': d_loss}
-            output = OrderedDict({
-                'loss': d_loss,
-                'progress_bar': tqdm_dict,
-                'log': tqdm_dict
-            })
-            return output
 
-        # Train generator
-        if optimizer_idx == 1:
-            # The generator is supposed to fool the discriminator.
-            est_labels = self.discriminator(self.estimates, inputs, targets)
-            adversarial_loss = self.g_loss(self.estimates, targets, est_labels)
-            tqdm_dict = {'g_loss': adversarial_loss}
-            output = OrderedDict({
-                'loss': adversarial_loss,
-                'progress_bar': tqdm_dict,
-                'log': tqdm_dict
-            })
-            return output
+        raise NotImplementedError
+        # # Get data from data_loader
+        # inputs, targets = batch
+        # # Forward inputs
+        # self.estimates = self(inputs)
+        # # Train discriminator
+        # if optimizer_idx == 0:
+        #     # Compute D loss for targets
+        #     est_true_labels = self.discriminator(targets, inputs, targets)
+        #     true_loss = self.d_loss(inputs, targets, self.estimates,
+        #                             est_true_labels, True)
+        #     # Compute D loss for self.estimates
+        #     est_false_labels = self.discriminator(self.estimates.detach(),
+        #                                           inputs, targets)
+        #     fake_loss = self.d_loss(inputs, targets, self.estimates,
+        #                             est_false_labels, False)
+        #     # Overall, the loss is the mean of these
+        #     d_loss = (true_loss + fake_loss) * 0.5
+        #     tqdm_dict = {'d_loss': d_loss}
+        #     output = OrderedDict({
+        #         'loss': d_loss,
+        #         'progress_bar': tqdm_dict,
+        #         'log': tqdm_dict
+        #     })
+        #     return output
+        #
+        # # Train generator
+        # if optimizer_idx == 1:
+        #     # The generator is supposed to fool the discriminator.
+        #     est_labels = self.discriminator(self.estimates, inputs, targets)
+        #     adversarial_loss = self.g_loss(self.estimates, targets, est_labels)
+        #     tqdm_dict = {'g_loss': adversarial_loss}
+        #     output = OrderedDict({
+        #         'loss': adversarial_loss,
+        #         'progress_bar': tqdm_dict,
+        #         'log': tqdm_dict
+        #     })
+        #     return output
 
     def validation_step(self, batch, batch_nb):
         """ Need to overwrite PL validation_step to do validation.
--- a/asteroid/engine/gan_system.py
+++ b/asteroid/engine/gan_system.py
@@ -53,84 +53,11 @@ class GanSystem(pl.LightningModule):
         return self.generator(*args, **kwargs)
 
     def training_step(self, batch, batch_nb, optimizer_idx):
-
         raise NotImplementedError
-        # # Get data from data_loader
-        # inputs, targets = batch
-        # # Forward inputs
-        # self.estimates = self(inputs)
-        # # Train discriminator
-        # if optimizer_idx == 0:
-        #     # Compute D loss for targets
-        #     est_true_labels = self.discriminator(targets, inputs, targets)
-        #     true_loss = self.d_loss(inputs, targets, self.estimates,
-        #                             est_true_labels, True)
-        #     # Compute D loss for self.estimates
-        #     est_false_labels = self.discriminator(self.estimates.detach(),
-        #                                           inputs, targets)
-        #     fake_loss = self.d_loss(inputs, targets, self.estimates,
-        #                             est_false_labels, False)
-        #     # Overall, the loss is the mean of these
-        #     d_loss = (true_loss + fake_loss) * 0.5
-        #     tqdm_dict = {'d_loss': d_loss}
-        #     output = OrderedDict({
-        #         'loss': d_loss,
-        #         'progress_bar': tqdm_dict,
-        #         'log': tqdm_dict
-        #     })
-        #     return output
-        #
-        # # Train generator
-        # if optimizer_idx == 1:
-        #     # The generator is supposed to fool the discriminator.
-        #     est_labels = self.discriminator(self.estimates, inputs, targets)
-        #     adversarial_loss = self.g_loss(self.estimates, targets, est_labels)
-        #     tqdm_dict = {'g_loss': adversarial_loss}
-        #     output = OrderedDict({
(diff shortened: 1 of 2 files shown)


asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
+from collections import OrderedDict
+import soundfile as sf
+import pytorch_lightning as pl
+import torch
+import os
+
+
+class GanSystem(pl.LightningModule):
+
```
mpariente, line 9: Need docstring

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(self, discriminator, generator, opt_d, opt_g,
+                 scheduler_d, scheduler_g, discriminator_loss, generator_loss,
+                 validation_loss, train_loader, val_loader=None, conf=None):
```
mpariente, line 12: I don't think the schedulers should be positional arguments (meaning mandatory), they should be keyword args with None as default. 
Is there a good default value for validation loss or is it mandatory? I think we can make it a kwarg as well.

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+    def on_batch_end(self):
+        """ Overwrite if needed. Called by pytorch-lightning"""
+        pass
+
+    def on_epoch_start(self):
+        """ Overwrite if needed. Called by pytorch-lightning"""
+        pass
+
+
+    @pl.data_loader
```
mpariente, line 155: The decorator is not needed anymore.

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+        self.train_loader = train_loader
+        self.val_loader = val_loader
+        self.scheduler_d = scheduler_d
+        self.scheduler_g = scheduler_g
+        self.conf = conf
+
+    def forward(self, z):
+        return self.generator(z)
+
+    def routine_step(self, batch, optimizer_idx):
```
mpariente, line 31: If routine_step is the same as training_step, what's the goal of routine step?

mpariente: I wonder if it wouldn't be beneficial to separate `generator_training_step`, `discriminator_training_step` or it's not generic enough?

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+        self.g_loss = generator_loss
+        self.validation_loss = validation_loss
+        self.train_loader = train_loader
+        self.val_loader = val_loader
+        self.scheduler_d = scheduler_d
+        self.scheduler_g = scheduler_g
+        self.conf = conf
+
+    def forward(self, z):
+        return self.generator(z)
```
mpariente, line 29: I would have *args, **kwargs here. It would be more generic for conditional GANs for example

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+        if optimizer_idx == 0:
+            # Compute D loss for targets (labels = 1)
+            est_true_labels = self.discriminator(targets, inputs, targets)
+            true_loss = self.d_loss(inputs, targets, estimates,
+                                    est_true_labels, true_labels)
+            # Compute D loss for estimates (labels = 0)
+            est_false_labels = self.discriminator(estimates.detach(), inputs,
+                                                  targets)
+            fake_loss = self.d_loss(inputs, targets, estimates,
+                                    est_false_labels, false_labels)
```
mpariente, line 51: I don't like that many positional arguments, it's very error-prone. Two or three would be a maximum, five is too much for me.

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+            # Overall, the loss is the mean of these
+            d_loss = (true_loss + fake_loss) * 0.5
+            tqdm_dict = {'d_loss': d_loss}
+            output = OrderedDict({
```
mpariente, line 55: Curiosity, why is it ordered?

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def routine_step(self, batch, optimizer_idx):
+
+        # Get data from data_loader
+        inputs, targets = batch
+        # Forward inputs
+        estimates = self(inputs)
+        # The discriminator is basically a binary classifier
+        true_labels = True
+        false_labels = False
```
mpariente, line 39: Why defining them here and not using True and False? For soft labels? Something else I don't see?

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,161 @@
...
+            batch: the object returned by the loader (a list of torch.Tensor
+                in most cases) but can be something else.
+            batch_nb (int): The number of the batch in the epoch.
+
+        Returns:
+            dict:
+
+            ``'val_loss'``: loss
+        """
+        inputs, targets = batch
```
mpariente, line 93: By default, don't we want d_loss and g_loss at validation time? 
I'm no GAN expert, just asking.

asteroid/engine/gan_system.py
JorisCos: I guess stoi is the best choice.

asteroid/engine/gan_system.py
JorisCos: This is a legacy from the previous version and should be in the training step.
I find it simpler to have all the training in the training step but it doesn't cost much to create `generator_training_step` `discriminator_training_step` as you wish.

asteroid/engine/gan_system.py
JorisCos: I don't understand how we can pass more arguments to the generator anyway since the call is done in `gan_system` and is never to be changed by the user or am I missing something ?

asteroid/engine/gan_system.py
JorisCos: This was taken from the PL docs,

asteroid/engine/gan_system.py
JorisCos: No you are right this useful when don't separate the generator loss and the discriminator loss but we do.

asteroid/engine/gan_system.py
JorisCos: In the validation step we only pass data through the generator and evaluate its performance.
The value of d_loss and g_loss is not very relevant at this point because

asteroid/engine/gan_system.py
JorisCos: I understand your point but I don't know how I can do a generic call to d_loss with less positional args it's the same problem as my previous comment

tests/engine/gan_system_test.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__init__()
+
+    def forward(self, estimates, targets, est_labels):
+        loss = torch.mean((est_labels - torch.ones_like(est_labels)) ** 2)
+        return loss
+
+
+class DiscriminatorLoss(_Loss):
+    def __init__(self):
+        super().__init__()
```
mpariente, line 34: Those lines are not needed. 
If you don't subclass a method, it stays the same. 
Same above

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,172 @@
...
+        self.estimates = None
+
+    def forward(self, z):
+        return self.generator(z)
+
+    def training_step(self, batch, batch_nb, optimizer_idx):
+        # Get data from data_loader
+        inputs, targets = batch
+        # Forward inputs
+        self.estimates = self(inputs)
```
mpariente, line 65: Why do you need `estimates` as an attribute?

mpariente: Also, I don't know how well this will behave when we train on multi GPU.

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,172 @@
...
+        # Train discriminator
+        if optimizer_idx == 0:
+            # Compute D loss for targets
+            est_true_labels = self.discriminator(targets, inputs, targets)
```
mpariente, line 69: Is this discriminator call completely generic?

mpariente:
@popcornell Can you have a a look please?

asteroid/engine/gan_system.py
mpariente: Because what?

asteroid/engine/gan_system.py
mpariente: Different GANs will have different `nn.Module` as generator. 
```suggestion
    def forward(self, *args, **kwargs):
        return self.generator(*args, **kwargs)
```
This is just more generic.

asteroid/engine/gan_system.py
```diff
@@ -0,0 +1,172 @@
...
+               of learning rate schedulers for d.
+           scheduler_g (torch.optim.lr_scheduler._LRScheduler): Instance
+               of learning rate schedulers for g.
+           conf: Anything to be saved with the checkpoints during training.
+               The config dictionary to re-instantiate the run for example.
+       """
+
+    def __init__(self, discriminator, generator, opt_d, opt_g,
+                 discriminator_loss, generator_loss, train_loader,
+                 validation_loss=PITLossWrapper(NegSTOILoss(), pit_from='pw_pt'),
```
mpariente, line 39: I would put `None` as the validation loss, and we don't run validation step if we don't have validation loss. 
Imposing STOI doesn't seem right to me.

asteroid/engine/gan_system.py
JorisCos: Ahah sry. Because they don't give information about the performance, and they will converge to a value after few iterations, so I don't think it's worth the computation

asteroid/engine/gan_system.py
JorisCos: This was done in case someone wants to use estimates during  on_epoch_ends or on_training_end etc to save examples or see the evolution of one estimates.

asteroid/engine/gan_system.py
JorisCos: I think so, we basically give everything to the network in case someone need the true ref or the noisy ref

mpariente: But this is written nowhere, so the user has to come here to find what the `GanSystem` expects of his network. If he comes this far to read the code, he can re-implement his own System. 
I'm not sure this will be useful. It will either require huge amount of docs or be too simplistic. 
And the code is not so because most args are positional.. 
Maybe it's useful to have the backbone with the optimizer, the scheduler etc.. so that it doesn't have to be rewritten. However, the training step should either be super simple, or `NotImplementedError` IMO. @popcornell any thoughts on that? You know GAN way better than me.

asteroid/engine/gan_system.py
popcornell: but wouldn't be better to save only validation examples ?

asteroid/engine/gan_system.py
popcornell: why it does take 3 arguments with two arguments identical ? Plain discriminator should have one argument (one time is the enhanced signal another time is the true clean signal). Are you using relativistic discriminator ?

popcornell: > Maybe it's useful to have the backbone with the optimizer, the scheduler etc.. so that it doesn't have to be rewritten. However, the training step should either be super simple, or NotImplementedError IMO. @popcornell any thoughts on that? You know GAN way better than me.

GANs are quite sensitive and difficult to train.... plus there are tons of different types some differs only for the output activation of discriminator and loss function but others (like relativistic discriminator ) requires changes to the training loop.  Also because are hard to train one usually logs everything to tensorboard in the training loop. I mean it is nice to have a GAN example and maybe a GAN system. But for real research it will be bypassed probably.

popcornell: I will actually like to do some barebone system which bypasses most of lightning functionality (also grad accumulation). For example one could wish to have different virtual batches for generator and discriminator ( now with wasserstein gan and such is less needed but still...) . Also useful for other techniques which require full control over the training loop.

mpariente: Thanks for having a look !

mpariente:
Well IDK @JorisCos, maybe have the "complicated" (hear varying from one arch to another) steps stay as `NotImplementedError` so that it's clear we don't take care of that?

asteroid/engine/gan_system.py
JorisCos: We don't make it mandatory to have a validation step, but we always have a training step.
`GanSystem `could take an arg that specifies if we want to save examples `on_training_end` or `on_epoch_end` but we don't do it in `System `I think I will just remove the attribute.

asteroid/engine/gan_system.py
JorisCos: > why it does take 3 arguments with two arguments identical ? Plain discriminator should have one argument (one time is the enhanced signal another time is the true clean signal). Are you using relativistic discriminator ?

Sometimes the discriminator requires more than just the enhanced signal or the true clean signal. Some discriminator actually compare the enhanced signal to the true Signal ( ex MetricGAN)  some discriminator use the enhanced or the clean signal + the noisy input (SEGAN). This is why I made the discriminator have the args.

JorisCos:
> Well IDK @JorisCos, maybe have the "complicated" (hear varying from one arch to another) steps stay as `NotImplementedError` so that it's clear we don't take care of that?

Just to be sure we agree on what to do next. I will go back to the first version of `GanSystem`  that had a `routine_step` which is specific to each GAN and that implements the training loop ? Which mean that for each new GAN based recipe, the  `routine_step` will have do be edited.

asteroid/engine/gan_system.py
mpariente: For the current use cases, this choice can make sense. But the audio GAN landscape is so broad that we cannot aim at covering it all..

mpariente:
I'd stick with:
```python
def training_step(etc, ...):
    raise NotImplementedError
```

mpariente:
Now that this is used in both SEGAN and metricGAN, I think we can merge this. 
Then, we'll get back to MetricGAN and SEGAN.

Issue #165, opened 2020-07-09
Bugs of reloading model


hangtingchen:
## 🐛 Bug

Bugs of reloading model when best_k_models.json not exist

### To Reproduce
https://github.com/mpariente/asteroid/blob/0bdec2644f2d770d037ce804b7f70cb98bd5c9fa/egs/wsj0-mix/DeepClustering/model.py#L156-L157

The lines use sort to get the last model, which will performs in a incorrect way  in the following situation
```
>>> all_ckpt=['ckpt_epoch_99.ckpt','ckpt_epoch_100.ckpt','ckpt_epoch_101.ckpt']
>>> all_ckpt.sort()
>>> all_ckpt[-1]
'ckpt_epoch_99.ckpt'
```

### Expected behavior

Maybe we can use the following methods:
```
>>> all_ckpt=[(ckpt,int("".join(filter(str.isdigit,os.path.basename(ckpt))))) for ckpt in all_ckpt if ckpt.find('ckpt')>=0]
>>> all_ckpt.sort(key=lambda x:x[1])
>>> all_ckpt[-1][0]
'ckpt_epoch_101.ckpt'
```

### Environment

 - Python 3.7

mpariente:
You're completely right, mind submitting a PR please?

mpariente:
Solved by #166

Issue #164, opened 2020-07-09
Bugs in generating wsj0-2mix dataset


hangtingchen:
## 🐛 Bug

Should use mv1 instead of mv2 to get wav files

### To Reproduce

https://github.com/mpariente/asteroid/blob/0bdec2644f2d770d037ce804b7f70cb98bd5c9fa/egs/wsj0-mix/DeepClustering/local/convert_sphere2wav.sh#L31

The line uses both mv2 and mv1 to get wav files. But the mv1 will be covered by mv2, resulting in the generated wav files being from mv2. The mv1 is noise-free while the mv2 is noisy. The wsj0-mix dataset is expected to use mv1.

The correct code is 
``
wav=`echo "$line" | sed "s:wv1:wav:g" | awk -v dir=$wav_dir -F'/' '{printf("%s/%s/%s/%s", dir, $(NF-2), $(NF-1), $NF)}'`
``

### Expected behavior

We tested the datasets generated by mv1 and mv2. It is observed that the former can reproduce the results, the latter is worse around 1-2 dB in SI-SNR.

Our results with mv1, the final validation loss was about 2950.
![微信截图_20200709163058](https://user-images.githubusercontent.com/30209378/87016695-a065bf80-c201-11ea-955d-3c603cdf6ac0.png)

I am sorry that our results with mv2 were deleted, its final validation loss was about 3500.

### Environment

 - Asteroid-master
 - PyTorch 1.4.0
 - PyTorchLightning 7.6.1

mpariente:
I think this is right and has been reported elsewhere as well. 
I think I only had wv1 when I generated wsj0 in the first place so I didn't notice the problem. 
I checked now and you're right. 

Would you like to submit a PR for that please? 
Thanks again !

mpariente:
Closed by #166

Issue #138, opened 2020-06-03
The poor performance of DeepClustering 


hangtingchen:
Hi
First thanks a lot for such an excellent tool for speech separation. I have tried the deep clustering part of wsj0-mix
[https://github.com/mpariente/asteroid/tree/master/egs/wsj0-mix/DeepClustering](url)
My performance was poor (si-sdr=3.5, sdr=4.5 in 35 epochs with 1 gpu for training). As reported here, the sdr is expected to be closed to 10dB. I am wondering the reason of the failure. Is there any tricks for training, or more epochs are needed for improvement?

Thanks a lot.

mpariente:
Hey, thanks for opening an issue. 
Let me paste my training config here: 
```yaml
data:
  n_src: 2
  sample_rate: 8000
  train_dir: data/2speakers/wav8k/min/tr
  valid_dir: data/2speakers/wav8k/min/cv
filterbank:
  kernel_size: 256
  n_filters: 256
  stride: 64
main_args:
  exp_dir: exp/train_chimera_dcalone_newlr/
  help: null
masknet:
  dropout: 0.3
  embedding_dim: 40
  hidden_size: 600
  n_layers: 4
  n_src: 2
  rnn_type: lstm
  take_log: true
optim:
  lr: 0.0001
  optimizer: adam
  weight_decay: 0.0
positional arguments: {}
training:
  batch_size: 32
  early_stop: true
  epochs: 200
  half_lr: true
  loss_alpha: 1.0
  num_workers: 8
```
Most importantly, the learning rate is `1e-4` instead of `1e-5` in the recipe. I'll change the default now. 

Here are the metrics 
```json
{
"si_sdr": 9.846718255569847,
"si_sdr_imp": 9.84787033402749,
"sdr": 10.364474047942608,
"sdr_imp": 10.213430163640966,
"sir": 19.018816263769104,
"sir_imp": 18.867772082693932,
"sar": 11.336185832787844,
"sar_imp": -64.43806377592414,
"stoi": 0.8787521784931114,
"stoi_imp": 0.1407063969268085
}
```

My best val loss was around 2930 and I trained for 130 epochs. 

I you come over slack, I can share the pretrained model folder (under research only license) to you.

hangtingchen:
thanks a lot, I am to try this config. The results may come out tomorrow. If it is fine,  I will close this issue. One more question, 
[...]

PR #166, opened 2020-07-09
fix bugs in generating wsj0-mix dataset with wv1 && loading model acc…

hangtingchen:
fix bugs in generating wsj0-mix dataset with wv1 instead of wv2 && loading model according to the epoch number

### DIFF
--- a/egs/wsj0-mix/DeepClustering/model.py
+++ b/egs/wsj0-mix/DeepClustering/model.py
@@ -153,8 +153,11 @@ def load_best_model(train_conf, exp_dir):
     except FileNotFoundError:
         # Get last checkpoint
         all_ckpt = os.listdir(os.path.join(exp_dir, 'checkpoints/'))
-        all_ckpt.sort()
-        best_model_path = os.path.join(exp_dir, 'checkpoints', all_ckpt[-1])
+        all_ckpt = [(ckpt, int("".join(filter(str.isdigit, \
+                    os.path.basename(ckpt))))) \
+                    for ckpt in all_ckpt if ckpt.find('ckpt') >= 0]
+        all_ckpt.sort(key=lambda x: x[1])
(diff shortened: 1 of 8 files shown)


mpariente:
Thanks a lot for the fix !
Next time, please make two separated PRs as those are two different problems, it makes easier to track changes afterwards.

egs/wsj0-mix/DeepClustering/model.py
```diff
@@ -153,8 +153,11 @@ def load_best_model(train_conf, exp_dir):
     except FileNotFoundError:
         # Get last checkpoint
         all_ckpt = os.listdir(os.path.join(exp_dir, 'checkpoints/'))
-        all_ckpt.sort()
-        best_model_path = os.path.join(exp_dir, 'checkpoints', all_ckpt[-1])
+        all_ckpt = [(ckpt, int("".join(filter(str.isdigit, \
```
mpariente, line 156: You don't need the backslash.

egs/wsj0-mix/DeepClustering/model.py
```diff
@@ -153,8 +153,11 @@ def load_best_model(train_conf, exp_dir):
...
+                    os.path.basename(ckpt))))) \
```
mpariente, line 157: This should be aligned with the `filter` starting parenthesis.

mpariente:
This will solve #164 and also #165 and the use of wv2 explained the poor performance in #138.

hangtingchen:
> This will solve #164 and also #165 and the use of wv2 explained the poor performance in #138.

We have updated the code.

mpariente:
Thanks ! 
The formatting is still not ok and I think you disabled "edit from the maintainers", I'll merge and fix it.

hangtingchen:
Thanks. Any guide for the coding style?

mpariente:
https://github.com/psf/black

PR #167, opened 2020-07-09
Update prepare_data.sh

subhanjansaha:
This is the correct directory structure


mpariente:
Thanks very for the PR. 
This is right, sorry! 
We should unify the downloading script accross wham recipes, they are diverging for no reason..

PR #170, opened 2020-07-20
Feat wise global layernorm

popcornell:
Feature-wise global layer normalization as it is used in TCN++ in FUSS and firstly proposed in Universal Sound Separation. 
Normalization is applied over all frames for each channel.


mpariente:
Thanks for the PR, that looks good ! 
Maybe we can add a comment in the docstring that it comes from TCN+ in the USS paper?

PR #179, opened 2020-07-23
adds Dataset for SmsWsj

jensheit:
This pull request adds a Dataset for the SMS_WSJ database. This is a first step in implementing a neural network based example.

### DIFF
--- a/asteroid/data/sms_wsj_dataset.py
+++ b/asteroid/data/sms_wsj_dataset.py
@@ -1,0 +1,219 @@
+import torch
+from torch.utils import data
+import numpy as np
+import soundfile as sf
+import lazy_dataset
+from lazy_dataset.database import JsonDatabase
+EPS = 1e-8
+
+DATASET = 'SMS_WSJ'
+# SMS_WSJ targets
+sep_source = {'mixture': 'observation',
+             'target': ['speech_source'],
+             'infos': {'num_channels': 6},
+             'default_nsrc': 2}
+sep_early = {'mixture': 'observation',
+             'target': ['speech_reverberation_early'],
+             'infos': {'num_channels': 6},
+             'default_nsrc': 2}
+speech_image = ['speech_reverberation_early', 'speech_reverberation_tail']
+sep_image = {'mixture': 'observation',
+             'target': speech_image,
+             'infos': {'num_channels': 6},
+             'default_nsrc': 2}
+
+
+SMS_TARGETS = {'source': sep_source,
+              'early': sep_early,
+              'image': sep_image}
+
+
+def normalize_tensor_wav(wav_tensor, eps=1e-8, std=None):
+    mean = wav_tensor.mean(-1, keepdim=True)
+    if std is None:
+        std = wav_tensor.std(-1, keepdim=True)
+    return (wav_tensor - mean) / (std + eps)
+
+
+class SmsWsjDataset(data.Dataset):
+    """ Dataset class for SMS WSJ source separation.
+
+    Args:
+        json_path (str): The path to the sms_wsj json file.
+        target (str): One of ``'source'``, ``'early'`` or ``'image'``.
+
+            * ``'source'`` non reverberant clean targets signals.
+            * ``'early'`` early reverberation target signals.
+            * ``'image'`` reverberant target signals
+        dset (str): train_si284 for train, cv_dev93 for validation and
+                test_eval92 for test
+        sample_rate (int, optional): The sampling rate of the wav files.
+        segment (float, optional): Length of the segments used for training,
+            in seconds. If None, use full utterances (e.g. for test).
+        single_channel (bool): if True all channels are used if false only
+                    a random channel is used during training
+                    and the first channel during test
+        nondefault_nsrc (int, optional): Number of sources in the training
+            targets.
+        normalize_audio (bool): If True then both sources and the mixture are
+            normalized with the standard deviation of the mixture.
+    """
+    dataset_name = 'SMS_WSJ'
+
+    def __init__(self, json_path, target, dset, sample_rate=8000,
+                 single_channel=True, segment=4.0, nondefault_nsrc=None,
+                 normalize_audio=False):
+        try:
+            import sms_wsj
+        except:
+            import warnings
+            warnings.warn(
+                "Some of the functionality relies on the sms_wsj package "
+                "downloadable from https://github.com/fgnt/sms_wsj ."
+                "The user is encouraged to install the package"
+            )
+        super().__init__()
+        if target not in SMS_TARGETS.keys():
+            raise ValueError('Unexpected task {}, expected one of '
+                             '{}'.format(target, SMS_TARGETS.keys()))
+
+        # Task setting
+        self.json_path = json_path
+        self.target = target
+        self.target_dict = SMS_TARGETS[target]
+        self.single_channel = single_channel
+        self.sample_rate = sample_rate
+        self.normalize_audio = normalize_audio
+        self.seg_len = None if segment is None else int(segment * sample_rate)
+        if not nondefault_nsrc:
+            self.n_src = self.target_dict['default_nsrc']
+        else:
+            assert nondefault_nsrc >= self.target_dict['default_nsrc']
+            self.n_src = nondefault_nsrc
+        self.like_test = self.seg_len is None
+        self.dset = dset
+
+        # Load json files
+        db = JsonDatabase(json_path)
+        dataset = db.get_dataset(dset)
+        # Filter out short utterances only when segment is specified
+        orig_len = len(dataset)
+        drop_utt, drop_len = 0, 0
+        if not self.like_test:
+            def filter_short_examples(example):
+                num_samples = example['num_samples']['observation']
+                if num_samples < self.seg_len:
+                    return False
+                else:
+                    return True
+
+            dataset = dataset.filter(filter_short_examples, lazy=False)
+        self.dataset = dataset
+
+    def __add__(self, sms_wsj):
+        if self.n_src != sms_wsj.n_src:
+            raise ValueError('Only datasets having the same number of sources'
+                             'can be added together. Received '
+                             '{} and {}'.format(self.n_src, sms_wsj.n_src))
+        if self.seg_len != sms_wsj.seg_len:
+            self.seg_len = min(self.seg_len, sms_wsj.seg_len)
+            print('Segment length mismatched between the two Dataset'
+                  'passed one the smallest to the sum.')
+        self.dataset = self.dataset.concatenate(sms_wsj.dataset)
+
+    def __len__(self):
+        return len(self.dataset)
+
+    def __getitem__(self, idx):
+        """ Gets a mixture/sources pair.
+        Returns:
+            mixture, vstack([source_arrays])
+        """
+        # Random start
+        example = self.dataset[idx]
+        in_signal = self.target_dict['mixture']
+        target = self.target_dict['target']
+        audio_path = example['audio_path']
+        num_samples = example['num_samples']['observation']
+        if num_samples == self.seg_len or self.like_test:
+            rand_start = 0
+        else:
+            rand_start = np.random.randint(0, num_samples - self.seg_len)
+        if self.like_test:
+            stop = None
+        else:
+            stop = rand_start + self.seg_len
+        # Load mixture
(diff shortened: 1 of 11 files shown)


asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+            * ``'early'`` early reverberation target signals.
+            * ``'image'`` reverberant target signals
+        dset (str): train_si284 for train, cv_dev93 for validation and
+                test_eval92 for test
+        sample_rate (int, optional): The sampling rate of the wav files.
+        segment (float, optional): Length of the segments used for training,
+            in seconds. If None, use full utterances (e.g. for test).
+        single_channel (bool): if True all channels are used if false only
+                    a random channel is used during training
+                    and the first channel during test
```
mpariente, line 53: Is this correct? Or the opposite?

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Args:
+        json_path (str): The path to the sms_wsj json file.
+        target (str): One of ``'source'``, ``'early'`` or ``'image'``.
+
+            * ``'source'`` non reverberant clean targets signals.
+            * ``'early'`` early reverberation target signals.
+            * ``'image'`` reverberant target signals
+        dset (str): train_si284 for train, cv_dev93 for validation and
+                test_eval92 for test
```
mpariente, line 47: Why not `train`, `cv` and `test`?

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+    def __add__(self, sms_wsj):
+        if self.n_src != sms_wsj.n_src:
+            raise ValueError('Only datasets having the same number of sources'
+                             'can be added together. Received '
+                             '{} and {}'.format(self.n_src, sms_wsj.n_src))
+        if self.seg_len != sms_wsj.seg_len:
+            self.seg_len = min(self.seg_len, sms_wsj.seg_len)
+            print('Segment length mismatched between the two Dataset'
+                  'passed one the smallest to the sum.')
+        self.dataset = self.dataset.concatenate(sms_wsj.dataset)
```
mpariente, line 122: Genuine question: Do you think this is useful? 
I wrote it for WHAM where you can concatenate clean and noisy separation datasets. But here, the mixtures are always the same and the training targets are different, does it make sense to mix the training targets for one network? Or what other use case would you see?

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+                for t in target:
+                    if t == 'speech_source':
+                        start = 0
+                        stop_ = None
+                    else:
+                        start = rand_start
+                        stop_ = stop
+                    s_, _ = sf.read(audio_path[t][idx], start=start,
+                                    stop=stop_, dtype='float32')
+                    s += s_.T
```
mpariente, line 173: Clarification, this sum is to handle this case : ` speech_image = ['speech_reverberation_early', 'speech_reverberation_tail']` where the training target is the whole reverberated signal, is that right?

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+        if self.seg_len != sms_wsj.seg_len:
+            self.seg_len = min(self.seg_len, sms_wsj.seg_len)
+            print('Segment length mismatched between the two Dataset'
+                  'passed one the smallest to the sum.')
+        self.dataset = self.dataset.concatenate(sms_wsj.dataset)
+
+    def __len__(self):
+        return len(self.dataset)
+
+    def __getitem__(self, idx):
```
mpariente, line 127: I proofread this but mainly I trust you for that because I don't have access to WSJ1..

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+
+sms_wsj_license = dict(
+    title='SMS-WSJ: A database for in-depth analysis of multi-channel source separation algorithms',
+    title_link='https://github.com/fgnt/sms_wsj',
+    author='Department of Communications Engineering University of Paderborn',
+    author_link='https://github.com/fgnt',
+    license='MIT License',
+    license_link='https://github.com/fgnt/sms_wsj/blob/master/LICENSE',
+    non_commercial=False,
+)
```
mpariente, line 219: SMS_WSJ is derived from WSJ so the models cannot be shared under MIT, they have to be shared under CC by SA 3.0 wrt to the original WSJ0 license. 
The code is shared under MIT, but the data cannot be. 

This part is used to know which license to use for publishing pretrained models. 
In particular the `infos['licenses']`. 
Are you ok to add it?

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+    def get_infos(self):
+        """ Get dataset infos (for publishing models).
+
+        Returns:
+            dict, dataset infos with keys `dataset`, `task` and `target`.
+        """
+        infos = dict()
+        infos['dataset'] = self.dataset_name
+        infos['task_dataset'] = self.dset
+        infos['target'] = self.target
```
mpariente, line 207:
```suggestion
        infos['target'] = self.target
        infos['license'] = [wsj0_license, sms_wsj_license]
```

mpariente: This will be used to generate a License notice similar to the one on the bottom of [this page](https://zenodo.org/record/3874420#.XxnvuBF8Lg4)

mpariente: (Plus the license import in the top of the file)

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+speech_image = ['speech_reverberation_early', 'speech_reverberation_tail']
+sep_image = {'mixture': 'observation',
+             'target': speech_image,
+             'infos': {'num_channels': 6},
+             'default_nsrc': 2}
+
+
+SMS_TARGETS = {'source': sep_source,
+              'early': sep_early,
+              'image': sep_image}
```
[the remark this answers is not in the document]
jonashaag: How about including the black formatter in the CI pipeline? You can even automate this, e.g. with DeepSource.

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+
+        # Load json files
+
+        from lazy_dataset.database import JsonDatabase
+        db = JsonDatabase(json_path)
+        dataset = db.get_dataset(dset)
+        # Filter out short utterances only when segment is specified
+        orig_len = len(dataset)
+        drop_utt, drop_len = 0, 0
+        if not self.like_test:
```
jonashaag, line 102: I have also duplicated this code in my WIP Voice Bank + DEMAND impl. So now there are at least 3 duplicates of the code. We should probably merge dataset logic (maybe even preprocessing logic) somehow.

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+        Returns:
+            mixture, vstack([source_arrays])
+        """
+        # Random start
+        example = self.dataset[idx]
+        in_signal = self.target_dict['mixture']
+        target = self.target_dict['target']
+        audio_path = example['audio_path']
+        num_samples = example['num_samples']['observation']
+        if num_samples == self.seg_len or self.like_test:
```
jonashaag, line 138: This part is also duplicated in my VBD impl.

asteroid/data/sms_wsj_dataset.py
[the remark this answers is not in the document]
mpariente: Yes, that's possible, and we would gain time in reviews, but I don't think we're black-compliant yet. 
I'll add it to my todo but this won't be so soon, sorry.

asteroid/data/sms_wsj_dataset.py
mpariente: If you want to give it a try feel free. At first I didn't want to write abstract classes because there is no _right_ way to do a Dataset class, but if we use it in three places, we could abstract it. 
For the preprocessing, what do you mean exactly?

asteroid/data/sms_wsj_dataset.py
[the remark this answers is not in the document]
jonashaag: If you want me to help I can send a PR with a suggested CI pipeline.

asteroid/data/sms_wsj_dataset.py
jonashaag: The scripts that create json files. There's some shared code there.

asteroid/data/sms_wsj_dataset.py
jensheit: I added this mostly because it was part of the WHAM example and I wanted to add the same amount of customizability.
However, this is basically just one line, so the user can also do this on his own. I will delete it.

Since I am using the lazy_dataset package the user can change everything he wants to change by doing sms_wsj.dataset.concatenate/filter/map etc.

asteroid/data/sms_wsj_dataset.py
jensheit: yes that is right. To minimize the disk space required we only write down the signals which cannot easily be generated from other signals. Therefore, the whole reverberated signal has to recalculated. I will add a comment

asteroid/data/sms_wsj_dataset.py
jensheit: sure, I will add that. Our RIRs are published under MIT.

asteroid/data/sms_wsj_dataset.py
jensheit: I have added these lines and created a wsj1_license information since SMS_WSJ uses both wsj0 and wsj1

mpariente: Oh yes, true, prefect !

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,233 @@
...
+        """ Get dataset infos (for publishing models).
+
+        Returns:
+            dict, dataset infos with keys `dataset`, `task` and `target`.
+        """
+        infos = dict()
+        infos['dataset'] = self.dataset_name
+        infos['task_dataset'] = self.dset
+        infos['target'] = self.target
+        infos['license'] = [wsj0_license, wsj1_license, sms_wsj_license]
```
mpariente, line 210: Actually, it's licenses, sorry.
```suggestion
         infos['licenses'] = [wsj0_license, wsj1_license, sms_wsj_license]
```

asteroid/data/sms_wsj_dataset.py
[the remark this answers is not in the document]
mpariente: That would be great, thank you !

mpariente:
Why do the earlystopping PR appear?  
Did you merge upstream/master into your branch?

jensheit:
> Why do the earlystopping PR appear?
> Did you merge upstream/master into your branch?

Oh yes it looks like it, strange. If necessary, I can try to remove the merge from the pull request history.

asteroid/data/sms_wsj_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+            normalized with the standard deviation of the mixture.
+    """
+    dataset_name = 'SMS_WSJ'
+
+    def __init__(self, json_path, target, dset, sample_rate=8000,
+                 single_channel=True, segment=4.0, nondefault_nsrc=None,
+                 normalize_audio=False):
+        try:
+            import sms_wsj
+        except:
```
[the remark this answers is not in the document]
jensheit: done

asteroid/data/sms_wsj_dataset.py
jensheit: you are right of course, I fixed this

mpariente:
It's not necessary, thanks. 
Is it ready to merge on your end?

jensheit:
> It's not necessary, thanks.
> Is it ready to merge on your end?

yes, I am happy with the current status. I will start working on a DC example on SMS-WSJ as soon as the dataset is merged.

mpariente:
Merging then, thanks !

PR #182, opened 2020-07-24
Fix #181: Add set -o pipefail

### LINKED ISSUE #181: set -e not working?
In ConvTasNet `run.sh`, there's a `set -e` at the start of the file, so I'd expect the `run.sh` script to stop for instance if training has failed. But for me it always "falls through" to the next step, e.g. evaluation, which then fails because training hasn't completed.

jonashaag:


mpariente:
Thanks a lot !

mpariente:
I see now that it could be replaced by `set -eo pipefail`.. Sorry, I'm late to the party :smile:

PR #174, opened 2020-07-22
Add SuDoRM-RF model

jonashaag:
### Changes from original:

- De-duplicated code that's already in Asteroid and code that's duplicated in SuDoRM-RF

### Example training config:

```yaml
# Model config
# SuDoRM-RF 1.0x:
model:
  out_channels: 128
  in_channels: 512
  num_blocks: 16
  upsampling_depth: 4
  enc_kernel_size: 21
  enc_num_basis: 512
  num_sources: 1
# Best model:
#model:
  #out_channels: 512
  #in_channels: 512 (?)
  #num_blocks: 34
  #upsampling_depth: 4
  #enc_kernel_size: 21
  #enc_num_basis: 512 (?)
  #num_sources: 1
# Default improved
#model:
  #out_channels: 256
  #in_channels: 512
  #num_blocks: 16
  #upsampling_depth: 5
  #enc_kernel_size: 21
  #enc_num_basis: 512
  #num_sources: 1
# Training config
training:
  epochs: 200
  batch_size: 4
  num_workers: 4
  half_lr: yes
  early_stop: yes
  gradient_clipping: 5
# Optim config
optim:
  optimizer: adam
  lr: 0.001
# Data config
data:
  train_dir: data/wav8k/min/tr/
  valid_dir: data/wav8k/min/cv/
  task: sep_clean
  sample_rate: 8000
  mode: min
  segment: 3.0
```

PR #185, opened 2020-07-27
Update code for Sudo RM RF model

etzinis:
Please review the changes in this branch and make the appropriate changes in https://github.com/mpariente/asteroid/pull/174 
@mpariente @jonashaag


mpariente:
Thanks @etzinis, I'll let @jonashaag review this. 
We'll filter out the parts which are important because it's a bit of a mess currently TBH

The architecture looks the same and using the encoder/decoder from asteroid looks good. I also like to have the model readily available in `models`, that's going to be useful. 
Regarding the training, I didn't check in detail but it looked similar to ConvTasNet. 

One thing we can't do is to change `System` like that, this is limited to 2-speaker mixtures without noise, maybe we should find a bit better but thanks for this piece of code as well, it's informative. 

Thanks again

etzinis:
OK @mpariente, I agree that @jonashaag  code is better. I was just wondering whether we should add this explicit recipe as well but it is similar to convtasnet of course. 

For the augmentation part inside system, it is extremely easy to do this for all sources by using a for loop.

jonashaag:
Let’s not worry too much about the model code. I don’t think there’s much of a difference between our versions. I spent maybe 30 minutes on my changes, and I was mainly worried about deduplicating Code between “improved” and other version. Is there any value in including both models here, or only the improved one?

The System changes look good to me if we adapt them to N sources and put behind a flag. 

As for training scripts, I want to work on that soon so don’t worry too much about duplication there.

mpariente:
No worries @etzinis, I appreciate your time and contribution nonetheless !

>  Is there any value in including both models here, or only the improved one?
> 

I think having both doesn't hurt but maybe it's clearer to have only one, IDK. I'll let @etzinis decide. 

> The System changes look good to me if we adapt them to N sources and put behind a flag.
> 

I think this could be more appropriate as a `collate_fn`. 

> I was just wondering whether we should add this explicit recipe as well but it is similar to convtasnet of course.

Yes, we should definitely have a recipe for it ! This will give us more work when will refactor :joy: but we need it.

mpariente:
Closing this in favour of #174.
The model in integrated in `asteroid/models`, the online mixing as a collate function in `asteroid/data/utils.py` and we'll make the recipe after the refactor.

PR #187, opened 2020-07-29
MetricGAN

JorisCos:
### About this PR
This PR implements an other GAN for the speech enhancement task named [MetricGAN](https://arxiv.org/pdf/1905.04874.pdf).
The model only implements the STOI version of MetricGAN. 
It was trained on LibriMix train-360  min 16k subset. 

The results show slight better performance over SEGAN.

### DIFF
--- a/asteroid/data/metricGAN.py
+++ b/asteroid/data/metricGAN.py
@@ -1,0 +1,47 @@
+import numpy as np
+import soundfile as sf
+import torch
+from asteroid.data.librimix_dataset import LibriMix
+import random as random
+from scipy import signal
+
+
+class MetricGAN(LibriMix):
+
+    def __getitem__(self, idx):
+        # Get the row in dataframe
+        row = self.df.iloc[idx]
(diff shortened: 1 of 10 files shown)


popcornell:
Hey Joris, 

thanks for this pull request. 
LGTM, one question only, why `asteroid/data/metricGAN.py` should be in the toolkit ? Wouldn't be better to put also that in the egs folder ?

JorisCos:
> Hey Joris,
> 
> thanks for this pull request.
> LGTM, one question only, why `asteroid/data/metricGAN.py` should be in the toolkit ? Wouldn't be better to put also that in the egs folder ?

Hey, I think you are right, the dataset `asteroid/data/metricGAN.py` only overrides the `__getitem__` function of `LibriMix`. It should probably go in the egs folder. The same reasoning applies for SEGAN's recipe. 
I wait for @mpariente  thought on this and I will make the change.
Thanks

PR #189, opened 2020-08-04
STOI loss example: Add sample rate

jonashaag:
I was trying STOI loss and realised after listening to some early examples that something is wrong with high frequencies.  Realised that I forgot to set the sample rate param for STOI loss.

This PR adds it to the docstring example.

I'd go as far and advocate for making `sample_rate` and *required* parameter that does not have a default – I would expect that almost everyone has to use resampling anyways because they're training in some other sample rate than 10kHz.

### DIFF
--- a/asteroid/losses/stoi.py
+++ b/asteroid/losses/stoi.py
@@ -6,7 +6,7 @@ Examples:
     >>> from asteroid.losses import PITLossWrapper
     >>> targets = torch.randn(10, 2, 32000)
     >>> est_targets = torch.randn(10, 2, 32000)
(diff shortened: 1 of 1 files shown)


mpariente:
Thanks for the PR, that's indeed better in the docstring. 
I agree to make `sample_rate` a positional argument, you're right that 10kHz is not common enough to be default.

jonashaag:
Ok, so you care for backwards compat in torch_stoi? If so we would have to work with a warning and still keep the 10kHz default. Otherwise we can simply remove the default value.

mpariente:
No, it's very young and not very used I guess for bx compatibility is not a priority

PR #193, opened 2020-08-11
Continuous Source Separation

popcornell:


mpariente:
Oh, I didn't understand what happened, I thought it was merged ^^

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +124,87 @@ def are_models_equal(model1, model2):
...
+class OverlapAddWrapper(torch.nn.Module):
+    def __init__(self, nnet, sources, window_size, device, window="hanning", reorder_chunks=True):
+        super(OverlapAddWrapper, self).__init__()
+
+        assert window_size % 2 == 0, "Window size must be even"
+
+        self.to(device)
+        self.device = device
+        if isinstance(nnet, torch.nn.Module):
+            nnet = nnet.to(device)
```
mpariente, line 137: I don't understand why we need that part.

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +124,87 @@ def are_models_equal(model1, model2):
...
+
+    def forward(self, x):
+        with torch.no_grad():
+            x = x.to(self.device)
+            assert len(x.shape) == 3
+
+            # Overlap and add:
+            # [batch, channels, n_frames] -> [batch, channels, window_size, n_chunks]
+            batch, channels, n_frames = x.size()
+            hop_size = self.window_size // 2
```
mpariente, line 169: Do we really want to hard code this? 
It would be cool to be able to reproduce table 5 from [this paper](https://arxiv.org/pdf/2003.01531.pdf). It wouldn't be too complicated with this class right?

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +124,87 @@ def are_models_equal(model1, model2):
...
+    def _reorder_sources(self, current, previous):
+        # we compute xcorr for all permutations
+        # we get index for min
+        batch, frames = current.size()
+        current = current.reshape(-1, self.sources, frames)
+        previous = previous.reshape(-1, self.sources, frames)
+
+        pw_losses = PITLossWrapper.get_pw_losses(lambda x, y: torch.sum((x.unsqueeze(1) * y.unsqueeze(2))),
+                                                 current[..., :self.window_size//2],
+                                                 previous[..., -self.window_size//2:])
```
mpariente, line 156: Would also need to change if we decide not to hard code the hope size.

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +124,87 @@ def are_models_equal(model1, model2):
...
+                # user must handle multichannel by reshaping to batch
+                assert len(tmp.size()) == 3
+                assert tmp.shape[1] == self.sources
+                tmp = tmp.reshape(batch * self.sources, -1)
+
+                if f == 0:
+                    out.append(tmp)
+                else:
+                    # we determine best perm based on xcorr with previous sources
+                    tmp = self._reorder_sources(tmp, out[-1])
```
mpariente, line 188: Should there be a check about `reorder_chunks`?

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +124,87 @@ def are_models_equal(model1, model2):
...
+        # we get index for min
+        batch, frames = current.size()
+        current = current.reshape(-1, self.sources, frames)
+        previous = previous.reshape(-1, self.sources, frames)
+
+        pw_losses = PITLossWrapper.get_pw_losses(lambda x, y: torch.sum((x.unsqueeze(1) * y.unsqueeze(2))),
+                                                 current[..., :self.window_size//2],
+                                                 previous[..., -self.window_size//2:])
+        _, perms = PITLossWrapper.find_best_perm(pw_losses, self.sources)
+        current = PITLossWrapper.reorder_source(current, self.sources, perms)
```
mpariente, line 158: This is beautiful !

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +124,87 @@ def are_models_equal(model1, model2):
         if p1.data.ne(p2.data).sum() > 0:
             return False
     return True
+
+class OverlapAddWrapper(torch.nn.Module):
+    def __init__(self, nnet, sources, window_size, device, window="hanning", reorder_chunks=True):
```
mpariente, line 129: Can we rename `sources` to `n_src`, as almost everywhere in the code?

asteroid/utils/torch_utils.py
```diff
@@ -123,3 +124,87 @@ def are_models_equal(model1, model2):
...
+            hop_size = self.window_size // 2
+            folded = torch.nn.functional.unfold(x.unsqueeze(-1), kernel_size=(
+            self.window_size, 1),
+                                                padding=(self.window_size, 0),
+                                                stride=(hop_size, 1))
+
+            out = []
+            for f in range(folded.shape[-1]):  # for loop to spare memory
+                tmp = self.function(folded[..., f])
+                tmp = tmp * self.window
```
mpariente, line 178: If you multiply with the window here, the correlation will be a bit weird. Probably it's going to work but it might be better to multiply after. Not a priority, just a thought

asteroid/utils/torch_utils.py
popcornell: in case nnet is a function for whatever reason one might want to apply a function window by window

asteroid/utils/torch_utils.py
popcornell: IDK i usually use hop_size//2 because we get perfect reconstruction with hamming window

asteroid/utils/torch_utils.py
mpariente: But we can do this at runtime right? Send the Module to the device of the input for example? Or is it better to specify the device beforehand. 
I would treat it like a normal `nn.Module`

asteroid/utils/torch_utils.py
mpariente: Yes, true. 
Other question then, is the window needed? 
Alright, let's keep it like this for now, but register `self.hop_size` in the `__init__`  instead of having it hardcoded in two places in the code. If we want to change in the future, we'll be able.

asteroid/utils/torch_utils.py
popcornell: yeah you are right. Do we also need the whole to be differentiable  ?

asteroid/utils/torch_utils.py
popcornell: I ll add an option for no window

popcornell:
Added a simple test and moved everything to `dsp` folder

tests/dsp/overlap_add_test.py
```diff
@@ -0,0 +1,19 @@
...
+from asteroid.dsp.overlap_add import OverlapAddWrapper
+
+
+@pytest.mark.parametrize("length", [10390, 89372])
+@pytest.mark.parametrize("batch_size", [1, 2])
+@pytest.mark.parametrize("n_src", [1, 2])
+@pytest.mark.parametrize("window", ["hanning"])
+@pytest.mark.parametrize("window_size", [128])
+@pytest.mark.parametrize("hop_size", [64])
+def test_consistency_noweight(length, batch_size, n_src, window, window_size, hop_size):
```
mpariente, line 14:
```suggestion
def test_overlap_add(length, batch_size, n_src, window, window_size, hop_size):
```

tests/dsp/overlap_add_test.py
```diff
@@ -0,0 +1,19 @@
...
+    mix = torch.randn((batch_size, length)).reshape(batch_size, 1, -1)
+    nnet = lambda x: x.unsqueeze(1).repeat(1, n_src, 1)
+    oladd = OverlapAddWrapper(nnet, n_src, window_size, hop_size, window)
+    oladded = oladd(mix)
+    assert_allclose(mix.repeat(1, n_src, 1), oladded)
```
mpariente, line 19:
```suggestion
    assert_allclose(mix.repeat(1, n_src, 1), oladded)
    
```

asteroid/dsp/overlap_add.py
```diff
@@ -0,0 +1,75 @@
+import torch
+from scipy.signal import get_window
+from asteroid.losses import PITLossWrapper
+
+
+class OverlapAddWrapper(torch.nn.Module):
+    def __init__(self, nnet, n_src, window_size, hop_size, window="hanning", reorder_chunks=True):
```
mpariente, line 7: Default hop_size to None and if None to half of window size.

mpariente:
I applied some small changes. 
Fix code-style, add docs and move the OLA logic into its own function so if we want to reshape things before/after we can do it cleanly in `forward`. Tell me if it's ok for you.

popcornell:
LambdaOverlapAdd is cool. Ok I will add docs now

popcornell:
So we use a function for OLA and not a static method in your opinion ?

mpariente:
> So we use a function for OLA and not a static method in your opinion ?

That's a good question. I'm not sure. Let's see how it is like this. If we start using work-arounds, then we'll make it static no?

popcornell:
If we make a static and sufficiently flexible version we could also use it in DPRNN. So yeah I ll do it.

mpariente:
Then, it's better to have it as a function rather than a static method IMO. As the functional/module API are organized in torch/torchaudio.
(Let's keep DPRNN refactoring for a follow up PR)

popcornell:
Made _reorder_sources an external function as it might be useful in other parts for whatever reason. 
Also we have perfect reconstruction also with no window. Added also enable_grad flag to enable backprop.

popcornell:
Oh let me add some docs on _reorder_sources before you merge

asteroid/dsp/overlap_add.py
```diff
@@ -0,0 +1,147 @@
...
+        """ Forward module: segment signal, apply func, combine with OLA.
+
+        Args:
+            x (:class:`torch.Tensor`): waveform signal of shape (batch, 1, time).
+
+        Returns:
+            :class:`torch.Tensor`: The output of the lambda OLA.
+        """
+        # Here we can do the reshaping
+        if self.enable_grad:
```
jonashaag, line 103: I think you could use ` with torch.autograd.set_grad_enabled(self.enable_grad): ...` here

asteroid/dsp/overlap_add.py
```diff
@@ -0,0 +1,147 @@
...
+            padding=(self.window_size, 0),
+            stride=(self.hop_size, 1),
+        )
+
+        out = []
+        n_chunks = folded.shape[-1]
+        for f in range(n_chunks):  # for loop to spare memory
+            tmp = self.nnet(folded[..., f])
+            # user must handle multichannel by reshaping to batch
+            assert len(tmp.size()) == 3, "nnet should return (batch, n_src, time)"
```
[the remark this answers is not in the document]
popcornell: we can leave the check maybe only for the first iteration.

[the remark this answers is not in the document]
popcornell: but it is pretty inexpensive wrt self.nnet which can be whole ConvTasNet so...

asteroid/dsp/overlap_add.py
popcornell: cool I will switch to that I did not knew

asteroid/dsp/overlap_add.py
```diff
@@ -0,0 +1,147 @@
...
+            x.unsqueeze(-1),
+            kernel_size=(self.window_size, 1),
+            padding=(self.window_size, 0),
+            stride=(self.hop_size, 1),
+        )
+
+        out = []
+        n_chunks = folded.shape[-1]
+        for f in range(n_chunks):  # for loop to spare memory
+            tmp = self.nnet(folded[..., f])
```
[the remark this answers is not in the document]
popcornell: switched to frame_idx is clearer

asteroid/dsp/overlap_add.py
```diff
@@ -0,0 +1,147 @@
...
+            self.use_window = False
+
+        self.register_buffer("window", window)
+        self.reorder_chunks = reorder_chunks
+        self.enable_grad = enable_grad
+
+    def ola_forward(self, x):
+        """Heart of the class: segment signal, apply func, combine with OLA."""
+
+        assert len(x.shape) == 3
```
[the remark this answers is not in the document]
popcornell: sure

asteroid/dsp/overlap_add.py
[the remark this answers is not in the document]
popcornell: I left a check on first iteration only, the check should be there because it is related to output of self.nnet

popcornell:
Thanks @jonashaag  for the review, I think we can merge it. This is a small step towards a proper inference engine.

mpariente:
Thanks to both of you ! 
Small note: we squash all the commits before merging for a clearer history, let's keep this in mind.

asteroid/dsp/overlap_add.py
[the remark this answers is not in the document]
jonashaag: OK! I do not think it's an issue with performance. But usually, as far as possible, it is considered best practise to check types, shapes, etc. at the very top of a function, so it is easy to see what assumption the function makes about its inputs. (Best case, if the language's type system is strong enough you can encode all the assumptions in the types so you do not need any runtime checks at all.)

PR #195, opened 2020-08-12
[tests] Add precommit hook with black

mpariente:
First steps towards adopting [`black`](https://github.com/psf/black) !

The idea is to gain time and energy :
- No need to think about code style when writing
- No need to worry about code style when reviewing (huge loss of time)
- Easier for new comers to open their PRs 

For now, we'll stick with 120 line length, but we'll probably move to 100 afterwards. 
Also need to add `flake8` but I don't know it much yet.

**Resources**
- [Django adopting black](https://github.com/django/deps/blob/master/accepted/0008-black.rst): see the end for the procedure we'll follow
- [Why prettier in JS](https://prettier.io/docs/en/why-prettier.html): good read.
- [Linus about line length](https://lkml.org/lkml/2020/5/29/1038).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,19 @@
+default_language_version:
+  python: python3.6
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      - id: check-yaml
+
+  - repo: local
+    hooks:
+      - id: black
+        name: black
+        entry: python -m black
+        args: [--skip-string-normalization, --config, ./pyproject.toml]
+        language: system
+        types: [python]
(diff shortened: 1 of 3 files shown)


jonashaag:
Nice!

Why no string normalization?

mpariente:
It'll probably come when we apply black to the whole codebase. 
For now, two many quotes are single (and I like them better but I should adhere to black..) to enforce double quote now.

jonashaag:
In my opinion changing some style settings of a standardized formatted defeats the purpose (of not having to make any choices). Unless there is no technical reason to not go with the default settings I personally would simply go with the default settings.

mpariente:
I add string normalization in the next PR that does the black formatting. 
However, I think I'll keep line length to 100 instead of 88. I don't see the point of 88 now that Github shows up to 120 chars without scrolling.

jonashaag:
Agreed!

PR #196, opened 2020-08-12
[CI] Add .flake8 configuration

mpariente:
@jonashaag I guess you're familiar with Flake8, does this look reasonable? 
I plan to handle line-length with black rather than with flake because the `# fmt: off` is more practical if needed.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,14 @@
+[flake8]
+max-line-length = 119
+exclude = docs/source,*.egg,build
+select = E,W,F
+verbose = 2
+# https://pep8.readthedocs.io/en/latest/intro.html#error-codes
+format = pylint
+ignore =
+    E731  # E731 - Do not assign a lambda expression, use a def
+    F401  # F401 - Imported but unused. Imports can be useful for users
+    W605  # W605 - invalid escape sequence '\_'. Needed for docs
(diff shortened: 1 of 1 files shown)


jonashaag:
I usually only use the F* rules. Style rules are different from what black does.

Do you have an example of F401 false positive? Usually you’d use `__all__` in the case you mentioned

mpariente:
This I think
https://github.com/mpariente/asteroid/blob/master/asteroid/masknn/blocks.py#L6

Black says it's ok. If it's problematic, I'll remove them after. 

> I usually only use the F* rules. 

Please be polite :joy:

jonashaag:
😅

I think we should keep all F* rules and use noqa for the few false positives. 

Oh btw I would also suggest to add isort to the tool chain. And maybe mypy in the future.

mpariente:
Then, you need to use `__all__` for each __init__ file that locally imports for the user?

mpariente:
Mypy would be useful if we'd have typing.
Addind type annotations for the whole codebase seems boring ^^ But later will only be more boring..

jonashaag:
Yes or you can simply add `# noqa` to these "alias" imports.

We can also add Mypy in the future.

PR #198, opened 2020-08-12
[black] Flake8 compatibility + Black strings

mpariente:
- Made `asteroid/asteroid` flake8 compliant
- Applied black to strings as well (forgot earlier..)

Now the source code and the tests are black + flake8 compliant, which is pretty cool. 

I'm think about adding flake8 to the pre commit hook.
I'm not sure it's very nice for new-comers. Chances are the code is not always flake compliant. just to commit will be annoying.
I'd add it in the CI but not as a pre-commit hook, @jonashaag @popcornell any thoughts?


jonashaag:
Sounds good! Maybe there's also some kind of warning mode for pre-commit that we could use with flake8.

mpariente:
I saw that `--exit-zero` does that, but I'm not sure it's useful because the commit is sent right? 
Or if we integrate deep source, the warnings appear locally so they can be fixed in the next commit?

mpariente:
TravisCI takes around 15mins per build, I'm bored. 
I'm going to switch to Github Actions soon.

PR #199, opened 2020-08-13
[tests] First Github CI trial

mpariente:
Let's try out Github actions instead of Travis because it's pretty slow.

For now, the build is the same to measure speed. We can split it later.


mpariente:
GitHub tests are running in around 2mins instead of 15mins, this is a game changer !

jonashaag:
Wow, didn't know that. Nice!

Installing dependencies manually (not from requirements.txt) isn't nice though, is there a reason?

mpariente:
It was the same in the travis build. 
numpy and Cython need to be installed before pesq. For the rest, it can be installed from requirements.txt

Yes, gaining 10minutes is really cool !

jonashaag:
Then they should be a dependency of pesq, or at least we should put that in the requirements.txt – because anyone trying to install Asteroid will run into the problem of not having these two things installed, right?

mpariente:
The problem is that pesq needs a two-step install. It is in the requirements, but it doesn't install at once. At least, it was the case at the time I wrote the travis build. I'll try tonight if I can get rid of this line

jonashaag:
I think the problem is that their list of dependencies is broken, but I think it was fixed in GH master

mpariente:
My previous commit tried to install pesq without install numpy and cython before and it didn't work (as expected). 
This problem was fixed in `python-pesq`'s master, so I asked for a new PyPI release with the fix ([ref](https://github.com/ludlows/python-pesq/issues/14). 
In the mean time, I'll do the install by hand in the CI.

mpariente:
@jonashaag, agree to move away from travis?

jonashaag:
Definitely! I have to preference at all

.github/workflows/test_formatting.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+
+      - name: Install Black and flake8
+        run: pip install black==19.10b0 flake8
```
jonashaag, line 17: Best case we can install these from a file like `requirements_dev.txt` instead of hardcoding

jonashaag:
What's a bit worrisome with the current CI scripts is that lots of things are hardcoded (requirements, flake8 options, paths to ignore etc.). I'm not sure if there are standard config files where we can put all of these things (`pyproject.toml` maybe?). Otherwise I suggest creating shell scripts for all of them, or putting everything together in a Makefile. Reason being that you should be able to run all of the CI stuff on your local machine before pushing to GH.

jonashaag:
Apparently you can also put some stuff into `setup.cfg`: https://github.com/django/django/blob/master/setup.cfg

Oh boy this entire Python packaging and CI business is such a mess :-D I'm not an expert in that area unfortunately so I'm not sure what's best practise

mpariente:
I understand your concern. Let's wait until `pesq` new release and reorganize everything to be ran easier locally before pushing. I'd go with a bash script, yes.


> Oh boy this entire Python packaging and CI business is such a mess :-D I'm not an expert in that area unfortunately so I'm not sure what's best practise

I've also read quite a bit about it in the past days but it's not crystal clear either ^^

PR #200, opened 2020-08-13
Dual Path Transformer Network 

popcornell:
we have to add the training learning rate schedulers. Any idea ?


popcornell:
Also look at the new Oladd (maybe be better to call it DualPathModule or smth like that)

mpariente:
I'll review tonight, sorry for the delay.. 
Did you set up the pre-commit hook yet? If not, please do so.

popcornell:
I ll tidy up the code when you all think is good. I have troubles right now with the pre-commit hook, I am going fast and I do not want to loose momentum on this for now

asteroid/engine/schedulers.py
```diff
@@ -0,0 +1,69 @@
...
+        self.__dict__.update(state_dict)
+
+    def state_dict(self):
+        return {key: value for key, value in self.__dict__.items()}
+
+
+
+class NoamScheduler(_BaseScheduler):
+    def __init__(self, optimizer, d_model, warmup_steps, scale=1.0):
+        super(NoamScheduler, self).__init__(optimizer)
```
mpariente, line 33: Let's have `super().__init__(optimizer)` instead.
This is in [PEP367](https://www.python.org/dev/peps/pep-0367/).
Any of you knows how to enforce this?

asteroid/masknn/attention.py
```diff
@@ -0,0 +1,229 @@
...
+        super(Oladd, self).__init__()
+        self.chunk_size = chunk_size
+        self.hop_size = hop_size
+        self.n_orig_frames = None
+
+    def unfold(self, x):
+        # x is (batch, chan, frames)
+        batch, chan, frames = x.size()
+        assert x.ndim == 3
+        self.n_orig_frames = x.shape[-1]
```
mpariente, line 47: I love this caching idea.
We'll need a warning somewhere in the docs that it keeps track of the last input seen, so that you cannot call `unfold` twice on different inputs and expect `fold` to work properly. 
We can add an optional arg on the `fold` that specifies n_orig_frames if we don't want to use the default.

asteroid/masknn/attention.py
```diff
@@ -0,0 +1,229 @@
...
+        x = self.dropout(out) + x
+        x = self.norm_mha(x.transpose(1, -1)).transpose(1, -1)
+
+        # lstm is applied
+        out = self.linear(self.dropout(self.activation(self.recurrent(x)[0])))
+        x = self.dropout(out) + x
+        return self.norm_ff(x.transpose(1, -1))
+
+
+class Oladd(nn.Module):
```
mpariente, line 35: This looks amazing ! Just the note about the length caching which is cool but a bit dangerous
.

mpariente: Also this will probably need to go somewhere else, right?

asteroid/masknn/attention.py
```diff
@@ -0,0 +1,229 @@
...
+        to_unfold = x.reshape(batch, chan * self.chunk_size, n_chunks)
+        x = torch.nn.functional.fold(
+            to_unfold,
+            (self.n_orig_frames, 1),
+            kernel_size=(self.chunk_size, 1),
+            padding=(self.chunk_size, 0),
+            stride=(self.hop_size, 1),
+        )
+
+        x /= (self.chunk_size / self.hop_size)
```
mpariente, line 69: This is to unit test against `fold(unfold(x)) == x` right?

asteroid/masknn/attention.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def inter_process(self, x, module):
+        batch, channels, chunk_size, n_chunks = x.size()
+        x = x.transpose(1, 2).reshape(batch*chunk_size, channels, n_chunks)
+        x = module(x)
+        x = x.reshape(batch, chunk_size, channels, n_chunks).transpose(1, 2)
+        return x
+
+
+class DPTNet(nn.Module):
```
mpariente, line 92: Looks perfect to me !

asteroid/masknn/attention.py
```diff
@@ -0,0 +1,229 @@
+import torch.nn.functional as F
+import torch.nn as nn
+from torch.nn.modules.activation import MultiheadAttention
+from asteroid.masknn import activations, norms
+import torch
+from asteroid.utils import has_arg
+
+class ImprovedTransformedLayer(nn.Module):
+    def __init__(self, d_model, n_heads, dim_ff, dropout=0., activation="relu", bidirectional=True, norm="gLN"):
```
mpariente, line 9: What is the accepted name for the  in other packages (torch, huggingface, fairseq, espnet)? I'd like to have the same name so that it's easier to enter Asteroid's source code.
I saw `embed_dim` in torch, `n_feats` in espnet now.

asteroid/masknn/attention.py
```diff
@@ -0,0 +1,229 @@
+import torch.nn.functional as F
+import torch.nn as nn
+from torch.nn.modules.activation import MultiheadAttention
+from asteroid.masknn import activations, norms
+import torch
+from asteroid.utils import has_arg
+
+class ImprovedTransformedLayer(nn.Module):
```
mpariente, line 8: I know they call it like that but is this really a transformer layer in your opinion?

asteroid/engine/schedulers.py
```diff
@@ -0,0 +1,69 @@
...
+    def load_state_dict(self, state_dict):
+        self.__dict__.update(state_dict)
+
+    def state_dict(self):
+        return {key: value for key, value in self.__dict__.items()}
+
+
+
+class NoamScheduler(_BaseScheduler):
+    def __init__(self, optimizer, d_model, warmup_steps, scale=1.0):
```
mpariente, line 32: `d_model` is the projection dim of the MHA?

asteroid/engine/schedulers.py
```diff
@@ -0,0 +1,69 @@
+
+class _BaseScheduler(object):
```
mpariente, line 2: This is super cool, love it !

asteroid/engine/schedulers.py
```diff
@@ -0,0 +1,69 @@
...
+        self.warmup_steps = warmup_steps
+
+    def _get_lr(self):
+        lr = self.scale * self.d_model ** (-0.5) * \
+                min(self.step_num ** (-0.5), self.step_num * self.warmup_steps ** (-1.5))
+        return lr
+
+
+class DPTNetScheduler(_BaseScheduler):
+    def __init__(self, optimizer, steps_per_epoch, d_model, warmup_steps, noam_scale=1.0,
```
mpariente, line 45: Can we have a sensible default for `warmup_steps`?

mpariente:
No problem for the pre-commit hook, did you follow [this](https://github.com/mpariente/asteroid/blob/master/CONTRIBUTING.md)? 
There is the `pre-commit install` step which is important. 
Anyway, I can format it with black before merging no problem.

asteroid/engine/schedulers.py
jonashaag: Interesting there doesn’t seem to be a linter first that

asteroid/masknn/attention.py
jonashaag: Or make it explicit by giving them descriptive names fold_xyz and unfold_xyz (can’t come up with a good name at the moment sorry) and have “non magic” versions h that are simply called fold/unfold

asteroid/masknn/attention.py
popcornell: yes this guarantees perfect reconstruction

asteroid/masknn/attention.py
popcornell: it is MHA on top of standard DPRNN layer

asteroid/engine/schedulers.py
popcornell: the input dimension of MHA. In our case input and projection dim are the same

popcornell:
Yes but i have it failing probably because I use python 3.8

asteroid/masknn/attention.py
popcornell: probably on dsp no ?

asteroid/masknn/attention.py
popcornell: why you can't unfold twice ? you can and as long as you fold it two times it is going to work

mpariente:
```python
u = unfold(x)
v = unfold(y)
new_x = fold(u)
```
`fold(u)` will use the `self.n_orig_frames` from `y`, not from `x`. So the output size of `new_x` won't be the one of `x` (if y and x have different length). 
Right?

mpariente:
I'll install python 3.8 today and see if I can make it work.

popcornell:
No need for that I ll solve it

asteroid/masknn/attention.py
mpariente: Yes, it makes sense

asteroid/engine/schedulers.py
```diff
@@ -1,6 +1,9 @@
+from torch.optim.optimizer import Optimizer
 
-class _BaseScheduler(object):
+
+class _BaseScheduler(Optimizer):
     def __init__(self, optimizer):
+        super(_BaseScheduler, self).__init__(optimizer.__dict__["param_groups"], optimizer.__dict__["defaults"])
```
popcornell, line 6: this is hacky... because lightnining checks the type of optimizer, the scheduler should be an optimizer. This does it but it would be better if there is a way to directly wrap the optimizer by overriding its methods

mpariente: I think, this is possible, with [this](https://stackoverflow.com/a/46757134/7248396)

mpariente:
It seems that this is the last place where codecov is "active", but waiting. I don't know what's happening, any of you have ideas? Maybe it's related to the github actions but now sure.

mpariente:
I'm trying to rebase master and force push to fix it. Sorry for that in advance.

mpariente:
Sorry for force-pushing to this branch but it seems the coverage got stuck here. 
I disabled the patch cov report temporally because this is the one which is stuck

asteroid/engine/schedulers.py
popcornell: I will go with this I think

```python
class _BaseScheduler(object):
    def __init__(self, optimizer):
        self.__class__ = type(optimizer.__class__.__name__,
                              (self.__class__, optimizer.__class__),
                              {})
        self.__dict__ = optimizer.__dict__
        self.optimizer = optimizer
        self.step_num = 0
```

popcornell: fails with distributed data parallel

asteroid/engine/schedulers.py
```diff
@@ -1,9 +1,11 @@
 from torch.optim.optimizer import Optimizer
 
 
-class _BaseScheduler(Optimizer):
+class _BaseScheduler(object):
     def __init__(self, optimizer):
-        super(_BaseScheduler, self).__init__(optimizer.__dict__["param_groups"], optimizer.__dict__["defaults"])
+        self.__class__ = type(optimizer.__class__.__name__,
```
jonashaag, line 6: I wonder if this will make problems with saving/loading models. I guess `_BaseScheduler` instances can't be serialized (e.g. pickled) because of this dynamic class creation.

I guess this is a workaround for a limitation in lightning; wouldn't it be better to fix it in lightning?

popcornell: yes exactly... it fails with distributed data parallel because it uses multiprocessing and it requires instances to be serialized. 
Works fine however with no ddp or dp.. it is a shame. I am looking at alternatives.

mpariente: Maybe fixing it in lightning is not a bad idea, but it we'll have to wait for a release.

jonashaag: We can keep the workaround here until it's fixed in lightning.

asteroid/masknn/attention.py
popcornell: Ok i see now your point. we can either enforce to pass the original sequence length to unfold or have a check to prevent double calling. What would be better ?

popcornell:
Okay, this is starting to look decent. We have new schedulers logic and support for multiple schedulers, we can also have native torch schedulers called at each step instead of each epoch.

asteroid/masknn/attention.py
popcornell: also can have a fully functional version... IDK really

asteroid/dsp/overlap_add.py
```diff
@@ -146,3 +147,61 @@ def _reorder_sources(
...
+            padding=(self.chunk_size, 0),
+            stride=(self.hop_size, 1),
+        )
+
+        return unfolded.reshape(
+            batch, chan, self.chunk_size, -1
+        )  # (batch, chan, chunk_size, n_chunks)
+
+    def fold(self, x, output_size=None):
+        output_size = output_size if output_size is not None else self.n_orig_frames
```
mpariente, line 176: This is what I would do instead of the assert in `unfold`.
If you have `DualPathProcessing` in your model, you would have to instantiate it everytime, it doesn't seem good to me.

mpariente:
Yes, this looks good ! 
Do you want me to add the docs and the tests?

popcornell:
i can add them

asteroid/dsp/overlap_add.py
popcornell: Ok I see. No need for assert because code will fail lol

mpariente:
Is this ready to be merged on your end?

mpariente:
Made some changes, mainly aesthetics, @popcornell can you give your approval (or not :upside_down_face: )?

popcornell:
LGTM I think we can merge ! Thanks for adding docs to schedulers I totally forgot that.

PR #207, opened 2020-08-18
[src & tests] Split SuDORMRF architectures in encoder/masker/decoder

mpariente:
Split SuDORMRF architectures in encoder/masker/decoder instead of the current design where things are entangled and the separator network cannot be used on custom features. 

#### Few advantages:
- More modular -> Separator networks can be used outside asteroid. 
- Fit in encoder/masker/decoder architecture -> Inherit from `BaseTasNet` -> pretrained models OK.

#### Notes
- Small blocks are stored in masknn/_local.py and are private. 
- Added support for "all" activation functions for the output mask.
- No support for custom number of output bins (renamed `out_chan` by `bn_chan` because it was the inner dimension of the Ublocks, but it suggested otherwise). Adding support for it would be quite easy, should I do it?  

@etzinis would like to review?

### DIFF
--- a/asteroid/masknn/__init__.py
+++ b/asteroid/masknn/__init__.py
@@ -1,4 +1,4 @@
-from .convolutional import TDConvNet
+from .convolutional import TDConvNet, SuDORMRF, SuDORMRFImproved
 from .recurrent import DPRNN
 from .attention import DPTransformer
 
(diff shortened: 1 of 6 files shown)


jonashaag:
Close this, duplicate of #208?

mpariente:
Yes, I don't know why, github duplicated my last two PRs.

PR #208, opened 2020-08-18
[src & tests] Split SuDORMRF architectures in encoder/masker/decoder

mpariente:
Split SuDORMRF architectures in encoder/masker/decoder instead of the current design where things are entangled and the separator network cannot be used on custom features. 

#### Few advantages:
- More modular -> Separator networks can be used outside asteroid. 
- Fit in encoder/masker/decoder architecture -> Inherit from `BaseTasNet` -> pretrained models OK.

#### Notes
- Small blocks are stored in masknn/_local.py and are private. 
- Added support for "all" activation functions for the output mask.
- No support for custom number of output bins (renamed `out_chan` by `bn_chan` because it was the inner dimension of the Ublocks, but it suggested otherwise). Adding support for it would be quite easy, should I do it?  

@etzinis would like to review?

### DIFF
--- a/tests/models/models_test.py
+++ b/tests/models/models_test.py
@@ -69,7 +69,7 @@ def test_save_and_load_dprnn(fb):
 
 def test_sudormrf():
     model = SuDORMRFNet(
-        2, bn_chan=10, num_blocks=4, upsampling_depth=2, kernel_size=21, n_filters=10,
+        2, bn_chan=10, num_blocks=4, upsampling_depth=2, kernel_size=21, n_filters=12,
     )
     test_input = torch.randn(1, 801)
     model_conf = model.serialize()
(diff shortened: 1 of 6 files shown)


tests/models/models_test.py
```diff
@@ -68,19 +68,25 @@ def test_save_and_load_dprnn(fb):
...
+
+    reconstructed_model = SuDORMRFNet.from_pretrained(model_conf)
+    assert_allclose(model.separate(test_input), reconstructed_model(test_input))
 
 
 def test_sudormrf_imp():
-    model = SuDORMRFImproved(
-        2, out_chan=10, in_chan=10, num_blocks=4, upsampling_depth=2, n_filters=10, kernel_size=21,
+    model = SuDORMRFImprovedNet(
+        2, bn_chan=10, num_blocks=4, upsampling_depth=2, kernel_size=21, n_filters=12,
```
etzinis, line 83: Maybe change these default values?

mpariente: What do you mean? 
This is to test with a small architecture to speed up CI

mpariente:
I'm merging this. If I didn't understand your initial comment about the defaults, we can always iterate from there.

PR #209, opened 2020-08-18
[src & tests] Add LSTMTasNet to serializable models 

mpariente:
The implementation is not absolutely faithful to the original article. 

#### Notable differences
- No per-segment L2 normalization/rescaling
- No skip connection in stacked LSTM
- GlobLN used as "bottleneck layer"
- Stride support. 

@popcornell does this sound reasonable to you? 
This is the implementation we used in the Asteroid paper on the WHAM and WHAMR dataset.


mpariente:
Merging, we can iterate on it if this doesn't look reasonable or needs more params.

PR #210, opened 2020-08-18
[hub] Add LSTMTasNet/DPTNet/SuDORMRFNet to torch.hub!

mpariente:
Everything is in the tittle ! 

In the rush of adding new supported models, we forgot to extend the hub, here it is !


mpariente:
Rebased + forced push to handle codecov patch better

PR #212, opened 2020-08-18
[docs] Deploy docs to github pages

mpariente:
As said by few people, RTD alone can be nasty so we should also rely on gh-pages.


mpariente:
This seems to be working. 
It's going to build the docs for push to master and PRs to master and push to gh-pages only on master

PR #214, opened 2020-08-18
added TDCNpp

popcornell:
we might want to add TDCN++ from Universal Sound Separation and FUSS


asteroid/masknn/convolutional.py
```diff
@@ -211,6 +211,448 @@ def get_config(self):
...
+            "n_blocks": self.n_blocks,
+            "n_repeats": self.n_repeats,
+            "n_src": self.n_src,
+            "norm_type": self.norm_type,
+            "mask_act": self.mask_act,
+        }
+        return config
+
+
+class SuDORMRF(nn.Module):
```
mpariente, line 386: Can we have the code of TDCNpp only? There are few other things in this commit.

mpariente (commented):
Apart from the code which doesn't come from the PR, it looks good. 
We would need a unit test as well please.

asteroid/masknn/convolutional.py
popcornell: SuDORMRF was not merged ?

mpariente:
`sudo rm -rf SuDORMRF network` :rofl:

mpariente:
There is still a problem with the PR, sorry. We should have rebased master.

mpariente:
The SuDORMRF code was duplicated, I removed it. Fix two small things and added tests. 
Ready to merge on my side.

PR #220, opened 2020-08-19
[tests] Add scheduler tests

mpariente:
Forgot to do that. 
It's also useful because it can serve as an example later (let's test mixed precision with the new code first).

Is it normal that the step is not incremented @popcornell ? 
[See here](https://github.com/mpariente/asteroid/blob/master/asteroid/engine/schedulers.py#L117)

### DIFF
--- a/asteroid/utils/test_utils.py
+++ b/asteroid/utils/test_utils.py
@@ -1,0 +1,14 @@
+import torch
+from torch.utils import data
+
+
+class DummyDataset(data.Dataset):
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


popcornell:
It is incremented in .step() no?

mpariente:
You're completely right, I'm sorry.

mpariente:
In the current state of test, we don't test all the behavior of the scheduler because the fast dev run does only one batch. It's not a very useful test then. 
I'll try to improve it today.

PR #222, opened 2020-08-20
DAMP-VSEP dataset

groadabike:
Signed-off-by: Gerardo Roa Dabike <gerardo.roa@gmail.com>

This is my current version of the DAMP-VSEP dataset.
It relais in pre-constructed files with the details of the train, valid and test set.

If you run dampvsep/test_dataloader/test_dataloader.py, the first time will construct and save the metadata.


asteroid/data/dampvsep_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+
+dampvsep_license = dict(
+    title='DAMP-VSEP: Smule Digital Archive of Mobile Performances - Vocal Separation (Version 1.0.1) ',
+    title_link='https://zenodo.org/record/3553059',
+    author='Smule, Inc',
+    author_link='https://zenodo.org/record/3553059',
+    license="Smule's Research Data License Agreement",
+    license_link='https://zenodo.org/record/3553059',
+    non_commercial=True
+)
```
mpariente, line 231: That's perfect !

asteroid/data/dampvsep_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+from tqdm import tqdm
+import torch.utils.data
+
+import numpy as np
+import pandas as pd
+import random
+import json
+import librosa
+import warnings
+warnings.filterwarnings('ignore')
```
mpariente, line 12: What does this filters? 
A comment would be welcome.

asteroid/data/dampvsep_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+        amplitude_scaler = _get_amplitude_scaling_factor(vocal, back, snr=0)
+
+        return (sample['perf_key'],
+                {'mean': f"{mix.mean():.16f}",
+                 'std': f"{mix.std():.16f}",
+                 'scaler': f"{amplitude_scaler:.16f}",
+                 'vocal': sample['vocal_path'],
+                 'background': sample['background_path'],
+                 'vocal_duration': vocal_dur,
+                 'background_duration': back_dur})
```
mpariente, line 174: Better to declare the dict first, and then return it, it's more readable.

asteroid/data/dampvsep_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+
+            metadata_path.parent.mkdir(parents=True, exist_ok=True)
+            json.dump(tracks, open(metadata_path, 'w'), indent=2)
+
+        return tracks
+        
+    @staticmethod
+    def _build_metadata(inputs):
+        sample, root, sample_rate = inputs
+        back, _ = librosa.load(root / sample['background_path'],
```
mpariente, line 154: We usually use `soundfile` for loading but it doesn't resample. 
Is resampling needed? If yes, isn't is expensive to resample on the fly when it could be fixed?

asteroid/data/dampvsep_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+            Default ``None`` which allows any silence length
+        num_workers (int, optional): Number of workers.
+            Default to ``None`` which utilise all available workers
+     """
+
+    dataset_name = 'DAMP-VSEP'
+
+    def __init__(self, root_path, task, split, samples_per_track=1,
+                 random_segments=False, sample_rate=16000,
+                 segment=None, silence_in_segment=None, num_workers=None, norm=None):
```
mpariente, line 50: Would need docs about the `split` and `norm` argument.

asteroid/data/dampvsep_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+            Defaults to ``None`` which loads the full-length audio tracks.
+        silence_in_segment (float, optional): Max duration of continuous silence in a segment.
+            Default ``None`` which allows any silence length
+        num_workers (int, optional): Number of workers.
+            Default to ``None`` which utilise all available workers
+     """
+
+    dataset_name = 'DAMP-VSEP'
+
+    def __init__(self, root_path, task, split, samples_per_track=1,
```
mpariente, line 48: How about having a default for `task` and `split`? 
Positional args are always confusing in the long term.

asteroid/data/dampvsep_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+        return infos
+
+
+def _get_amplitude_scaling_factor(v, b, snr=0):
+    """Given v and b, return the scaler s according to the snr.
+    Args:
+      v: ndarray, vocal.
+      b: ndarray, backgroun.
+      snr: float, SNR. Default=0
+    Outputs:
```
mpariente, line 209: `Returns`

egs/dampvsep/test_dataloader/test_dataloader.py
```diff
@@ -0,0 +1,50 @@
...
+from pathlib import Path
+from tqdm import tqdm
+
+if __name__ == "__main__":
+    """dataset tests -> these parameters will go into recipes"""
+    parser = argparse.ArgumentParser(description='Dataset dataset test')
+
+    parser.add_argument(
+        '--root', type=str, help='root path of dataset',
+        default='/media/gerardo/TOSHIBA/DataSets/DAMP/DAMP-VSEP'
```
mpariente, line 12: We try to remove all user-specific absolute path.

mpariente (commented):
Overall, it looks great, thanks !
I have this concern about on-the-fly resampling that I'd like to discuss. 
Also, instead of passing the root_dir, I would probably use `sed` to prepend it to all the path in the csv files. But we might get to that when working on the recipe !

asteroid/data/dampvsep_dataset.py
groadabike: Librosa it is also used in AVSpeech and they have it as a requirement in the recipe.
I could do the same

mpariente:
Hey @groadabike, could we get this in? 
Thanks ! =D

groadabike:
Hi @mpariente, didn't have much time to do all the modifications I wanted. 
Think it would be better if I cancel this PR and create a new one that includes the data loader, and the ConvTasNet recipe.
It's that ok?

mpariente:
Thanks for getting back! Yes, it's completely fine

groadabike:
Close PR. Will be resubmitted with updated code and recipe.

PR #217, opened 2020-08-19
[src & egs] Prepare lightning upgrade to 0.9.0

mpariente:
- `optimizer_step` change
- Fix trainer args
- Fix verbose is boolean

PR #223, opened 2020-08-20
[src & tests] Fix no scheduler bug 

mpariente:
As pointed out by @jonashaag in #217 

Quite critical, because 0.3.1 was just released. We're lucky we have schedulers in all our recipes.
Tag @popcornell


mpariente:
Ok, I'm merging this to release 0.3.2.

PR #236, opened 2020-08-23
[cli] Add asteroid-infer CLI to enhance/separate from the command line

mpariente:
The is a first iteration, it works, but it's not optimal and raises some flaws of the current code. 

### Usage 

```bash
asteroid-infer url_or_path --files filename
asteroid-cli 'mpariente/ConvTasNet_WHAM!_sepclean' --files tmp.wav
asteroid-cli 'mpariente/ConvTasNet_WHAM!_sepclean'  --files a_folder a_filename a_glob/**/*.wav
```
`files` can be a list of filenames, foldernames and globs, we'll handle it. In case of a folder, we check the supported file formats in the folder (recursively)

### Cool things
- Thanks to `BaseModel`, we can now load all the models, just given the model pack, which makes it very easy. 
- Based on  `separate`, not much new code here.
- The `--files` argument is cool because versatile.

### Bad things
- soundfile doesn't support all format. Should we have a backup strategy? Librosa supports more formats? 
- Bash breaks on `!` unless we use single quotes on them, avoid `!` in the future.
- We currently have no idea the sample rate the model supports and this can break everything, we should add it in the saved models
- We also don't know the number of sources, so we cannot use the current version of `LambdaOverlapAdd`, could we remove the `n_src` argument in `LambdaOverlapAdd` so that we could use it in the CLI? It can be infered at runtime IMO (and if the (batch, n_src, time) format is not respected, errors will be raised. 

@jonashaag and @popcornell please? 

### Note
You need to reinstall Asteroid with `python setup.py develop` to have the new CLI modes I think.

### DIFF
--- a/asteroid/scripts/asteroid_cli.py
+++ b/asteroid/scripts/asteroid_cli.py
@@ -90,6 +90,13 @@ def upload():
 
 
 def infer():
+    """ CLI function to run pretrained model inference on wav files.
+
+    Args:
+        url_or_path(str): Path to the pretrained model.
+        files (List(str)): Path to the wav files to separate. Also support list
+            of filenames, directory names and globs.
+    """
     import argparse
 
     parser = argparse.ArgumentParser()
@@ -98,7 +105,8 @@ def infer():
         "--files",
         default=None,
         type=str,
-        help="Path to the wav files to separate. Also support list of filenames, directory names and globs.",
+        help="Path to the wav files to separate. Also support list of filenames, "
+        "directory names and globs.",
         nargs="+",
     )
     args = parser.parse_args()
--- a/asteroid/scripts/asteroid_cli.py
+++ b/asteroid/scripts/asteroid_cli.py
@@ -1,8 +1,20 @@
 import os
 import yaml
+import itertools
+import glob
+import warnings
+from typing import List
 
 import asteroid
 from asteroid.models.publisher import upload_publishable
+from asteroid.models.base_models import BaseModel
+
+
+SUPPORTED_EXTENSIONS = [
+    ".wav",
+    ".flac",
+    ".ogg",
+]
 
 
 def upload():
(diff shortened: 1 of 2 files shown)


asteroid/scripts/asteroid_cli.py
```diff
@@ -75,3 +87,49 @@ def upload():
...
+            if not local_list:
+                warnings.warn(f"Could find any file that matched {f}", UserWarning)
+            all_files.extend(local_list)
+    return all_files
+
+
+def glob_dir(d):
+    """ Return all filenames in directory that match the supported extensions."""
+    return list(
+        itertools.chain(*[glob.glob(os.path.join(d, "**/*" + ext)) for ext in SUPPORTED_EXTENSIONS])
```
jonashaag, line 134: I'd add `recursive=True` to `glob()`

jonashaag:
Nice and simple!

- We can use librosa to load all the files, it uses `soundfile` and falls back to `audioread` (ffmpeg on Linux IIRC). But it requires librosa to be installed. We could also try soundfile first and then fall back to librosa. Soundfile will get mp3 support soon https://github.com/erikd/libsndfile/issues/258
- I think we should add a check to `separate()` to make sure it never overwrites any files. That's particularly important now that we have code that goes through entire directories.

Am I correct that users can now install Asteroid and do something like `asteroid-infer mpariente/convtasnet-xyz /my/files` without ever manually downloading anything? That would be awesome.

faroit:
> We can use librosa to load all the files, it uses soundfile and falls back to audioread (ffmpeg on Linux IIRC). But it requires librosa to be installed. We could also try soundfile first and then fall back to librosa. Soundfile will get mp3 support soon erikd/libsndfile#258

I would use torchaudio here (which is using sox, and recently support for opus and ogg was added) or directly use ffmpeg-python which is supports almost all formats but is quite slow.

asteroid/scripts/asteroid_cli.py
popcornell: second this

popcornell:
Yes maybe torchaudio will be better also because they plan to support more formats in the future.

mpariente:
I'll use torchaudio for loading then. 
This will come in a follow up PR because it's not really related to the CLI. Thanks for the feedback

mpariente:
> I think we should add a check to separate() to make sure it never overwrites any files. That's particularly important now that we have code that goes through entire directories.

Yes, I agree. It will be in the follow up PR as well. 

> Am I correct that with this, user can now install Asteroid and do something like asteroid-infer mpariente/convtasnet-xyz /my/files without ever manually downloading any models? That would be awesome.

This is correct. This was also the case without the CLI. 
Note that the downloaded models are cached as well.

mpariente:
Only tested locally, we'll need to find a way to correctly test the CLI in the near future.

PR #237, opened 2020-08-23
Add jit compatibility

faroit:
#### Original message from @faroit 
* [x] remove global variables (e.g. `EPS`)
* [x] remove `@property` decorator

this PR only makes changes to make Demask torch scriptable. There are probably more changes required for other models. And some won't work at all currently: https://github.com/pytorch/pytorch/pull/37749

#### Edit from @mpariente  for visibility
Models are not scriptable in general (needs more typing support from torch), so we trace them instead. Functions that need scripting are scripted when tracing the model (e.g. encoder, decoder, some norms etc..). 
This PR adds unit-tested tracing for `ConvTasNet`, `DPRNNTasNet`, `LSTMTasNet`, `DeMask`, `DPTNet`, `SuDORMRFNet` and `SuDORMRFImprovedNet`. (See [this comment](https://github.com/mpariente/asteroid/pull/237#issuecomment-716035465) for a more complete list).
Notably, `Encoder` and `Decoder` were refactored to enable runtime scripting and multichannel support with jit.  

__Left to do__: `complex_nn`, `DCCRNet` and `DCUNet`


asteroid/filterbanks/enc_dec.py
```diff
@@ -253,7 +251,7 @@ def forward(self, spec):
                 padding=self.padding,
                 output_padding=self.output_padding,
             )
-        elif spec.ndim > 3:
+        else:
```
faroit, line 254: do you have an alternative to this? this is an important fix for torch.script as it does not allow forward passes that would have `Optional[torch.Tensor]`

mpariente: I don't exactly understand the question. 
This looks like it will always return a torch.Tensor with the `else` right?

mpariente:
This is a first step, thanks a lot !

mpariente:
There is this test failure which I don't fully understand, was not here before and doesn't really seem related

```
tests/engine/scheduler_test.py:16: in common_setup
    trainer = Trainer(max_epochs=1, fast_dev_run=True)
/opt/hostedtoolcache/Python/3.6.11/x64/lib/python3.6/site-packages/pytorch_lightning/trainer/trainer.py:310: in __init__
    torch.backends.cudnn.deterministic = self.deterministic
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <torch.backends.ContextProp object at 0x7f3cbd0ad7f0>
obj = <module 'torch.backends.cudnn' from '/opt/hostedtoolcache/Python/3.6.11/x64/lib/python3.6/site-packages/torch/backends/cudnn/__init__.py'>
val = False

    def __set__(self, obj, val):
        if not flags_frozen():
            self.setter(val)
        else:
            raise RuntimeError("not allowed to set %s flags "
>                              "after disable_global_flags; please use flags() context manager instead" % obj.__name__)
E           RuntimeError: not allowed to set torch.backends.cudnn flags after disable_global_flags; please use flags() context manager instead

/opt/hostedtoolcache/Python/3.6.11/x64/lib/python3.6/site-packages/torch/backends/__init__.py:39: RuntimeError
```
Do you know what this is about?

asteroid/filterbanks/enc_dec.py
faroit: yes, and this is required for torch.script as a path without a returned tensor is not allowed

faroit: the questions is if setting this path to the default is okay for yiou

mpariente: Yes, that seems alright. 
I don't think people will give 1-dim tensor as spectrograms..

faroit:
> Do you know what this is about?

@mpariente no, not really. But maybe we can take out a few commits to get this merged.

I think at least 3cbfb991b60bbc7353947e5e1401d0038ac2a483 should be picked

egs/kinect-wsj/DeepClustering/train.py
```diff
@@ -130,7 +128,7 @@ def validation_end(self, outputs):
             "progress_bar": {"val_loss": avg_loss},
         }
 
-    def unpack_data(self, batch):
+    def unpack_data(self, batch, EPS=1e-8):
```
mpariente, line 131: Actually not needed here?

mpariente:
Only 5 errors left but I don't know why, you didn't touch schedulers or system's code, that seems weird.

mpariente:
@mhu-coder will have a look at that as well, so let's look out for merge conflicts. 
We'll be three on the same branch. Please rebase (`git pull --rebase`) when possible and merge only when there are merge rebase conflicts. 

Also, there will be some tricky decisions to make to fix jitability, take a stab at it and add a comment on the PR at the places you're not sure about, we'll iterate from there.

Thanks in advance !

asteroid/masknn/convolutional.py
```diff
@@ -62,7 +62,7 @@ def forward(self, x):
         shared_out = self.shared_block(x)
         res_out = self.res_conv(shared_out)
         if not self.skip_out_chan:
-            return res_out
+            return res_out, res_out
```
mpariente, line 65: We should probably have a comment here to explain that this is for scriptability. 
I wonder, @faroit, is there a prettier way to do that? 
For example, this is not going to change when a model is exported (the control flow op is static), so this could be traced instead of scripted? And we could keep the same behaviour as currently?

asteroid/masknn/convolutional.py
```diff
@@ -175,17 +175,17 @@ def forward(self, mixture_w):
             :class:`torch.Tensor`:
                 estimated mask of shape [batch, n_src, n_filters, n_frames]
         """
-        batch, n_filters, n_frames = mixture_w.size()
+        batch, _, n_frames = mixture_w.size()
```
mpariente, line 178: What this related to the jitability?

mhu-coder: Oops, really sorry for that. As you have certainly guessed, it has nothing to do with jitability :s. Is there any way to rewrite the commit history without making it worse?

mpariente: It's better to leave unrelated code unchanged. This is fine for this one, you can leave it.

mpariente:
I wonder if we shouldn't trace models instead of scripting them, the static typing can be more annoying than the tracing. And we would script the methods we know won't be faithful. I don't know if we have a lot of _dynamic_ control flow in the models. Any opinion?

faroit:
> I wonder if we shouldn't trace models instead of scripting them, the static typing can be more annoying than the tracing. And we would script the methods we know won't be faithful. I don't know if we have a lot of dynamic control flow in the models. Any opinion?

yes, tracing would surely be favorable but I thought scripting is easier and should be done first. Also tracing requires fixed batch sizes for recurrent models. Not sure if we want to go that route?

mpariente:
But, you can mix scripting in tracing. Where needed, we would add the `@torch.jit.script` decorator (which can then handle the batch sizes in RNNs I guess), for the rest the tracing would handle the static control flow so we don't need to fix all the scripting problems. 
The intermediate representation is the same between scripting and tracing so the end models can be the same if we handle it well. 

I asked for opinions/feedback on twitter, just in case.

asteroid/masknn/convolutional.py
mhu-coder: If you want to go branchless in the different `forward` methods, we could add a pointer to a function that switches to  the actual forward method when the class is initialized. Not necessarily the most elegant solution, but it should remove the need for `if` statements.

```python
    # In Conv1DBlock
    def forward(self, x):
        """ Input shape [batch, feats, seq]"""
        shared_out = self.shared_block(x)
        res_out = self.res_conv(shared_out)
        if not self.skip_out_chan:
            return res_out, res_out
        skip_out = self.skip_conv(shared_out)
        return res_out, skip_out
```

```python
    def __init__(self, ...):
        #...
        if self.skip_conv:
            self._forward = self._skip_out_forward
        else:
            self._forward = self._no_skip_forward

    def forward(self, x):
        """ Input shape [batch, feats, seq]"""
        return self._forward(x)

    def _common_forward(self, x):
        shared_out = self.shared_block(x)
        return self.res_conv(shared_out)

    def _skip_out_forward(self, x):
        res_out = self._common_forward(x)
        skip_out = self.skip_conv(shared_out)
        return res_out, skip_out

    def _no_skip_forward(self, x):
        res_out = self._common_forward(x)
        return res_out, torch.tensor([0.])
```

What do you think?

mpariente: But you needed to keep the `torch.tensor([0.])` in the `_no_skip_forward` because script would raise an error otherwise right?  I thought about that and I don't think it solves the problem. Scripting will raise the same typing issues and tracing will still work in the same way. 

Also, if we return two tensors, there is a simpler solution without if in the forward, which is much more readable IMO
```python
self.skip_conv = nn.Conv1d if skip_chan else nn.Identity()
...
return res_out, self.skip_conv(res_out)
```
but this still doesn't solve the problem. 

We can really on the dev to decide whether the control flow is static or not, if it's static, trace is fine, otherwise script is needed, and in this case, typing issues need to be fixed.

faroit:
> But, you can mix scripting in tracing. Where needed, we would add the @torch.jit.script decorator (which can then handle the batch sizes in RNNs I guess), for the rest the tracing would handle the static control flow so we don't need to fix all the scripting problems.

yep, thats how I handled it in open-unmix now. The core (including the LSTM) is traceable, the wiener filter is scriptable.

mhu-coder:
Torchscript assumes that every variable with undefined types are Tensors, then crashes at runtime when it discovers that it wasn't the case or when the type of the variable changed. This also includes function keyword arguments as it does not use the default value to infer the type. The doc of [torch.jit](https://pytorch.org/docs/stable/jit_language_reference.html) advises to use a MyPy-style type annotations or type hints (like I did in `param_sinc_fb.py`) to force non `Tensor` types.

So it would be probably better to stick to one for consistency. @mpariente , do you have a preference between the two possible solutions?

mpariente:
> yep, thats how I handled it in open-unmix now. The core (including the LSTM) is traceable, the wiener filter is scriptable.

And in the end, do you trace or script the model? 

> Torchscript assumes that every variable with undefined types are Tensors, then crashes at runtime when it discovers that it wasn't the case or when the type of the variable changed. This also includes function keyword arguments as it does not use the default value to infer the type. The doc of [torch.jit](https://pytorch.org/docs/stable/jit_language_reference.html) advises to use a MyPy-style type annotations or type hints (like I did in `param_sinc_fb.py`) to force non `Tensor` types.
> 

> So it would be probably better to stick to one for consistency. @mpariente , do you have a preference between the two possible solutions?

I saw part of that, and adding typing to Asteroid's Modules wouldn't hurt.
But I still think, that both scripting and tracing can be mixed without problem, it's actually meant to be used this way, right? 
I don't think it's about consistency, tracing is limited by essence but more flexible, and scripting is more general but also more rigid and torch gives us those two tools, I think we can use both of them.

mhu-coder:
> I saw part of that, and adding typing to Asteroid's Modules wouldn't hurt.
I meant, would you prefer MyPy-style typing or Python type hints? Or you don't mind having a mix of both?

mpariente:
Oh ok sorry, didn't understand it this way. 
I'll have a look now then

mpariente:
Python type hints please :wink:

asteroid/masknn/norms.py
```diff
@@ -33,16 +32,17 @@ def forward(self, x):
         Returns:
             :class:`torch.Tensor`: gLN_x `[batch, chan, *]`
         """
-        dims = list(range(1, len(x.shape)))
+        dims: List[int] = torch.arange(1, len(x.shape)).tolist()
         mean = x.mean(dim=dims, keepdim=True)
         var = torch.pow(x - mean, 2).mean(dim=dims, keepdim=True)
-        return self.apply_gain_and_bias((x - mean) / (var + EPS).sqrt())
+        value = torch.sqrt((x - mean) / (var + EPS))
```
mpariente, line 38: @mhu-coder pointed out that this is wrong, the sqrt should only be applied to the variance. He'll fix it.

mhu-coder: thanks for reporting it here. I have added a test and fixed the issue (in 1 commit each) but have not pushed yet.

mhu-coder:
Regarding tracing vs. scripting, since the idea is to use `torch.jit.trace` everywhere and to switch to script mode by decorating functions that need this mode with `torch.jit.script`, would it be alright to replace calls to `torch.jit.script` with calls to `torch.jit.trace` in `jit_test.py`?

mhu-coder:
Hey guys,
A bit late for the hackathon and I don't know if this feature is still planned but I have had another look and solved some of the issues (there only warnings left when tracing the models in the tests). Below is a summary of the issues that are left:
* `DPTNet` expects the number of features to be a multiple of the number of heads in the multi attention layer but the `__init__` method of the class gives no control to the `n_heads` argument of `DPTransformer`. This makes writing the tracing test for `DPTNet` with `STFT` rather cumbersome. A way around it would be the addition of the keyword argument `n_heads` to the `__init__` of `DPTNet`, which would add another argument to a method that already has many...
* When tracing a model, the jit compiler raises a warning anytime it encounters a branch in the code as only one of these branches will be taken by the traced model. In particular, all the checks written in the models are effectively bypassed

On the upside, properties do not seem problematic.

mpariente:
Didn't read the code in detail but that sounds great ! 
Which version of torch are you using btw? 

> On the upside, properties do not seem problematic.

Which properties in particular? 




> * `DPTNet` expects the number of features to be a multiple of the number of heads in the multi attention layer but the `__init__` method of the class gives no control to the `n_heads` argument of `DPTransformer`. This makes writing the tracing test for `DPTNet` with `STFT` rather cumbersome. A way around it would be the addition of the keyword argument `n_heads` to the `__init__` of `DPTNet`, which would add another argument to a method that already has many...

I think it was an oversight, we should add `n_heads` in `DPTNet`. Many arguments to architectures is not a problem IMO. 
Regarding the problem between number of features and `n_head`, I wonder if there isn't a Linear layer between the input and the masker. I'll have a look. 


>  * When tracing a model, the jit compiler raises a warning anytime it encounters a branch in the code as only one of these branches will be taken by the traced model. In particular, all the checks written in the models are effectively bypassed

Can you elaborate please?

mhu-coder:
> Which version of torch are you using btw?

I am using torch 1.6.0. Are you using something else?

> Which properties in particular?

I remember that jit.script had issues with class properties but I may be wrong here. When using `trace`, tracing a `DeMask` model, to which I added a shameless copy/paste `sample_rate` property, did not raise errors.

> Can you elaborate please?

Example of warning:
```
tests/models/jit_test.py::test_trace_bss_model[LSTMTasNet]
  anaconda3/envs/asteroid/lib/python3.7/site-packages/torch/nn/modules/rnn.py:176: TracerWarning: Converting a
  tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python
  values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if self.input_size != input.size(-1):
```
I guess, the condition will always evaluate to the value that was found during tracing. Kind of ironic that the code that raises this warning is part of torch. But it also happens for some lines in asteroid too:

```
tests/models/jit_test.py::test_trace_bss_model[LSTMTasNet]
  asteroid/asteroid/filterbanks/enc_dec.py:172: TracerWarning: Converting a tensor to a Python boolean might
  cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as
  a constant in the future. This means that the trace might not generalize to other inputs!
    if channels == 1 and self.as_conv1d:
```

mpariente:
- 1.6.0 is fine, I was just asking. 
- Ok for the property. I guess it'll still break if there are graph computations inside it. 
- Thanks for the examples. We have to browse the code very carefully to find where the tracing isn't enough, and we need to script that part. There are graphs that'll stay static at run time (is there skip connection etc..) and others won't (number of input channels etc..). 

Regarding the DPTNet, I re-checked, there is no Linear layer in the paper. We should add one if the input dim is not divisible by the number of heads, otherwise it's pretty useless. @mhu-coder, do you feel like doing it?

mhu-coder:
> Regarding the DPTNet, I re-checked, there is no Linear layer in the paper. We should add one if the input dim is not divisible by the number of heads, otherwise it's pretty useless. @mhu-coder, do you feel like doing it?

Sure, will have a look in the afternoon.

FYI, I had a look at scripting the models instead of tracing them:
* `ConvTasNet` was fine
* `DPRNNTasNet` and `LSTMTasNet`: I ran into an issue in `torch/nn/modules/rnn.py` that read `GeneratorExp aren't supported`, which can be removed by converting a set comprehension into a regular loop but then ended up with a `AttributeError` caused by `torch/jit/frontend`.
* `DPTNet`: some issues with parts of torch and asteroid but ultimately, `torch.script` does not support methods/functions that take callables as argument (cf https://discuss.pytorch.org/t/function-type-in-torchscript/65416). This is the case of some functions in `asteroid/dsp/overlap_add.py`.

mpariente:
Ok, what's left to do to merge this @mhu-coder ?

mhu-coder:
Just added the linear layers (one at the input, another at the output) to DPTransform when `in_chan` is not a multiple of `n_heads`.

`DPTNet` is still missing an argument for `n_heads` in its `__init__` method.

Other than that, there is still the question of which model can be traced and which should be scripted instead. But I wonder if it wouldn't be better to give this responsibility the person that is going to trace/script a model.

asteroid/masknn/attention.py
```diff
@@ -122,7 +124,22 @@ def __init__(
...
+                f"output [{self.mha_in_dim} x {self.in_chan}])"
+            )
+            self.input_layer = nn.Linear(self.in_chan, self.mha_in_dim)
+            self.output_layer = nn.Linear(self.mha_in_dim, self.in_chan)
+        else:
+            self.input_layer = None
+            self.output_layer = None
+
+        self.in_norm = norms.get(norm_type)(self.mha_in_dim)
+        self.ola = DualPathProcessing(self.chunk_size, self.hop_size)
```
mpariente, line 142: @popcornell do you have an argument to keep the `DualPathProcessing` definition in the `forward`?

asteroid/masknn/attention.py
```diff
@@ -151,11 +168,11 @@ def __init__(
...
                 )
             )
-        net_out_conv = nn.Conv2d(self.in_chan, n_src * self.in_chan, 1)
+        net_out_conv = nn.Conv2d(self.mha_in_dim, n_src * self.mha_in_dim, 1)
         self.first_out = nn.Sequential(nn.PReLU(), net_out_conv)
         # Gating and masking in 2D space (after fold)
-        self.net_out = nn.Sequential(nn.Conv1d(self.in_chan, self.in_chan, 1), nn.Tanh())
-        self.net_gate = nn.Sequential(nn.Conv1d(self.in_chan, self.in_chan, 1), nn.Sigmoid())
+        self.net_out = nn.Sequential(nn.Conv1d(self.mha_in_dim, self.mha_in_dim, 1), nn.Tanh())
+        self.net_gate = nn.Sequential(nn.Conv1d(self.mha_in_dim, self.mha_in_dim, 1), nn.Sigmoid())
```
mpariente, line 175: The mask should be computed with `in_chan`, not with `mha_in_dim`. So actually, we don't need an output linear layer, but all these layers will `use `in_chan`, from `net_out_conv`.

asteroid/masknn/convolutional.py
mpariente: Still very unsure about that..

asteroid/utils/generic_utils.py
```diff
@@ -106,3 +106,16 @@ def unet_decoder_args(encoders, *, skip_connections):
             (enc_out_chan + skip_in_chan, enc_in_chan, enc_kernel_size, enc_stride, enc_padding)
         )
     return decoder_args
+
+
+def ceil(num, den):
```
mpariente, line 111: Isn't there a built-in function to do that?

tests/models/jit_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+    return params
+
+
+def test_enhancement_model(small_model_params):
+    params = small_model_params["DeMask"]
+    filter_banks = ["free", "stft", "analytic_free", "param_sinc"]
+    device = get_default_device()
+    inputs = ((torch.rand(1, 1000, device=device) - 0.5) * 2,)
+    test_data = torch.rand(1, 32, device=device)
```
mpariente, line 62: Why from 1000 to 32?

tests/models/jit_test.py
```diff
@@ -0,0 +1,105 @@
...
+            "kernel_size": 8,
+        },
+    }
+
+    return params
+
+
+def test_enhancement_model(small_model_params):
+    params = small_model_params["DeMask"]
+    filter_banks = ["free", "stft", "analytic_free", "param_sinc"]
```
mpariente, line 59: Let's test the filterbanks independently of the models, otherwise it'll grow too much. 
By using `BaseModel` with a `nn.Identity` masker, and test consistency with different input shapes once traced.

tests/models/jit_test.py
```diff
@@ -0,0 +1,105 @@
...
+    test_data = torch.rand(1, 32, device=device)
+    for filter_bank in filter_banks:
+        model = DeMask(**params, fb_type=filter_bank).eval().to(device)
+        traced = torch.jit.trace(model, inputs)
+
+        # check forward
+        with torch.no_grad():
+            ref = model(test_data)
+            out = traced(test_data)
+            assert torch.allclose(ref, out)
```
mpariente, line 71: Why not `torch.testing.assert_allclose`?

mpariente (commented):
Overally, this is great !!

Few points: 
- I'm still undecided about the Conv block with or without the slip connection, I think we'll remove the dummy output. This will be traced, not scripted. 
- I think tests about filterbanks should be done separately. And we should test that scripting works for real and that all input shapes that we want to support are supported. 
- `master` changed a bit since we started this PR so there are conflicts to solve, and also new architectures that we'll want to jit (DCCRN, DCUNet), so we need to merge `master`.

tests/models/jit_test.py
mhu-coder: Just to avoid using the same length at the input but does the value really matter?

mpariente: 32 samples makes 2 frames, it seems small. But I agree to have samples with different sizes.

tests/models/jit_test.py
mhu-coder: Do you mean having a class inheriting from `BaseModel` with a `masker` component set to `torch.nn.Identity` just to check that the traced model can process input data with various values of `ndim`?

In this case, why not a class inheriting from `BaseEncoderMaskerDecoder` instead? It makes writing the class much easier.

I have written the test but, as suspected, tracing will only take one of the branches in the `forward`s of the encoder and the decoder meaning that if a model has been traced with, say, a 2D input, the traced model will only work with 2D tensors. From here, we can either script the `forward`s of the encoder and decoder or rewrite these methods to not have branches. I'll have a look at the scripting solution.

mpariente: I meant  `BaseEncoderMaskerDecoder`, yes. 
And yes, scripting will be the solution.

mpariente:
Cool, thanks !

I actually thought that it was possible to decorate functions in `nn.Module`, but it's not for now ([see language reference](https://pytorch.org/docs/stable/jit_language_reference.html#language-reference)). My idea (before that) was to add a `script` decorator to all methods needing scripting, and that would have been satisfying. But that's not possible. So the `trace` method you propose is good for now.

mpariente:
Actually, if we make an external method out of `forward`, we can script automatically, which would be prettier. And for the Encoder/Decoder, it's possible. 

Also, we'll use `torch.jit.script_if_tracing` instead of `script` so we avoid the start up time.

mpariente:
Finally scripted the `forward` by making it external. I guess we don't need the `trace` function anymore, can we remove it?

mpariente:
So, I merged master. 
Tracing tests take quite a while. I'm wondering if it shouldn't be a different CI build altogether, so that we can really test thing rigorously. Thoughts? 

Note to myself (assistance accepted): before merging, need to go through the whole `masknn` codebase to be sure that tracing is enough on all the modules that we don't script.

asteroid/masknn/attention.py
popcornell: Sorry saw this now. No my bad this was an error actually

mpariente: Thanks !

mhu-coder:
> Finally scripted the `forward` by making it external. I guess we don't need the `trace` function anymore, can we remove it?

Neat! Sure, let's remove `trace`, no need to keep around lines that are unused

> before merging, need to go through the whole masknn codebase to be sure that tracing is enough on all the modules that we don't script.

It's not as good as carefully reading the code of each mask but we can add tests to check that every model can handle input variable shapes (actually, I am facing some runtime errors but will try to fix as many as I can). This will increase the time `jit_test.py` takes to run though.

mhu-coder:
Just added a couple of tests to check that traced models produce the same output for n-D input tensors (with n in [1, ..., 4]). Is it normal that most of the original/non-traced models crash?

mpariente:
> Just added a couple of tests to check that traced models produce the same output for n-D input tensors (with n in [1, ..., 4]). Is it normal that most of the original/non-traced models crash?

Yes. I don't see why not. 
Conv1d expects a 3D input, and doesn't support 4d inputs, LSTM either etc.. 

Thanks but those tests can be removed.

mhu-coder:
I think I see what you mean.

BTW, I saw that `DeMask` could not handle 1-D input when traced with a 2-D input but I guess this comes from the way the forward is written and the whole class could be refactored anyway.

When a `(batch, time)` data (with batch > 1) is given to LSTMTasNet, there's a RuntimeError that is raised in the forward of the decoder:
```
E       RuntimeError: The following operation failed in the TorchScript interpreter.
E       Traceback of TorchScript (most recent call last):
E         File "asteroid/filterbanks/enc_dec.py", line 296, in multishape_conv_transpose1d
E               view_as = (-1,) + spec.shape[-2:]
E               out = F.conv_transpose1d(
E                   spec.view(view_as),
E                   ~~~~~~~~~ <--- HERE
E                   filters,
E                   stride=stride,
E       RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.
```

I guess this could be the object of another pull request?

mpariente:
> BTW, I saw that DeMask could not handle 1-D input when traced with a 2-D input but I guess this comes from the way the forward is written and the whole class could be refactored anyway.

I agree. 

> I guess this could be the object of another pull request?

This PR is already so big that we can use `reshape` instead, no problem.

mhu-coder:
yeah, you're right. It's time to finish this PR.
I have just reverted the `res, res` return in `asteroid/masknn/convolutional.py` to what it was before as the method output consistency in type is only needed for `torch.jit.script`. So, in other word, if we always trace/never script a model relying on `Conv1DBlock` or `TDConvNet`, the inconsistency is fine.

Is the thorough check of the modules in `asteroid/masknn` the only thing left for this PR?

mpariente:
I guess, yes. 
Maybe we'll "fix" DeMask in another PR.

mpariente:
Not 100% sure 

- [x] masknn/norms
- [x] masknn/activations
- [x] masknn/attention
- [x] `TDConvNet`, `TDConvNetpp`, `Conv1dBlock`, `UConvBlock`, `UBlock`, `SuDORMRFImproved`, `SuDORMRF`. All but `# CHECK-JIT`
- [x] masknn/recurrent (missing   # CHECK-JIT)   
- [x] models: `BaseModel`, `ConvTasNet`, `DPRNNTasNet`, `LSTMTasNet`, `DPTNet` (+ added  # CHECK-JIT)
- [x] filterbanks: all filterbanks + encoder + decoder.
- [ ] filterbanks/transforms
- [ ] dsp
- [ ] 

This PR won't assure tracability/jitability for the complex module, the DCCRN and DCUNet. We'll do that on a follow-up PR (Cc. @jonashaag). Most of it is tagged with # CHECK-JIT on the class.

mpariente:
Going back to norms, this doesn't work
```python
from asteroid.masknn import norms

model = norms.GlobLN(10)
x = torch.randn(1, 10, 12)
traced = torch.jit.trace(model, x)
y = torch.randn(2, 10, 18, 12)
assert_allclose(traced(y), model(y))
```

If we script GlobLN instead, it's going to work.

mpariente:
`script_if_tracing` is not working on 1.6.0 because it returns a function, instead of calling the function. 
So we'll make a patch for it. (Works on nightly.)

mpariente:
Reading for merging IMO.

mpariente:
Added SuDORMRF networks as well finally. 
@mhu-coder @jonashaag @popcornell @faroit any feedback appreciated.

asteroid/models/base_models.py
```diff
@@ -314,10 +315,7 @@ def forward(self, wav):
         decoded = self.postprocess_decoded(decoded)
 
         reconstructed = pad_x_to_y(decoded, wav)
-        if was_one_d:
-            return reconstructed.squeeze(0)
-        else:
-            return reconstructed
+        return reconstructed.squeeze(0) * was_one_d + reconstructed * (1 - was_one_d)
```
jonashaag, line 318: Sorry for being late with the review. This is the only change that looks a bit weird to me. Was this necessary for Jit compat? If so, I think we should at least add a comment, or even better move this logic into a function with a descriptive name.

mpariente: I don't think this was necessary, @mhu-coder can you tell? 
Why does it look weird?

mhu-coder: It was necessary for tracing but as a traced model would have replaced `was_one_d` with a constant (with the value that was seen during tracing actually).
However, if we move this output shaping block in its own function, I am sure we could keep the old more readable logic and add the `torch.jit.scripting_if_tracing` decorator added to it. So the changes would be something like:

```python
def forward(self, wav):
   # ...
   return shape_output(reconstructed, wav.ndim)

@torch.jit.scripting_if_tracing
def shape_output(reconstructed, wav_dim):  # really not sure about the name of the function
    if wav_dim == 1:
        return reconstructed.squeeze(0)
    return reconstructed
```

mpariente: Sounds better. Mind submitting a PR?

mhu-coder: sure, on it

jonashaag: Cool! It’s just a very uncommon way to write an if else clause :-) and also implicitly assumes 0/1 == True/False for Torch tensors

PR #238, opened 2020-08-23
[src] Improve BaseModel.separate (#236 follow up)

mpariente:
Add two arguments to `BaseModel.separate`. 

- We can now decide if we overwrite existing files, by default we don't. 
- We can also choose a `save_dir` where all the files will be saved. Instead of saving each file next to the original one. Default: previous behavior. If there are files with the same name, in different folders, this will break. But I don't think it's a big problem. 

- Removed the `*args` which are passed to `_separate` in the `separate`. This will avoid confusion about positional arguments and the new args but is not completely backward compatible. I wouldn't recommend using positional arguments anyway.


mpariente:
We don't add `torchaudio` for loading yet as there are some subtleties but we'll probably add it in the future.

PR #243, opened 2020-08-24
[src & tests] Add Hungarian matcher to solve the linear sum assignement in PITLossWrapper

mpariente:
This is WIP, but I think we could merge it pretty fast. 

@popcornell, if we were to change the `PITLossWrapper`, what would you like in the output? Currently the `min_loss_idx` in `find_best_perm` is really not useful because you need to rebuild the permutations to get the indices. 

We might have a backward-breaking release with the Filterbank `filters` from #237, maybe we could change that as well because I know it was blocking you in the past. 
I'll comment in the code and I'd like your feedback

### DIFF
--- a/asteroid/dsp/overlap_add.py
+++ b/asteroid/dsp/overlap_add.py
@@ -1,6 +1,6 @@
 import torch
 from scipy.signal import get_window
-from asteroid.losses import PITLossWrapper
+from asteroid.losses.pit_wrapper import PITReorder
 from torch import nn
 
 
@@ -176,8 +176,8 @@ def _reorder_sources(
         return -torch.sum(x.unsqueeze(1) * y.unsqueeze(2), dim=-1)
 
     # We maximize correlation-like between previous and current.
-    pit = PITLossWrapper(reorder_func)
-    current = pit(current, previous, return_est=True)[1]
+    pit = PITReorder(reorder_func)
+    current = pit(current, previous)
     return current.reshape(batch, frames)
 
 
--- a/asteroid/losses/pit_wrapper.py
+++ b/asteroid/losses/pit_wrapper.py
@@ -1,10 +1,11 @@
 from itertools import permutations
 import torch
 from torch import nn
+from scipy.optimize import linear_sum_assignment
 
 
 class PITLossWrapper(nn.Module):
-    """Permutation invariant loss wrapper.
+    r"""Permutation invariant loss wrapper.
 
     Args:
         loss_func: function with signature (targets, est_targets, **kwargs).
@@ -234,6 +235,10 @@ class PITLossWrapper(nn.Module):
         # Indexes and values of min losses for each batch element
         min_loss_idx = torch.argmin(loss_set, dim=1)
         min_loss, _ = torch.min(loss_set, dim=1, keepdim=True)
+
+        # Have a look here, would this be helful
+        batch_indices = torch.stack([perms[m] for m in min_loss_idx], dim=0)
+
         return min_loss, min_loss_idx
 
     @staticmethod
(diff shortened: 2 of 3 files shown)


asteroid/losses/pit_wrapper.py
```diff
@@ -234,6 +235,10 @@ def find_best_perm(pair_wise_losses, n_src, perm_reduce=None, **kwargs):
         # Indexes and values of min losses for each batch element
         min_loss_idx = torch.argmin(loss_set, dim=1)
         min_loss, _ = torch.min(loss_set, dim=1, keepdim=True)
+
+        # Have a look here, would this be helful
+        batch_indices = torch.stack([perms[m] for m in min_loss_idx], dim=0)
```
mpariente, line 240: These will be of shape `(batch, n_src)` and give, for each batch, the optimal permutation on the sources.

mpariente: And actually, in `reorder_sources`, we re-create them (which is a bit stupid ^^)

asteroid/losses/pit_wrapper.py
```diff
@@ -264,3 +269,14 @@ def reorder_source(source, n_src, min_loss_idx):
...
         return reordered_sources
+
+    @staticmethod
+    def find_best_perm_hungarian(pair_wise_losses: torch.Tensor, n_src):
+        # After transposition, dim 1 corresp. to sources and dim 2 to estimates
+        pwl = pair_wise_losses.transpose(-1, -2)
+        # Just bring the numbers to cpu(), not the graph
+        pwl_copy = pwl.detach().cpu()
+        # Loop over batch + row indices are always ordered for square matrices.
+        batch_indices = [linear_sum_assignment(pwl)[1] for pwl in pwl_copy]
```
mpariente, line 280: Same here. Optimal permutation indices per batch.

mpariente (commented):
I'm happy to have some feedback to change the variables returned by `find_best_perm`, I think it can be more clever.

asteroid/losses/pit_wrapper.py
popcornell: we return for each batch the optimal permutation `(batch, n_src) ` and `reorder_sources` will take the optimal permutation as argument. But this will require some re-factoring in a bunch of places

popcornell:
```python
pw_losses = PITLossWrapper.get_pw_losses(lambda x, y: torch.sum((x.unsqueeze(1) * y.unsqueeze(2))),
                                                 current[..., :self.window_size//2],
                                                 previous[..., -self.window_size//2:])
_, perms = PITLossWrapper.find_best_perm(pw_losses, self.sources)
 current = PITLossWrapper.reorder_source(current, self.sources, perms)
```
It can be already done in this way but ideally we would want to have it in just two steps

mpariente: Yes, some place will need to be refactored but the output would be more useful in this way right?

mpariente:
What about this for the `LambdaOverlapAdd`? 
```python
    pit = PITLossWrapper(lambda x, y: torch.sum((x.unsqueeze(1) * y.unsqueeze(2))))
    current = pit(current[..., :overlap_f], previous[..., -overlap_f:], return_est=True)[1]
```

It was nice to use the static methods, but it's actually not necessary, the class can reorder the estimates.

popcornell:
> What about this for the LambdaOverlapAdd?

it looks way cleaner IMO

mpariente:
Thanks, I'll change it now.

asteroid/dsp/overlap_add.py
```diff
@@ -154,13 +154,12 @@ def _reorder_sources(
     previous = previous.reshape(-1, n_src, frames)
 
     overlap_f = window_size - hop_size
-    pw_losses = PITLossWrapper.get_pw_losses(
-        lambda x, y: torch.sum((x.unsqueeze(1) * y.unsqueeze(2))),
```
mpariente, line 158: I realize now that this might be weird. 
It constructs a matrix of pointwise products of target vs. estimates and makes the total sum, even with the batch.. I think it is wrong, what was the intention @popcornell ?

asteroid/dsp/overlap_add.py
```diff
@@ -154,13 +154,12 @@ def _reorder_sources(
...
-        current[..., :overlap_f],
-        previous[..., -overlap_f:],
+    pit = PITLossWrapper(
+        lambda x, y: torch.sum(
+            (x[..., :overlap_f].unsqueeze(1) - y[..., -overlap_f:].unsqueeze(2)).abs(), dim=-1
```
mpariente, line 159: Here we directly compute the pairwise matrix of absolute difference. We could do something else (squared difference, or correlation) if we want though.

mpariente (commented):
I realize now that the formula for the reordering is actually pretty weird. 

This PR will be merged for 0.4.0 but I think we should fix the `LambdaOverlapAdd` on a version before that. @popcornell can you have a look please?

asteroid/dsp/overlap_add.py
mpariente: Should be solved in #271

PR #256, opened 2020-09-20
Add support for vocal mixtures from MedleyDB

saurjya:
This is the first commit with a working recipe. Allows parsing MedleyDB metadata to filter subset of tracks to use for creating mixtures.


mpariente:
Thanks a lot for the PR, this is great ! I'll take time in the upcoming week to review it. 
One thing though, your commit is not signed which makes that you won't appear as contributor. Can you try to open the same PR (or force-push to this one) with a signed commit please? That would be cool, thanks.

PR #224, opened 2020-08-20
DCUNet and DCCRNet

jonashaag:
This is a WIP implementation of [Phase-aware Single-stage Speech Denoising and Dereverberation with U-Net](https://arxiv.org/abs/2006.00687).

I also want to implement [DCCRN: Deep Complex Convolution Recurrent Network for Phase-Aware Speech Enhancement](https://arxiv.org/abs/2008.00264) soon.

I'd like to get feedback on

- Usage of PyTorch 1.7+ native complex128 dtype rather than Asteroid style complex type or torchaudio style complex type.
- Code structure, class names, code placing. Many things I wasn't sure where to place in the Asteroid codebase.
- Location of complex norms, activations, etc.
- Dealing with input shapes in `DCUMaskNet.forward`. Due to the encoder/decoder structure only specific T/F dimensions will work without padding or other workarounds.

PR #266, opened 2020-10-01
Split base model

jonashaag:
Part 2 of #224

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ docstrings in the codebase for examples.
 
 ### Coding style
 
-We use [pre-commit hooks][../.pre-commit-config.yaml] to format the code using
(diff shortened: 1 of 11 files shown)


mpariente:
I know it was my idea but I think the name BaseEMDNet is not that clear. 
Do we plan that people will discover it easily because it's the baseclass of all the models we use for now? Otherwise, the name doesn't speak for itself.

jonashaag:
Unfortunately I don't have a better naming suggestion. Maybe `BaseEncoderMaskerDecoderNet` is a bit better. But yes, I would expect people to find and understand the class pretty quickly despite its name.

mpariente:
Maybe `BaseEncoderMaskerDecoder`? We don't really need the `Net`.

mpariente (approved):
Renamed to `BaseEncoderMaskerDecoder` instead, but otherwise it's perfect. 
Waiting for the CI to complete.

PR #268, opened 2020-10-02
[src&tests] Improve scheduler's docs + add plot method

mpariente:
- `plot` can be useful for the lazy ones (me included)
- It was a bit undiscoverable till now. It's still pretty undiscoverable but at least we mention it in the docs.

### DIFF
--- a/asteroid/engine/schedulers.py
+++ b/asteroid/engine/schedulers.py
@@ -1,3 +1,4 @@
+import torch
 from torch.optim.optimizer import Optimizer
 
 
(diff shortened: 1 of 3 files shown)


asteroid/engine/schedulers.py
```diff
@@ -128,3 +146,12 @@ def _get_lr(self):
...
         return lr
+
+    def as_tensor(self, start=0, stop=None):
+        stop = stop or self.warmup_steps * 2
+        lr_list = []
+        for _ in range(start, stop):
+            self.step_num += 1
+            lr_list.append(self._get_lr())
+        self.step_num = 0
+        return torch.tensor(lr_list)
```
mpariente, line 157:
```suggestion
```

popcornell:
Looks good to me, pretty useful addition !!

PR #271, opened 2020-10-06
[src] Fix LambdaOverlapAdd

mpariente:
The previous loss function to reorder the chunks was weird, this should fix it. 

Which loss function would make the most sense? Currently it's kind of unnormalized correlation. 

Also, we probably need a better unit test for this. 

@groadabike, could you checkout this branch and check if the reordering works better in your case please? That'd be great

### DIFF
--- a/asteroid/dsp/overlap_add.py
+++ b/asteroid/dsp/overlap_add.py
@@ -165,12 +165,18 @@ def _reorder_sources(
     previous = previous.reshape(-1, n_src, frames)
 
     overlap_f = window_size - hop_size
+
+    def reorder_func(x, y):
+        x = x[..., :overlap_f]
+        y = y[..., -overlap_f:]
+        # Mean normalization
+        x = x - x.mean(-1, keepdim=True)
+        y = y - y.mean(-1, keepdim=True)
+        # Negative mean Correlation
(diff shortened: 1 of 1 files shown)


faroit:
@mpariente while you are at it, it had trouble understanding the purpose of `reorder_chunks` when I glanced at the function. Maybe adding more detail to the docstring and a typical use-case would help others?

mpariente:
Ok, on it. Thanks for the feedback.

asteroid/dsp/overlap_add.py
```diff
@@ -5,15 +5,24 @@
...
-        reorder_chunks (bool): whether to reorder each consecutive segment.
+        hop_size (int): Segmentation hop size.
+        window (str): Name of the window (see scipy.signal.get_window) used
+            for the synthesis.
+        reorder_chunks (bool): Whether to reorder each consecutive segment.
+            This might be useful when `nnet` is permutation invariant, as
+            source assignements might change output channel from one segment
+            to the next (in classic speech separation for example).
+            Reordering is performed based on the correlation between
+            the overlapped part of consecutive segment.
```
faroit, line 25: reads well!

faroit (approved):

mpariente:
Still needs to be tested in real scenarios but that should work better now.

Issue #264, opened 2020-09-30
Add required sample_rate arg to all models 


mpariente:
This is needed to know which files you can hope to enhance/separate. 
- Should filterbanks all have a sample_rate argument as well? Because they are also tied to a sampling rate. 
- Should it be only throught the filterbanks? 
- Should it be in `BaseModel`?
- Should it be in both? 

It's not clear yet.

mpariente:
Closed by #274 #283 and #284

PR #274, opened 2020-10-11
[src&tests] Add sample_rate in `BaseEncoderMaskerDecoder` 

mpariente:
We need a way to know what is the sampling rate supported by a model. That's one attempt to do it. 

Should we enforce a `samle_rate` argument in the `BaseModel`? What would be the pros and cons? 

Is it the good place to put it in the `Filterbank` + `_EncDec`? An alternative is to place it in `BaseEncoderMaskerDecoder` (and all the child classes) but some filterbanks already have a `sample_rate` argument (param sinc and MPGT) and there would be some conflict on that. 

Any feedback is appreciated on that as it will impact pretrained models and automatic resampling in `asteroid-infer` CLI.

Linked issue #264

### DIFF
--- a/asteroid/filterbanks/enc_dec.py
+++ b/asteroid/filterbanks/enc_dec.py
@@ -42,6 +42,7 @@ class Filterbank(nn.Module):
             "n_filters": self.n_filters,
             "kernel_size": self.kernel_size,
             "stride": self.stride,
+            "sample_rate": self.sample_rate,
         }
         return config
 
--- a/asteroid/filterbanks/enc_dec.py
+++ b/asteroid/filterbanks/enc_dec.py
@@ -13,12 +13,14 @@ class Filterbank(nn.Module):
         kernel_size (int): Length of the filters.
         stride (int, optional): Stride of the conv or transposed conv. (Hop size).
             If None (default), set to ``kernel_size // 2``.
+        sample_rate (int): Sample rate of the expected audio.
+            Defaults to 8000.
 
     Attributes:
         n_feats_out (int): Number of output filters.
     """
 
-    def __init__(self, n_filters, kernel_size, stride=None):
+    def __init__(self, n_filters, kernel_size, stride=None, sample_rate=8000):
         super(Filterbank, self).__init__()
         self.n_filters = n_filters
         self.kernel_size = kernel_size
@@ -26,6 +28,7 @@ class Filterbank(nn.Module):
         # If not specified otherwise in the filterbank's init, output
         # number of features is equal to number of required filters.
         self.n_feats_out = n_filters
+        self.sample_rate = sample_rate
 
     @property
     def filters(self):
--- a/asteroid/filterbanks/enc_dec.py
+++ b/asteroid/filterbanks/enc_dec.py
@@ -66,6 +66,7 @@ class _EncDec(nn.Module):
     def __init__(self, filterbank, is_pinv=False):
         super(_EncDec, self).__init__()
         self.filterbank = filterbank
+        self.sample_rate = getattr(filterbank, "sample_rate", None)
         self.stride = self.filterbank.stride
         self.is_pinv = is_pinv
 
--- a/asteroid/filterbanks/enc_dec.py
+++ b/asteroid/filterbanks/enc_dec.py
@@ -13,14 +13,14 @@ class Filterbank(nn.Module):
         kernel_size (int): Length of the filters.
         stride (int, optional): Stride of the conv or transposed conv. (Hop size).
             If None (default), set to ``kernel_size // 2``.
-        sample_rate (int): Sample rate of the expected audio.
+        sample_rate (float): Sample rate of the expected audio.
             Defaults to 8000.
 
     Attributes:
         n_feats_out (int): Number of output filters.
     """
 
-    def __init__(self, n_filters, kernel_size, stride=None, sample_rate=8000):
+    def __init__(self, n_filters, kernel_size, stride=None, sample_rate=8000.0):
         super(Filterbank, self).__init__()
         self.n_filters = n_filters
         self.kernel_size = kernel_size
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -20,6 +20,17 @@ def _unsqueeze_to_3d(x):
 
 
 class BaseModel(nn.Module):
+    """Base class for serializable models.
+
+    Defines saving/loading procedures as well as separation methods from
+    file, torch tensors and numpy arrays.
+    Need to overwrite the `foward` method, the `sample_rate` property and
+    the `get_model_args` method.
+    For models whose `forward` doesn't return waveform tensors,
+    overwrite `_separate` to return waveform tensors.
+
+    """
+
     def __init__(self):
         super().__init__()
 
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -224,7 +224,7 @@ class BaseEncoderMaskerDecoder(BaseModel):
         self.decoder = decoder
         self.encoder_activation = encoder_activation
         self.enc_activation = activations.get(encoder_activation or "linear")()
-        self.sample_rate = getattr(encoder, "sample_rate")
+        self.sample_rate = getattr(encoder, "sample_rate", None)
 
     def forward(self, wav):
         """Enc/Mask/Dec model forward
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -222,9 +222,9 @@ class BaseEncoderMaskerDecoder(BaseModel):
         self.encoder = encoder
         self.masker = masker
         self.decoder = decoder
-
         self.encoder_activation = encoder_activation
         self.enc_activation = activations.get(encoder_activation or "linear")()
+        self.sample_rate = getattr(encoder, "sample_rate")
 
     def forward(self, wav):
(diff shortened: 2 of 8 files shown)


mhu-coder:
I only had a quick look at the changes but I wondered if the sample rate added to the models should also be used during training and inference as well. In other words:
* for training: either the sample rate of the model is set to that of the training data or the training data are resampled to that of the model before being fed to the network. In the latter case, also have a check over the whole training set before starting to warn about any audio file that has a sample rate smaller than what is expected by the model?
* for inference: would it make sense to constantly resample the data to that of the model?

About where to put the sample rate, IMHO, it makes sense to have it in the encoder as some of the filter banks actually need this piece of information to be instantiated.

asteroid/models/base_models.py
```diff
@@ -222,9 +222,9 @@ def __init__(self, encoder, masker, decoder, encoder_activation=None):
         self.encoder = encoder
         self.masker = masker
         self.decoder = decoder
-
         self.encoder_activation = encoder_activation
         self.enc_activation = activations.get(encoder_activation or "linear")()
+        self.sample_rate = getattr(encoder, "sample_rate", None)
```
mhu-coder, line 227: Since the attribute is supposed to be read-only anyway, would having `sample_rate` as a property, like
```python
    @property
    def sample_rate(self):
        return get_attr(self.encoder, "sample_rate", None)
```
help in consistency?

asteroid/filterbanks/enc_dec.py
```diff
@@ -62,6 +66,7 @@ class _EncDec(nn.Module):
     def __init__(self, filterbank, is_pinv=False):
         super(_EncDec, self).__init__()
         self.filterbank = filterbank
+        self.sample_rate = getattr(filterbank, "sample_rate", None)
```
mhu-coder, line 69: Same as for `BaseModel`, would it be better to have a property instead?

```python
@property
def sample_rate(self):
    return getattr(self.filterbank, "sample_rate", None)
```

mpariente:
Thanks for the feedback. 

For training, we pass the waveform to the model, and it doesn't include the sampling rate. The user will have to specify the sampling rate in both the model and the DataLoader. We have to assume they are consistent. 

For inference, yes it does. When a filename is given, we can do it without changes. When an array/tensor is given into `separate`, we cannot infer the sampling rate, so we could add an argument to enable resampling.

mhu-coder (commented):
Just some feedback after a quick look. I haven't had time to check consistency between the classes in asteroid yet.

asteroid/models/base_models.py
mpariente: Thanks for the feedback. I don't think a property is needed here. You'd need a property if you expect the sample_rate to change in the encoder, so that it can be dynamically retrieved, right?

mhu-coder: True. Having a property which redirects to the value stored in the encoder allows to have one fewer attribute in `BaseEncoderMaskerDecoder`, ensures consistency between the value in `BaseEncoderMaskerDecoder` and the encoder and prevents silly mistakes like setting `sample_rate` on instances of `BaseEncoderMaskerDecoder` to a different value. I know, the latter case is not supposed to happen, but why take the risk?

jonashaag:
On a conceptual level, are models and filterbanks bound to a specific sample rate? Both yes, right? I mean in theory you can use a set of (n_fft, hop_size) on any sample rate, but it means something very different for different sample rates, so it's not actually the same thing, right? And for models, a model trained with a specific sample rate cannot really be expected to work with other sample rates even if the filterbank is exchanged for a filterbank that correctly deals with the sample rate, right? Because models learn that some frequencies should be dealt with differently than other frequencies.

So in conclusion I think we should have sample rate argument on model, filterbank (and also datasets). And maybe even add sanity checks to the model code that the filterbank has the same sample rate as the model. You can also think of a filterbank as a function of type

```
filterbank: (sr1: int, audio: float array [n_src, n_samples]) -> (sr2: int, audio_repr: float array [n_src, n_dim1, n_dim2])
```

and a model

```
model: (sr2: int, audio_repr: float array [n_src_in, n_dim1, n_dim2]) -> audio_repr: float array [n_src_out, n_dim1, n_dim2]
```

This could even be type checked statically...

mpariente:
Thanks for the feedback !

I think there are some edge cases where a model, or filterbank, can support multiple sampling rate, but that's not the purpose of the PR. 

Can we agree on the following: 
- Any source separation model needs a sample rate argument/attribute. 
- Nothing in `masknn` depends on sample rate. The maskers can be used on Mel, MFCC or non audio data, they should work. 

Then I would propose: 
- In the special case of the encoder-masker-decoder framework, the model's sample rate is defined through the filterbank. 
- Other source separation models without filterbanks will define the sample rate in a different way (that's why I didn't implement anything in the `BaseModel` yet). 

One last thing regarding the `@property` argument. It could be `NotImplemented` in `BaseModel` to force child classes to define it, that's one pro. OTHO, there was this torchscript issue with properties and I'm not sure where this would stand.

jonashaag:
>     * Nothing in `masknn` depends on sample rate. The maskers can be used on Mel, MFCC or non audio data, they should work.

I don't think this is true unless we're talking about different definitions of "it works". I mean it shouldn't crash but aren't the results going to be very different if you train something with a different sample rate or even TF representation than you're using for inference? Like, if I train a masknet using an STFT representation of shape `(f, t)`, and then swap then put in Mel representation that happens to have the same shape, the `f` bins are going to have a different meaning, and the masknet will do random things. That is to say, generally speaking, a trained masknet is only ever going to be useful with the same input sample rate and the same filterbank that was used for training. Am I confusing things here?

jonashaag:
>     * In the special case of the encoder-masker-decoder framework, the model's sample rate is defined through the filterbank.

Agreed

jonashaag:
I agree with making `sample_rate` a read only property of `BaseEncoderMaskerDecoder`, or a read only getter function, or something like that; just not an actual property. Or simply leave it out and access sample rate using `.filterbank.sample_rate`.

mpariente:
> Am I confusing things here?

No, this is correct. But this goes way further than just the sample rate. Just knowing the sample rate will lead you nowhere, that's why I think it doesn't make sense to include it there (in `masknn`). 
OTOH, for the filterbank, if you know the sample rate, you can use it, same goes for the total model. 

> I agree with making sample_rate a read only property of BaseEncoderMaskerDecoder, or a read only getter function, or something like that; just not an actual property. Or simply leave it out and access sample rate using .filterbank.sample_rate.

This is not clear to me sorry, can you elaborate please?

jonashaag:
> Just knowing the sample rate will lead you nowhere, that's why I think it doesn't make sense to include it there (in `masknn`).
> OTOH, for the filterbank, if you know the sample rate, you can use it, same goes for the total model.

Makes a lot of sense! Agreed!

> This is not clear to me sorry, can you elaborate please?

Yeah sorry, was a bit confusing indeed. What I meant is that we should define an interface for getting a `BaseModel`'s sample rate so that we can use it in `separate()`. We could make it a read only attribute, then we can use a property in the EMD base class. Or we could make it a read only getter function.

mpariente:
> Yeah sorry, what I wrote didn't make sense indeed. What I meant is that we should define an interface for getting a `BaseModel`'s sample rate so that we can use it in `separate()`. We could make it a read only attribute, then we can use a property in the EMD base class. Or we could make it a read only getter function.

Is this what you mean? Or something else? 
```python
class BaseModel():
    @property
    def sample_rate(self):
        raise NotImplementedError

class BaseEMD(BaseModel):
    @property
    def sample_rate(self):
        return getattr(self.encoder, "sample_rate", None)
```

jonashaag:
Yep!

mpariente:
Well, I guess @mhu-coder will also be happy with the `property` :wink:

mpariente:
I'd like to restrict the PR to this, this seems ok? 

Follow-up PRs will 
- Use the sample rate in the `separate` methods to automatically resample input audio (maybe output audio as well, to be decided). 
- Add a sample rate argument to all the models for transparency. This will somehow be backward breaking: previous models saved for 16k will have 8k (if we set it as default) as sample rate  if not specified and will resample when inference from file. We have only one pretrained model at fs!=8k, DeMask so we are ok, but we'll need to notify users.

jonashaag:
Yes, looks ok!

Backwards compat for adding sample rate: I suggest to raise an error for any models that don’t have a sample rate argument, and have a list of known pretrained models hard coded into the Asteroid code for which we can set the sample rate automatically. It’s not nice, but we are breaking backwards compat anyways and then at least users are very explicitly told what to do.

mpariente:
I thought about that, with a warning, but yeah, sounds good.

mhu-coder:
Looks good to me too.

In addition to your list of changes for the follow-up PRs, is there a need to indicate somewhere that a sample rate has to be specified during training (if not equal to 8kHz)? We could ensure consistency between the model and the dataloader in term of sample rate if every dataloader had a `sample_rate` attribute but changing all the dataloaders may be a lot of work...

mpariente:
> We could ensure consistency between the model and the dataloader in term of sample rate

That makes complete sense. I guess we would initialize the network with the sample rate value from the dataloader. 
Adding sample rate argument to dataloaders is easy, what is cumbersome would be to change all the recipes to include the sample rate in it. 
We're planning to change the recipe format so I'm not sure I want to invest the time to add the sample rate to all models in the recipes..

PR #276, opened 2020-10-12
Fix DCCRN and DCUNet-Large

jonashaag:
This is the real DCCRN architecture. I got some of the details wrong in the initial version.

- Are we simply going to release the fixed version without any backwards compat? I think that's reasonable since we have no pretrained models so far and it's only been a few days since the release.
- The code the concatenative skip connection with the intermediate layer is a draft. I'm not sure how to do it in a clean way just yet. The way the intermediate layer works in this new version requires that we are aware of its existence in the u-net params calculation, so we can't really keep them entirely separate anymore. Maybe we have to add the concept of this concatenative intermediate layer to the u-net param calculation code. If you have any ideas that would be helpful as well!


jonashaag:
@kssk16 Note that our LSTM is of size 640 rather than 512 because we're not stripping the Nyquist frequency so have 257 rather than 256 frequency bins.

Maybe we should change this as well so the implementation matches the paper.

kssk16:
> @kssk16 Note that our LSTM is of size 640 rather than 512 because we're not stripping the Nyquist frequency so have 257 rather than 256 frequency bins.
> 
> Maybe we should change this as well so the implementation matches the paper.

@jonashaag Thanks for letting me know.

jonashaag:
More differences:

- They use PReLu, not Leaky ReLu
- They use (2, 1) padding for all encoders and (2, 0) padding for all decoders
- In the output layer they use output_padding=(1, 0)
- They use a window of size 400, pad it (centered) to 512 and then use a 512 FFT. We don't do the 400 thing.

jonashaag:
Note: New WIP version here https://github.com/jonashaag/asteroid/blob/b053ed67806da48270742ff01202d158a3faaef8/asteroid/masknn/recurrent.py#L549 https://github.com/jonashaag/asteroid/blob/b053ed67806da48270742ff01202d158a3faaef8/asteroid/models/dccrnet.py

jonashaag:
OK, this should be ready to review.

The DCCRN implementation isn't 100% identical to the paper one, but very similar. Differences I know of: using 512 FFT instead of 400 + 2 * 56 zero padding; slightly different padding in T dimension.

While fixing the DCCRN implementation I realized that you don't need the `intermediate_layer` concept because you can simply replace it by an "asymmetric" u-net structure where the deepest layer's "partner" is the identity. In fact, the `intermediate_layer` concept doesn't actually fit the DCCRN architecture, so I removed it.

I also included a fix to the DCUNet architectures that I realized was wrong.

In general, I'd appreciate if someone could compare my implementations to the papers because I'm quite unsure if there are more differences that I didn't spot. Code can be found here https://paperswithcode.com/paper/dccrn-deep-complex-convolution-recurrent-1 https://paperswithcode.com/paper/phase-aware-speech-enhancement-with-deep-1

Commit fc0232cdf9892de50ecdaff23183da293776ccf4 is sort of unrelated, we can exclude it.

mpariente:
Maybe @kssk16 @huyanxin @sweetcocoa and @chanil1218 would like to have a look? That would be nice!

asteroid/models/dccrnet.py
```diff
@@ -25,6 +28,18 @@ def __init__(
...
+        tf_rep = self.encoder(wav)
+        # Remove Nyquist frequency bin
+        return complex_nn.as_torch_complex(tf_rep)[..., :-1, :]
+
+    def apply_masks(self, tf_rep, est_masks):
+        masked_tf_rep = est_masks * tf_rep.unsqueeze(1)
+        # Pad Nyquist frequency bin
+        return from_torchaudio(
+            torch.view_as_real(torch.nn.functional.pad(masked_tf_rep, (0, 0, 0, 1)))
+        )
```
mpariente, line 45: Very clear.

asteroid/models/dcunet.py
```diff
@@ -17,7 +17,7 @@ class BaseDCUNet(BaseEncoderMaskerDecoder):  # CHECK-JIT
 
     """
 
-    masknet_class = DCUMaskNet
+    masknet_class = NotImplemented
```
mpariente, line 20: What happened here?

tests/jit/jit_models_test.py
```diff
@@ -46,14 +48,11 @@ def small_model_params():
...
-            "n_layers": 1,
-            "dropout": 0.0,
-            "n_filters": 32,
-            "kernel_size": 32,
-            "stride": 16,
+        DCCRNet.__name__: {
+            "architecture": "DCCRN-CL",
+        },
+        DCUNet.__name__: {
+            "architecture": "DCUNet-10",
```
mpariente, line 55: Is it possible to make "mini" architectures for this in their respective files? For tests, demos on collab etc.. That would be cool.

tests/jit/jit_models_test.py
```diff
@@ -178,7 +194,7 @@ def get_default_device():
         "n_filters": 32,
         "kernel_size": 21,
     }
-    model = model_def(**params, fb_name="free")
+    model = model_def(**params)
```
mpariente, line 197: Oh, I'll need to remove that after this is merged..

tests/jit/jit_models_test.py
```diff
@@ -90,26 +98,35 @@ def small_model_params():
     return params
 
 
+@pytest.mark.parametrize(
+    "model_def",
+    (
+        DCCRNet,
+        DCUNet,
```
mpariente, line 105: Ohh great that this is passing !

tests/jit/jit_models_test.py
```diff
@@ -90,26 +98,35 @@ def small_model_params():
...
-        (torch.rand(240) - 0.5) * 2,
-        (torch.rand(1, 220) - 0.5) * 2,
-        (torch.rand(4, 256) - 0.5) * 2,
-        (torch.rand(1, 1, 301) - 0.5) * 2,
-        (torch.rand(3, 1, 290) - 0.5) * 2,
+        (torch.rand(4740) - 0.5) * 2,
+        (torch.rand(1, 4720) - 0.5) * 2,
+        (torch.rand(4, 4756) - 0.5) * 2,
+        (torch.rand(1, 1, 4801) - 0.5) * 2,
+        (torch.rand(3, 1, 4790) - 0.5) * 2,
```
mpariente, line 116: Can you explain the new shapes please? 
It's to give the name number of frames of STFT but have different wav size? Or what is the motivation?

tests/jit/jit_models_test.py
```diff
@@ -67,6 +66,15 @@ def small_model_params():
             "n_filters": 32,
             "kernel_size": 32,
         },
+        LSTMTasNet.__name__: {
+            "n_src": 2,
+            "hid_size": 4,
+            "n_layers": 1,
```
mpariente, line 72: Better test if there are two layers I guess
```suggestion
            "n_layers": 2,
```

mpariente (commented):
This looks great IMO ! 
I didn't check the implementation details and the correspondence with the original code very thoroughly though.

asteroid/complex_nn.py
```diff
@@ -31,10 +31,13 @@ def is_torch_complex(x):
...
 
 def torch_complex_from_magphase(mag, phase):
-    return as_torch_complex((mag * torch.cos(phase), mag * torch.sin(phase)))
+    return torch.view_as_complex(
+        torch.stack((mag * torch.cos(phase), mag * torch.sin(phase)), dim=-1)
+    )
 
 
-def as_torch_complex(x, asteroid_dim=-2):
+@torch.jit.script
```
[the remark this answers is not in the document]
jonashaag: OK I'm not very familiar with all of this. I just added this to make the tests pass 😅 Please change to whatever is better!

asteroid/models/dcunet.py
jonashaag: It was bad in the initial implementation; the subclasses override this attribute and it doesn't make sense to have it in the parent class IMO

tests/jit/jit_models_test.py
jonashaag: Makes sense, I will try

tests/jit/jit_models_test.py
jonashaag: The problem is that DCUNet only works with certain shapes (T - 1 must be divisible by 16). I found these by trial and error. It's the same shapes as elsewhere + 4500.

asteroid/complex_nn.py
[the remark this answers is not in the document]
mpariente: No problem. I didn't know anything about it until recently. 
I'll check what was the error.

tests/jit/jit_models_test.py
mpariente: Ok, thanks. This is what I thought. 
Would it make sense to handle the "slicing" in the model? So that we don't have to be careful about it everywhere else? 

I'm worried that these tests pass because the effective input size, after the STFT, is the same. In this case, the scripting unit test is not useful.

mpariente: I'll have a deeper look at that asap

mpariente: This is what I had in mind. All the inputs have the same shape once out of the STFT, so this test ensure that if the input has the same shape, scripting keeps consistency. This is already good, but not enough. 
I'll keep digging

mpariente:
If I print in the BaseUNet's forward in this way
```python
    def forward(self, x):
        enc_outs = []
        for idx, enc in enumerate(self.encoders):
            x = enc(x)
            enc_outs.append(x)
            print(f"Encoding {idx}", x.shape)
        for idx, (enc_out, dec) in enumerate(zip(reversed(enc_outs[:-1]), self.decoders)):
            x = dec(x)
            print(f"Decoding {idx}", x.shape)
            x = torch.cat([x, enc_out], dim=1)
        return self.output_layer(x)
```

I get
```
print(tf_rep.shape) torch.Size([1, 256, 36])
Encoding 0 torch.Size([1, 16, 128, 35])
Encoding 1 torch.Size([1, 32, 64, 34])
Encoding 2 torch.Size([1, 64, 32, 33])
Encoding 3 torch.Size([1, 128, 16, 32])
Encoding 4 torch.Size([1, 128, 8, 31])
Encoding 5 torch.Size([1, 128, 4, 30])
Encoding 6 torch.Size([1, 128, 4, 30])
Decoding 0 torch.Size([1, 128, 4, 30])
Decoding 1 torch.Size([1, 128, 8, 31])
Decoding 2 torch.Size([1, 128, 16, 32])
Decoding 3 torch.Size([1, 64, 32, 33])
Decoding 4 torch.Size([1, 32, 64, 34])
Decoding 5 torch.Size([1, 16, 128, 35])
```

where the time dimension is not divided by two at every iteration, but just looses one. Is this intended? At least this doesn't seem consistent with the docstring and also your comments that suggest that the time dim has to be divisible by the prod of strides. Are the strides on both freq and time dimension? 

By the way, for DCCRNet, the jit test passes even with different size, which is great news !

asteroid/complex_nn.py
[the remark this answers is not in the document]
mpariente: Fixed

jonashaag:
Could it be that you confused DCUNet and DCCRNet here?

```
$ python -c 'import asteroid.models as m, torch; m.DCUNet("DCUNet-10")(torch.randn(1,50000))'
Encoding 0 torch.Size([1, 32, 129, 97])
Encoding 1 torch.Size([1, 64, 65, 49])
Encoding 2 torch.Size([1, 64, 33, 25])
Encoding 3 torch.Size([1, 64, 17, 13])
Encoding 4 torch.Size([1, 64, 9, 13])
Decoding 0 torch.Size([1, 64, 17, 13])
Decoding 1 torch.Size([1, 64, 33, 25])
Decoding 2 torch.Size([1, 64, 65, 49])
Decoding 3 torch.Size([1, 32, 129, 97])
...
```

The paddings are different in DCUNet. In DCCRNet we use the "official" padding (or very similar), that is, the encoders each lose 1 frame in the T dim and the decoders add it back using output padding.

In DCUNet, there is not official implementation available, so I used what's most common with other implementations, which seems to be padding so that for each encoder, `T -> (T+1)/2` roughly, for example `17 -> 9 -> 5 -> 3 -> 2 -> 2`. We can also use some other padding, I'm not an expert in u-nets (not expert in anything in deep learning really 😅)

mpariente:
It is completely possible, yes ! 
I'm not an expert in either of these architectures and didn't read the papers in detail recently. 

Given that DCCRNet is much more flexible, I guess it makes sense to isolate the tests of DCUNet, right? 
I'll do that

mpariente:
IMO, this should be good. 

I found the correct durations for DCUNet by hand as well, so we test 3 shapes and the traced model is consistent, which is great ! 

Few things:
- Having to guess input shape is very bad, and the error message is half helpful because it is in the frequency domain, so it doesn't directly help to get the right shape in the time domain. How can we improve that? Do we want to improve that? 
- Those tests are pretty slow, so it would be nice to have "mini" architectures. 

image

Extending DCCRNet to separation shouldn't be too hard, we just need to overwrite the second entry of the last line in the architecture file that is past to the complex conv, right. So we could easily add a `n_src` arg to it and enable separation. Am I missing something? 

image

tests/jit/jit_models_test.py
mpariente: I did it. Have a look if you can.

mpariente:
I made "mini" architecture. 

Still wondering about the guessing of the input shape. 
Maybe padding trimming the time axis and raising a warning would be better? WDYT?

mpariente:
In the line of: 
```patch
Index: asteroid/masknn/convolutional.py
<+>UTF-8
===================================================================
--- asteroid/masknn/convolutional.py	(revision dd47fce87c6d72878c6fffc38f14b1ef74ff1c27)
+++ asteroid/masknn/convolutional.py	(date 1606073650145)
@@ -516,15 +516,29 @@
             **kwargs,
         )
 
-    def _check_input_dims(self, x):
-        # TODO: We can probably lift the shape requirements once Keras-style "same"
-        # padding for convolutions has landed: https://github.com/pytorch/pytorch/pull/42190
+
+    def fix_input_dims(self, x):
         freq_prod, time_prod = self.encoders_stride_product
-        if (x.shape[1] - 1) % freq_prod or (x.shape[2] - 1) % time_prod:
+        if (x.shape[1] - 1) % freq_prod:
             raise TypeError(
                 f"Input shape must be [batch, freq + 1, time + 1] with freq divisible by "
-                f"{freq_prod} and time divisible by {time_prod}, got {x.shape} instead"
+                f"{freq_prod}, got {x.shape} instead"
             )
+        if (x.shape[2] - 1) % time_prod:
+            div, mod = divmod((x.shape[2] - 1), time_prod)
+            pad_shape = [0, time_prod - mod]
+            x = nn.functional.pad(x, pad_shape, mode="constant")
+        return x
+
+    # def _check_input_dims(self, x):
+    #     # TODO: We can probably lift the shape requirements once Keras-style "same"
+    #     # padding for convolutions has landed: https://github.com/pytorch/pytorch/pull/42190
+    #     freq_prod, time_prod = self.encoders_stride_product
+    #     if (x.shape[1] - 1) % freq_prod or (x.shape[2] - 1) % time_prod:
+    #         raise TypeError(
+    #             f"Input shape must be [batch, freq + 1, time + 1] with freq divisible by "
+    #             f"{freq_prod} and time divisible by {time_prod}, got {x.shape} instead"
+    #         )
 
 
 class SuDORMRF(nn.Module):
Index: asteroid/masknn/base.py
<+>UTF-8
===================================================================
--- asteroid/masknn/base.py	(revision dd47fce87c6d72878c6fffc38f14b1ef74ff1c27)
+++ asteroid/masknn/base.py	(date 1606072909225)
@@ -92,8 +92,9 @@
         )
 
     def forward(self, x):
-        self._check_input_dims(x)
+        x = self.fix_input_dims(x)
         return super().forward(x.unsqueeze(1))
 
-    def _check_input_dims(self, x):
+    def fix_input_dims(self, x):
         """Overwrite this in subclasses to implement dimension checks."""
+        return x
```

which doesn't work because the padding doesn't match the `tf_rep` when applying the mask but we could also correct for that without problems.

jonashaag:
> So we could easily add a `n_src` arg to it and enable separation. Am I missing something?

I think so, yes! Probably also applies to DCUNet.

> Maybe padding trimming the time axis and raising a warning would be better?

I like the padding/trimming approach but I dislike warnings when you can raise an error instead. I suggest to make things explicit: By default it raises an error that tells you to enable padding/trimming or fix your input size. PoC, maybe the choice between padding and trimming is over engineered?

```py
# time_prod
strideprod = 16
# x.shape[1]
l = 35
# New self.fix_length parameter to __init__
fix_length = None  # or "pad" or "trim"

l_remainder = (l - 1) % strideprod
if l_remainder:
    if fix_length == "trim":
        l -= l_remainder
        print(f"Trimmed length to {l}")
    elif fix_length == "pad":
        l += strideprod - l_remainder
        print(f"Padded length to {l}")
    else:
        print('Invalid shape, use fix_length="trim" or fix_length="pad"')
```

mpariente (commented):
Forgot to validate those comments this morning..

mpariente:
I understand. 
In the current code, I don't raise a warning, I just pad. Feel free to adapt if how you'd like.

asteroid/masknn/_dcunet_architectures.py
```diff
@@ -14,14 +14,40 @@ def make_unet_encoder_decoder_args(encoder_args, decoder_args):
...
+            for in_chan, out_chan, kernel_size, stride, padding, output_padding in decoder_args
+        ]
 
     return encoder_args, decoder_args
 
 
 # fmt: off
 
 DCUNET_ARCHITECTURES = {
+    "DCUNet-MINI": make_unet_encoder_decoder_args(
```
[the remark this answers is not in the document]
jonashaag: Yes, "mini" or "test" or "Asteroid-Mini" or ...

In any case we should add a comment that this isn't in the paper.

jonashaag:
> In the current code, I don't raise a warning, I just pad. Feel free to adapt if how you'd like.

OK, working on it now!

jonashaag:
Made some changes, what do you think? It didn't have time to look at the tests yet, but the tests work for both fix modes, and fail for None.

Btw there is a test failure related to Torch complex number support (already fixed in Torch nightly)

mpariente:
Do you have a link to the issue or the PR? 
Maybe we should run tests on torch-nightly as well? 
I'll have a look at the changes.

asteroid/utils/torch_utils.py
```diff
@@ -119,6 +119,42 @@ def pad_x_to_y(x: torch.Tensor, y: torch.Tensor, axis: int = -1) -> torch.Tensor
...
+
+    Returns:
+        torch.Tensor, `x` padded/trimmed to match `y`'s shape.
+    """
+    if x.shape[axis] < y.shape[axis]:
+        return pad_x_to_y(x, y, axis=axis)
+    else:
+        return trim_x_to_y(x, y, axis=axis)
+
+
```
mpariente, line 157: This is great, and very useful !

tests/jit/jit_models_test.py
```diff
@@ -132,7 +132,7 @@ def test_enhancement_model(small_model_params, model_def, test_data):
 def test_dcunet_model(test_shape: Tuple, matching_samples):
     n_samples = 5010
     device = get_default_device()
-    model = DCUNet(architecture="mini").eval().to(device)
+    model = DCUNet(architecture="mini", masknet_kwargs={"fix_length_mode": "pad"}).eval().to(device)
```
mpariente, line 135: This is very ugly, isn't it?..

mpariente (commented):
This is ready for merging after two things: 
- Should masknet kwargs be kept in this way? It is ugly in this way IMO. 
- Comment the test that's failing, or skip if torch is under 1.8, sounds ok? 
- Add support for `n_src` argument in both architectures.

jonashaag:
> Do you have a link to the issue or the PR?

I think it is https://github.com/pytorch/pytorch/issues/47157

> Maybe we should run tests on torch-nightly as well?

That would be a good idea.

jonashaag:
> * Comment the test that's failing, or skip if torch is under 1.8, sounds ok?

Better skip if Torch < 1.8

asteroid/utils/torch_utils.py
jonashaag: It's similar to `librosa.util.fix_length`. I wonder if we should call it something similar, but then it'd be inconsistent with the other names

tests/jit/jit_models_test.py
jonashaag: I agree, I think I did it this way because there where some argument overriding issues with the DCCRN constructor but I don't remember the details and I'm sure we can also get it to work without the extra level. Would you suggest to add the masknet kwargs to the model kwargs, ie. flat? Seems to least ugly to me

asteroid/utils/torch_utils.py
mpariente: Also don't know. 
It would be ok to change the name of `pad_x_to_y`, if we think of a better one, that unifies those things.

tests/jit/jit_models_test.py
mpariente: Yes, this will be much better IMO.

mpariente:
I let you take care of those changes (skip the test and masknet_kwargs). We'll add tests on nigthly afterwards.

asteroid/utils/torch_utils.py
jonashaag: Let's do it later maybe

jonashaag:
Ok, done! I also changed the args specifications to be all tuples because I accidentally modified some of the lists because they are mutable, which is a terrible thing to debug.

jonashaag:
I created a PR against PyTorch for better version checks https://github.com/pytorch/pytorch/pull/48414

asteroid/utils/test_utils.py
```diff
@@ -29,3 +29,8 @@ def __getitem__(self, idx):
         mixed = torch.randn(self.inp_n_sig, self.inp_len_wave)
         srcs = torch.randn(self.out_n_sig, self.out_len_wave)
         return mixed, srcs
+
+
+def torch_version_tuple():
+    version, suffix = torch.__version__.split("+")
+    return tuple(map(int, version.split("."))) + (suffix,)
```
mpariente, line 36:
```suggestion
def torch_version_tuple():
    version, *suffix = torch.__version__.split("+")
    return tuple(map(int, version.split("."))) + tuple(suffix)
```

asteroid/masknn/base.py
```diff
@@ -68,18 +68,22 @@ class BaseDCUMaskNet(BaseUNet):
...
+            kwargs (optional): Passed to ``__init__``.
         """
         encoders, decoders = cls._architectures[architecture]
+        # Fix n_src in last decoder
+        in_chan, _ignored_out_chan, *rest = decoders[-1]
+        decoders = (*decoders[:-1], (in_chan, n_src, *rest))
         return cls(encoders, decoders, **kwargs)
 
-    def __init__(self, encoders, decoders, output_layer=None, mask_bound="tanh", **kwargs):
+    def __init__(self, encoders, decoders, output_layer=None, mask_bound="tanh", n_src=1, **kwargs):
```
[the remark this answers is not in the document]
jonashaag:
```suggestion
    def __init__(self, encoders, decoders, output_layer=None, mask_bound="tanh", **kwargs):
```

asteroid/models/dccrnet.py
```diff
@@ -21,15 +22,12 @@ class DCCRNet(BaseDCUNet):
 
     masknet_class = DCCRMaskNet
 
-    def __init__(
-        self, *args, stft_kernel_size=512, sample_rate=16000.0, masknet_kwargs=None, **kwargs
```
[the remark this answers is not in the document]
jonashaag: I removed all the duplicate params everywhere and kept only the ones we need (`stft_kernel_size` in this case). We can also add it back if you think it's more clear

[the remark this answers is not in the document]
mpariente: Ok, thanks.

mpariente:
I'll review the whole PR again today but I think it's ready to be merged !

jonashaag:
> I'll review the whole PR again today

Thank you! Much appreciated. Somehow this PR got MUCH bigger than expected :-D

asteroid/masknn/recurrent.py
```diff
@@ -555,14 +571,34 @@ class DCCRMaskNet(BaseDCUMaskNet):  # CHECK-JIT
...
-        self.n_freqs = n_freqs
+
+    def _check_input_dims(self, x):
+        # TODO: We can probably lift the shape requirements once Keras-style "same"
+        # padding for convolutions has landed: https://github.com/pytorch/pytorch/pull/42190
+        freq_prod, _ = self.encoders_stride_product
+        if x.shape[1] % freq_prod:
+            raise TypeError(
+                f"Input shape must be [batch, freq, time] with freq divisible by {freq_prod}, got {x.shape} instead"
+            )
```
jonashaag, line 604:
```suggestion
```

asteroid/masknn/convolutional.py
```diff
@@ -553,7 +553,8 @@ def _fix_dcu_input_dims(fix_length_mode: Optional[str], x, encoders_stride_produ
             pad_shape = [0, time_prod - time_remainder]
             x = nn.functional.pad(x, pad_shape, mode="constant")
         elif fix_length_mode == "trim":
-            x = x[..., :-time_remainder]
+            pad_shape = [0, -time_remainder]
+            x = nn.functional.pad(x, pad_shape, mode="constant")
```
mpariente, line 557: Can you have a look at this? 
Maybe this solves the issue you were mentioning? I didn't find the failure on my end. Checking if it works with torch 1.7

asteroid/masknn/convolutional.py
```diff
@@ -491,6 +506,65 @@ class DCUMaskNet(BaseDCUMaskNet):  # CHECK-JIT
...
+        else:
+            raise ValueError(f"Unknown fix_length mode '{fix_length_mode}'")
+    return x
+
+
+@script_if_tracing
+def _fix_dcu_output_dims(fix_length_mode: Optional[str], out, x):
+    """Fix shape of `out` to the original shape of `x`."""
+    return pad_x_to_y(out, x)
```
mpariente, line 566: Checking if negative padding works here (trimming and padding in the same function) and solves the 1.7 error

jonashaag:
Found missing shape check for DCCRN, added tests for shape checks, and also fixed the PyTorch 1.8 issue (it only breaks if you do `foo[:x]` where `x = len(foo)`, ie. a noop slice)

tests/models/models_test.py
```diff
@@ -143,20 +142,35 @@ def test_dptnet(fb):
...
+    DCUNet("mini").masker(torch.zeros((1, 9, 17), dtype=torch.complex64))
+    with pytest.raises(TypeError):
+        DCUNet("mini").masker(torch.zeros((1, 42, 17), dtype=torch.complex64))
+
+    # DCUMaskNet should fail with wrong time dimensions if fix_length_mode is not used
+    DCUNet("mini", fix_length_mode="pad").masker(torch.zeros((1, 9, 17), dtype=torch.complex64))
+    DCUNet("mini", fix_length_mode="trim").masker(torch.zeros((1, 9, 17), dtype=torch.complex64))
+    with pytest.raises(TypeError):
+        DCUNet("mini").masker(torch.zeros((1, 9, 16), dtype=torch.complex64))
+
```
mpariente: Cool !

asteroid/masknn/convolutional.py
jonashaag: The issue was somewhere else, in `trim_x_to_y` if both have the same length, but it's fixed now

mpariente:
So the trimming padding can be done with a single function, `trim_x_to_y`  and `pad_or_trim_x_to_y` are not needed it seems.

mpariente:
It's quite dangerous to push to the same branch in the same time (my bad :sweat_smile: ) but we did well ^^

asteroid/masknn/convolutional.py
mpariente: I finally saw that, thanks.

jonashaag:
> So the trimming padding can be done with a single function

Indeed it seems so! https://github.com/pytorch/pytorch/blob/367426494759ddde0896665ed55c6f9af2870cf0/aten/src/ATen/native/ConstantPadNd.cpp#L25

jonashaag:
So we could remove the trim functions and change the pad docstring to say "right-pad or right-trim".

Btw I won't be pushing any code for the next few hours, so feel free to push :P

mpariente:
I'll do that.

mpariente:
I re-reviewed everything and it LGTM
Waiting for your approval to merge it.

jonashaag:
Also did a quick review, LGTM, go ahead :)

PR #283, opened 2020-10-21
[src] Automatic resampling in separate.

mpariente:
Use the sample rate property to automatically resample input wav files. 

Two questions: 
- Should the output be resampled at the original sample rate? 
- Should we use `librosa` or `torchaudio`? Librosa has more deps and is longer to load.. 

Ref #274

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -43,7 +43,7 @@ class BaseModel(nn.Module):
         raise NotImplementedError
 
     @torch.no_grad()
-    def separate(self, wav, output_dir=None, force_overwrite=False, **kwargs):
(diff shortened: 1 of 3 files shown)


jonashaag:
I think we should not do any of this. It's easy for the user to resample for model input and resample the model output. So better be explicit and require the user to type a bit more. As a compromise we could have automatic resampling behind a flag: if the flag is true, resample without warning; if the flag is false, raise an error.

mpariente:
It makes sense to do it because of the CLI IMO..

jonashaag:
This is exactly how I’d do it. If you u prefer you could even make resample default to true in the CLI, but not sure if that’s good.

PR #284, opened 2020-10-21
[src] Add sample_rate argument to all supported models.

mpariente:
Add samplerate argument to all models and pass it to the filterbank. This should work as is. 

Ref: #274 #283


jonashaag:
What does this mean for pretrained models that we already have?

mpariente:
For now, nothing. 
If there is no `sample_rate` in the pretrained model, we'll handle it in `from_pretrained`. But that'll be for a next PR.

jonashaag:
Ok, LGTM, except for the denoising models which I’d say should have 16k (DCUNet and DCCRN)

mpariente:
Changing it then.

PR #285, opened 2020-10-21
Load models without sample_rate

mpariente:
This implements the loading of known models with known sample rate, and raises error when sample rate is not known. 

Also made a CLI to register sample rate to model (for the lazy ones). Is there a better name (more explicit) than `asteroid-cli register-samplerate`?

Refs: #274 #283 #284

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -15,6 +15,6 @@ jobs:
       with:
         fetch-depth: 0
     - name: Automatic Rebase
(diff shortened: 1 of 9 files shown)


jonashaag:
This is great. LGTM!

mpariente:
\rebase

mpariente:
\rebase

mpariente:
Missing test for the CLI. Merging anyway.

PR #288, opened 2020-10-27
Rewrite forward of BaseEncoderMaskerDecoder to be more readable

mhu-coder:
The step of shaping `reconstructed` is taken out of the class and turned into a torch scriptable function.

Thing I don't really like: keeping around the non squeezed version of `wav` just for reshaping.

Errors I got when I tried to avoid this issue:
* If `_shape_reconstructed` takes a tensor and a target dimension, e.g. (tensor, wav_dim), `wav_dim` is frozen during tracing and becomes a constant in the traced model. In other words, inputs of shapes other than what was seen during tracing may have the wrong output dimension
* To force the execution of `wav_dim = wav.ndim` in the traced model, I tried to write a function that just returns the `ndim` attribute of a tensor which would have been called inside `forward`. Turns out that a function that returns an `int` cannot be scripted...

If you see a more elegant solution, feel free to modify the code!

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -6,7 +6,7 @@ import torch
 from torch import nn
 
 from ..masknn import activations
-from ..utils.torch_utils import pad_x_to_y, script_if_tracing
+from ..utils.torch_utils import pad_x_to_y, script_if_tracing, get_shape
 from ..utils.hub_utils import cached_download
 
 
@@ -385,23 +385,6 @@ class BaseEncoderMaskerDecoder(BaseModel):
             "encoder_activation": self.encoder_activation,
         }
         return model_args
-
-
-@script_if_tracing
-def get_shape(tensor):
-    """Gets shape of `tensor` as `torch.Tensor` type for jit compiler
-
-    Note:
-        Returning `tensor.shape` of `tensor.size()` directly is not torchscript
-        compatible as return type would not be supported.
-
-    Args:
-        tensor (torch.Tensor): Tensor
-
-    Returns:
-        torch.Tensor: Shape of `tensor`
-    """
-    return torch.tensor(tensor.shape)
 
 
 @script_if_tracing
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -6,7 +6,7 @@ import torch
 from torch import nn
 
 from ..masknn import activations
-from ..utils.torch_utils import pad_x_to_y, script_if_tracing, get_shape
+from ..utils.torch_utils import pad_x_to_y, script_if_tracing, jitable_shape
 from ..utils.hub_utils import cached_download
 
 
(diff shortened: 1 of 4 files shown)


mpariente (commented):
It's not that pretty, indeed.. @jonashaag, thoughts?

jonashaag:
Maybe it works if you just remember the shape of the input, not the input itself?

mhu-coder:
> Maybe it works if you just remember the shape of the input, not the input itself?

Good call! Thanks for the suggestion. Passing the shape instead seems to work too according to the tests despite the following warning
```
tests/jit/jit_models_test.py: 24 warnings
  asteroid/asteroid/models/base_models.py:299: TracerWarning: torch.tensor results are registered as constants
  in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables
  that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
    shape = torch.tensor(wav.size())
```

I tried to type hint `size` in `_shape_reconstructed` with `torch.Size` to avoid the conversion to a tensor but the type is not recognized by the jit compiler :confused:.

jonashaag:
Hmm but I'm worried about the warning. If it works despite the warning saying it doesn't, maybe it's an implementation detail subject to change in the future

jonashaag:
Maybe you can wrap the shape-getting code into a function with `@script_if_tracing`?

```py
@script_if_tracing
def get_shape(t):
    return t.shape

@script_if_tracing
def reconstruct_shape(t, shape):
    return t.reshape(*shape)
```

mpariente:
The output shape is not the input shape, ex: `(L,) -> (2, L)`, `(1, L) -> (1, 2, L)`. the only thing we want to do is to remove the batch dimension if the input didn't have one.

mpariente:
But script a `was1d` function and keeping the original code would work I guess (thanks jonas for the idea), right?

mhu-coder:
> Maybe you can wrap the shape-getting code into a function with @script_if_tracing

We can wrap the instruction to get the shape and avoid the warning indeed. It would be something like

```python
@script_if_tracing
def get_shape(tensor):
    # Cannot return shape directly with jit compiler
    return torch.tensor(tensor.shape)

@script_if_tracing
def _shape_reconstructed(reconstructed, size):
    if size.ndim == 1:
        return reconstructed.squeeze(0)
    return reconstructed

def forward(self, wav):
    shape = get_shape(wav)
    # ...
    return _shape_reconstructed(reconstructed, shape)
```

> But script a was1d function and keeping the original code would work I guess (thanks jonas for the idea), right?

If I got you right, we would have a `script_if_tracing` decorated function that takes a tensor as input and returns a boolean (`was1d` in this case)? I don't think a function that returns a boolean is torchscript compatible.

mpariente:
Yeah, I tried, my idea didn't work ^^

mpariente:
And yes, just tried what Jonas suggests, it works.

jonashaag:
> The output shape is not the input shape

Oops, yes, the example was bad.

Note that Jit also seems to support Python tuples so maybe we can also avoid the warning by simply converting the shape to a Python tuple?

asteroid/models/base_models.py
```diff
@@ -387,6 +387,23 @@ def get_model_args(self):
         return model_args
 
 
+@script_if_tracing
+def get_shape(tensor):
```
mpariente, line 391: Let's call that `jit_get_shape` and move it to `utils.torch_utils`, we might need it somewhere else and it's more discoverable there.

mpariente:
Thanks @mhu-coder !

Issue #289, opened 2020-10-29
seems like something incompatible after torchaudio 0.7.0


SouppuoS:
## 🐛 Bug

try to install asteroid on Colab, follow the installation steps in README

>ERROR: tensorflow 2.3.0 has requirement tensorboard<3,>=2.3.0, but you'll have tensorboard 2.2.0 which is incompatible.
ERROR: torchaudio 0.7.0 has requirement torch==1.7.0, but you'll have torch 1.6.0+cu101 which is incompatible.
ERROR: pytorch-lightning 0.9.0 has requirement future>=0.17.1, but you'll have future 0.16.0 which is incompatible.
ERROR: pytorch-lightning 0.9.0 has requirement PyYAML>=5.1, but you'll have pyyaml 3.13 which is incompatible.

and run a recipe
>Traceback (most recent call last):
  File "train.py", line 11, in <module>
    from asteroid import DPRNNTasNet
  File "/content/asteroid/asteroid/__init__.py", line 3, in <module>
    from .models import ConvTasNet, DCCRNet, DCUNet, DPRNNTasNet, DPTNet, LSTMTasNet, DeMask
  File "/content/asteroid/asteroid/models/__init__.py", line 2, in <module>
    from .conv_tasnet import ConvTasNet
  File "/content/asteroid/asteroid/models/conv_tasnet.py", line 1, in <module>
    from ..filterbanks import make_enc_dec
  File "/content/asteroid/asteroid/filterbanks/__init__.py", line 6, in <module>
    from .griffin_lim import griffin_lim, misi
  File "/content/asteroid/asteroid/filterbanks/griffin_lim.py", line 7, in <module>
    from ..dsp.consistency import mixture_consistency
  File "/content/asteroid/asteroid/dsp/__init__.py", line 2, in <module>
    from .overlap_add import LambdaOverlapAdd, DualPathProcessing
  File "/content/asteroid/asteroid/dsp/overlap_add.py", line 3, in <module>
    from asteroid.losses.pit_wrapper import PITReorder
  File "/content/asteroid/asteroid/losses/__init__.py", line 8, in <module>
    from .stoi import NegSTOILoss as SingleSrcNegSTOI
  File "/content/asteroid/asteroid/losses/stoi.py", line 1, in <module>
    from torch_stoi import NegSTOILoss as _NegSTOILoss
  File "/usr/local/lib/python3.6/dist-packages/torch_stoi/__init__.py", line 1, in <module>
[...]

PR #292, opened 2020-10-29
[install] Pin torchaudio 0.6.0 and torch 1.6.0 (#289)

SouppuoS:
add limitation on torch and torchaudio version


mpariente:
This is temporary, to find out what's happening. We need to enable new versions of torch asap.

PR #295, opened 2020-10-29
Revert "[install] Pin torchaudio 0.6.0 and torch 1.6.0 (#289)"

mpariente:
Reverts mpariente/asteroid#292


mpariente:
CI is passing so the import is going fine. 
I initially thought it was for matching torch/torchaudio installs.

PR #297, opened 2020-10-30
[docs] Fix Reference and Example blocks in docs

mpariente:
It annoyed me.. 

Need to add lot of new things in the docs..

### DIFF
--- a/asteroid/models/zenodo.py
+++ b/asteroid/models/zenodo.py
@@ -78,12 +78,12 @@ class Zenodo(object):
 
         Examples
             >>> metadata = {
-            >>>     'title': 'My first upload',
-            >>>     'upload_type': 'poster',
-            >>>     'description': 'This is my first upload',
-            >>>     'creators': [{'name': 'Doe, John',
-            >>>                   'affiliation': 'Zenodo'}]
-            >>> }
+            ...     'title': 'My first upload',
+            ...     'upload_type': 'poster',
+            ...     'description': 'This is my first upload',
+            ...     'creators': [{'name': 'Doe, John',
+            ...                   'affiliation': 'Zenodo'}]
+            ... }
         """
         data = {"metadata": metadata}
         r = requests.put(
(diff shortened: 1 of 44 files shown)


asteroid/models/zenodo.py
```diff
@@ -76,14 +76,14 @@ def change_metadata_in_deposition(self, dep_id, metadata):
...
-            metadata = {
-                'title': 'My first upload',
-                'upload_type': 'poster',
-                'description': 'This is my first upload',
-                'creators': [{'name': 'Doe, John',
-                              'affiliation': 'Zenodo'}]
-            }
+        Examples
+            >>> metadata = {
+            >>>     'title': 'My first upload',
```
jonashaag, line 81: all prefixes except the first should be `...`

mpariente: Because it's one instruction?

mpariente:
The docs build doesn't work because of git's deps problems, do you know this problem? Have a solution? @jonashaag

asteroid/models/zenodo.py
```diff
@@ -76,14 +76,14 @@ def change_metadata_in_deposition(self, dep_id, metadata):
...
+            >>>     'upload_type': 'poster',
+            >>>     'description': 'This is my first upload',
+            >>>     'creators': [{'name': 'Doe, John',
+            >>>                   'affiliation': 'Zenodo'}]
+            >>> }
```
mpariente, line 86: Like this I guess. 
```suggestion
            >>> metadata = {
            ...     'title': 'My first upload',
            ...     'upload_type': 'poster',
            ...     'description': 'This is my first upload',
            ...     'creators': [{'name': 'Doe, John',
            ...                   'affiliation': 'Zenodo'}]
            ... }
```

asteroid/models/zenodo.py
jonashaag: Yep! I assume you wanted it to look like the Python repl.

PR #299, opened 2020-10-31
Add more shape checks to SDR and MSE losses

jonashaag:
We can also add this to more losses.

### DIFF
--- a/asteroid/losses/mse.py
+++ b/asteroid/losses/mse.py
@@ -25,6 +25,10 @@ class PairwiseMSE(_Loss):
     """
 
     def forward(self, est_targets, targets):
+        if targets.size() != est_targets.size() or targets.ndim < 3:
+            raise TypeError(
+                f"Inputs must be of shape [batch, n_src, *], got {targets.size()} and {est_targets.size()} instead"
+            )
         targets = targets.unsqueeze(1)
         est_targets = est_targets.unsqueeze(2)
         pw_loss = (targets - est_targets) ** 2
@@ -58,6 +62,10 @@ class SingleSrcMSE(_Loss):
     """
 
     def forward(self, est_targets, targets):
+        if targets.size() != est_targets.size() or targets.ndim < 2:
+            raise TypeError(
+                f"Inputs must be of shape [batch, *], got {targets.size()} and {est_targets.size()} instead"
+            )
         loss = (targets - est_targets) ** 2
         mean_over = list(range(1, loss.ndim))
         return loss.mean(dim=mean_over)
--- a/tests/losses/loss_functions_test.py
+++ b/tests/losses/loss_functions_test.py
@@ -5,33 +5,67 @@ import warnings
 
 from asteroid.filterbanks import STFTFB, Encoder, transforms
 from asteroid.losses import PITLossWrapper
-from asteroid.losses import sdr
-from asteroid.losses import singlesrc_mse, pairwise_mse, multisrc_mse
+from asteroid.losses import sdr, mse
 from asteroid.losses import deep_clustering_loss, SingleSrcPMSQE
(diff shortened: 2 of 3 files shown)


asteroid/losses/mse.py
```diff
@@ -25,6 +25,10 @@ class PairwiseMSE(_Loss):
     """
 
     def forward(self, est_targets, targets):
+        if targets.size() != est_targets.size() or targets.ndim != 3:
+            raise TypeError(
+                f"Inputs must be of shape [batch, n_src, *], got {targets.size()} and {est_targets.size()} instead"
+            )
```
mpariente, line 31: The first part of the check is fine but the second one isn't. 
Those forward are made to work for any input shape. 
Sorry if you lost time on the checks and the tests.

jonashaag: Ah, right. But for *SDR we always have `2 <= ndim <= 3`?

jonashaag: Or shall accept any shape for `time`, i.e. `(batch, n_src, time1, time2, ...)`?

mpariente: In *SDR, the checks are correct. We can remove them in the MSE.

tests/losses/loss_functions_test.py
```diff
@@ -5,33 +5,67 @@
...
 
 
+def assert_loss_checks_shape(loss_func, shape, arbitrary_last_dim=False, no_batch_ok=False):
+    """Test that `loss_func` raises a TypeError if you are passing anything that isn't of the expected shape.
+
+    Args:
+        loss_func (callable): The loss to check, signature: `loss_func(x, y) -> Any`
+        shape (tuple): Shape that the loss is expected to accept (without batch dimension).
+        arbitrary_last_dim (bool, optional): Whether the last dimension may be replaced by any number of dimensions.
+        no_batch_ok (bool, optional): Whether no batch dimension is acceptable.
```
jonashaag, line 21: (Testing suggestions)

```suggestion
        no_batch_ok (bool, optional): Whether having no batch dimension is acceptable.
```

jonashaag:
Good for another review :)

tests/losses/loss_functions_test.py
```diff
@@ -48,6 +82,17 @@ def test_sisdr(n_src, function_triplet):
...
 
+@pytest.mark.parametrize("loss", loss_properties)
+def test_sisdr_and_mse_shape_checks(loss):
+    pairwise, singlesrc, multisrc, arbitrary_last_dim = loss
+    assert_loss_checks_shape(pairwise, (3, 1000), arbitrary_last_dim)
+    assert_loss_checks_shape(singlesrc, (1000,), arbitrary_last_dim)
+    # Special case for multisrc_mse that shares the same code with
+    # singlesrc_mse and thus accepts 1-dim tensors.
+    no_batch_ok = multisrc == mse.multisrc_mse
+    assert_loss_checks_shape(multisrc, (3, 1000), arbitrary_last_dim, no_batch_ok)
```
mpariente, line 93: I don't see the purpose of the `no_batch_ok`.
Without batch, the `multisrc_mse` will return a tensor of size `n_src`. It's not really what we want.

Is there something I'm missing?

jonashaag: If we don't add the special case, the checks will fail: It will assert that `multisrc_mse` fails with `(3, 1000)` shaped input. But since `MultiSrcMSE = SingleSrcMSE`, this is identical to testing `singlesrc_mse` with `(1000,)` input

mpariente: I see now, thanks. 
I'm not convinced by how useful it is to check that it doesn't work without batch, but it's fine.

mpariente (approved):

PR #302, opened 2020-11-02
Add Sinkhorn PIT loss

tachi-hi:
### asteroid/engine/system.py 
- A change was made to pass the epoch number to the PIT wrapper.

### asteroid/losses/pit_wrapper.py
- Add Sinkhorn iteration-based permutation search.
    - Some parameters are still hard coded.
- Possibility of stopping on ASSERT during the validation, which is painful if it occurs in a very heavy training. We may need to devise a way to separate valid and test by flags. Replacing it with the Hungarian algorithm is another option.

### tests
- Check if Sinkhorn and Hungarian return almost the same results for 2 <= N <=9.
- Shape check.
- Check if Sinkhorn-based reordering is correct.

### DIFF
--- a/asteroid/engine/system.py
+++ b/asteroid/engine/system.py
@@ -6,8 +6,6 @@ from torch.optim.optimizer import Optimizer
 
 from ..utils import flatten_dict
 
-from asteroid.losses import PITLossWrapper
-
 
 class System(pl.LightningModule):
     """Base class for deep learning systems.
@@ -93,10 +91,7 @@ class System(pl.LightningModule):
         """
         inputs, targets = batch
         est_targets = self(inputs)
-        if isinstance(self.loss_func, PITLossWrapper):
-            loss = self.loss_func(est_targets, targets, epoch=self.current_epoch)
-        else:
-            loss = self.loss_func(est_targets, targets)
+        loss = self.loss_func(est_targets, targets)
         return loss
 
     def training_step(self, batch, batch_nb):
--- a/asteroid/engine/system.py
+++ b/asteroid/engine/system.py
@@ -6,6 +6,8 @@ from torch.optim.optimizer import Optimizer
 
 from ..utils import flatten_dict
 
+from asteroid.losses import PITLossWrapper
+
 
 class System(pl.LightningModule):
     """Base class for deep learning systems.
@@ -91,7 +93,10 @@ class System(pl.LightningModule):
         """
         inputs, targets = batch
         est_targets = self(inputs)
-        loss = self.loss_func(est_targets, targets)
+        if isinstance(self.loss_func, PITLossWrapper):
+            loss = self.loss_func(est_targets, targets, epoch=self.current_epoch)
+        else:
+            loss = self.loss_func(est_targets, targets)
         return loss
 
     def training_step(self, batch, batch_nb):
--- a/asteroid/losses/sinkpit_wrapper.py
+++ b/asteroid/losses/sinkpit_wrapper.py
@@ -1,0 +1,185 @@
+from itertools import permutations
+import torch
+from torch import nn
+from torch.testing import assert_allclose
+
+import pytorch_lightning as pl
+
+from asteroid.losses import PITLossWrapper
+
+
+class SinkPITLossWrapper(nn.Module):
+    r"""Permutation invariant loss wrapper.
+
+    Args:
+        loss_func: function with signature (targets, est_targets, **kwargs).
+        n_iter (int): number of the Sinkhorn iteration (default = 200).
+            Supposed to be an even number.
+
+        `loss_func` computes pairwise
+        losses and returns a torch.Tensor of shape
+        :math:`(batch, n\_src, n\_src)`. Each element
+        :math:`[batch, i, j]` corresponds to the loss between
+        :math:`targets[:, i]` and :math:`est\_targets[:, j]`
+        It evaluates an approximate value of the PIT loss
+        using Sinkhorn's iterative algorithm.
+        See :meth:`~PITLossWrapper.best_softperm_sinkhorn`
+        and http://arxiv.org/abs/2010.11871
+
+    For each of these modes, the best permutation and reordering will be
+    automatically computed.
+
+    Example 1
+        >>> import torch
+        >>> from asteroid.losses import pairwise_neg_sisdr
+        >>> sources = torch.randn(10, 3, 16000)
+        >>> est_sources = torch.randn(10, 3, 16000)
+        >>> # Compute SinkPIT loss based on pairwise losses
+        >>> loss_func = SinkPITLossWrapper(pairwise_neg_sisdr)
+        >>> loss_val = loss_func(est_sources, sources)
+
+    A fixed temperature parameter `beta` (=10) is used unless a cooling callback is set.
+    A cooling callbacks regulates the value of `beta`.
+
+    Example 2
+        TODO
+    """
+
+    def __init__(self, loss_func, n_iter=200, *args, **kwargs):
+        super().__init__()
+        self.loss_func = loss_func
+        self._beta = 10
+        self.n_iter = n_iter
+
+    @property
+    def beta(self):
+        return self._beta
+
+    @beta.setter
+    def beta(self, beta):
+        assert beta > 0
+        self._beta = beta
+
+    def forward(self, est_targets, targets, return_est=False, *args, **kwargs):
+        """Evaluate the loss using Sinkhorn's algorithm.
+        Args:
+            est_targets: torch.Tensor. Expected shape [batch, nsrc, *].
+                The batch of target estimates.
+            targets: torch.Tensor. Expected shape [batch, nsrc, *].
+                The batch of training targets
+            return_est: Boolean. Whether to return the reordered targets
+                estimates (To compute metrics or to save example).
+            **kwargs: additional keyword argument that will be passed to the
+                loss function.
+
+        Returns:
+            - Best permutation loss for each batch sample, average over
+                the batch. torch.Tensor(loss_value)
+            - The reordered targets estimates if return_est is True.
+                torch.Tensor of shape [batch, nsrc, *].
+        """
+        n_src = targets.shape[1]
+        assert n_src < 100, f"Expected source axis along dim 1, found {n_src}"
+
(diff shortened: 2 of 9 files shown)


mpariente:
Thanks a lot for the implementation, it's great !! 
Is this ready for review in your opinion? Should I give you feedback?

tachi-hi:
I'd be grateful for all kinds of feedback!

The test was successful in my local environment, but the CI here failed.
(Failure in quantizing a soft permutation matrix into a hard permutation matrix.)
It is probably due to a difference in the seeds of random numbers.

In light of the CI results, it may be better to use the strict Hungarian algorithm than 
to quantize a soft permutation matrix, during the valid/test(`reorder=True`) stage. 
(There is also an option to leave both and select one of them with a flag, 
but it may be more complicated than necessary).

asteroid/engine/system.py
```diff
@@ -91,7 +93,10 @@ def common_step(self, batch, batch_nb, train=True):
         """
         inputs, targets = batch
         est_targets = self(inputs)
-        loss = self.loss_func(est_targets, targets)
+        if isinstance(self.loss_func, PITLossWrapper):
+            loss = self.loss_func(est_targets, targets, epoch=self.current_epoch)
+        else:
+            loss = self.loss_func(est_targets, targets)
```
mpariente, line 99: I don't think we'll accept that. 
Otherwise, we'll never stop growing the codebase for specific use cases.

mpariente (commented):
Overall, it looks good, thanks ! 

I think we should rather implement `SinkPITLossWrapper` instead of adding yet another mode to `PITLossWrapper`. 
SinkPIT is actually quite different because it retains more informations than the other methods, and it would make sense to make it a different object altogether. Could you implement it as a separate class please? That would be great !

Also, there would be less constraint on it, and it would make experimentation easier. I'll have a look at the tests when it's a separate class.

asteroid/engine/system.py
tachi-hi: OK!
In the modified version (in another branch), I've tried using a callback class that updates the `beta` at the beginning of each epoch.

mpariente: Cool !
Feel free to open a PR with the other branch :smiley:

mpariente:
I see a TODO in the docstring, but otherwise this seems ready. Should it be reviewed?

tachi-hi:
I think I'm ready to ask for a code review.
- Implemented a `SinkPITLossWrapper` class
- Epoch number is passed to `SinkPITLossWrapper` through an inheritance of `pl.callbacks.Callback`.
- I filled in the TODO in the docstring with a snippet.
- Added an option whether to use the Hungarian algorithm for the validation.
- I did an experiment replacing the PIT in an `egs` sample with SinkPIT. It seems to be working (the loss is decreasing). But I haven't written a test for this part.

asteroid/losses/sinkpit_wrapper.py
```diff
@@ -0,0 +1,211 @@
...
+        >>>
+        >>> trainer = pl.Trainer(
+        >>>     max_epochs=100,
+        >>>     callbacks=[SinkPITBetaScheduler(lambda epoch : 1.02 ** epoch)],
+        >>>     )
+        >>>
+        >>> trainer.fit(system)
+    """
+
+    def __init__(self, loss_func, n_iter=200, hungarian_validation=True, *args, **kwargs):
```
mpariente, line 64: `args` and `kwargs` are unused, why are they here?

mpariente:
Did you untick `Allow edits from maintainers`? 
I tried to commit to the PR but couldn't.

mpariente (commented):
This is mostly ready for merging, thanks. 

I wonder, would it make sense to make `PITBase` class with all the static methods, and inherit both wrappers from it? It would be more pythonic IMO. @jonashaag?

After we decide about, I'd like to change a few details and this is ready. 
The docstrings are really pretty, thanks for the extra mile !

mpariente:
> Did you untick `Allow edits from maintainers`?
> I tried to commit to the PR but couldn't.

My bad, I made a silly mistake. I could push now

asteroid/losses/sinkpit_wrapper.py
```diff
@@ -0,0 +1,192 @@
...
+        :math:`(batch, n\_src, n\_src)`. Each element
+        :math:`[batch, i, j]` corresponds to the loss between
+        :math:`targets[:, i]` and :math:`est\_targets[:, j]`
+        It evaluates an approximate value of the PIT loss
+        using Sinkhorn's iterative algorithm.
+        See :meth:`~PITLossWrapper.best_softperm_sinkhorn`
+        and http://arxiv.org/abs/2010.11871
+
+    Examples
+        >>> import torch
```
jonashaag, line 29: This is not something that we have to fix in this PR, but still: Since none of these lines actually show the return values of any of the expressions, I think we can do away with all of the `>>>`. (Btw they aren't "correct" since in the multiline expressions below the lines should start with `...`.) What do you think @mpariente?

tests/losses/sinkpit_wrapper_test.py
```diff
@@ -0,0 +1,150 @@
...
+    loss(est_targets, targets)
+    loss_value, reordered_est = loss(est_targets, targets, return_est=True)
+    assert reordered_est.shape == est_targets.shape
+
+
+@pytest.mark.parametrize("batch_size", [1, 2])
+@pytest.mark.parametrize("n_src", [2, 3, 4])
+@pytest.mark.parametrize("beta,n_iter", [(100.0, 2000)])
+@pytest.mark.parametrize(
+    "function_triplet",
```
jonashaag, line 60: Maybe import this from the other loss tests instead of redefining it here?

tests/losses/sinkpit_wrapper_test.py
```diff
@@ -0,0 +1,150 @@
...
+    mean_loss_sinkhorn = loss_hungarian(est_targets, targets, return_est=False)
+
+    # compute loss by hungarian
+    mean_loss_hungarian = loss_hungarian(est_targets, targets, return_est=False)
+
+    # compare
+    assert_allclose(mean_loss_sinkhorn, mean_loss_hungarian)
+
+
+class testCallback(pl.callbacks.Callback):
```
jonashaag, line 95: This should have an uppercase T: `TestCallback`. I wonder why the linters don't complain about it.

jonashaag:
> I wonder, would it make sense to make `PITBase` class with all the static methods, and inherit both wrappers from it? It would be more pythonic IMO. @jonashaag?

What's the reason for having these in the class in the first place? Convenience? Or so that subclasses can override them?

mpariente:
Convenience.
I use the static method from `PITLossWrapper` in the end.

tests/losses/sinkpit_wrapper_test.py
mpariente: Also don't know. 
Useless linter ^^

tests/losses/sinkpit_wrapper_test.py
mpariente: How to import locally in tests?

tests/losses/sinkpit_wrapper_test.py
```diff
@@ -0,0 +1,150 @@
...
+        targets[:, torch.randperm(n_src), :] + noise
+    )  # reorder channels, and add small noise
+
+    # initialize wrappers
+    loss_sinkhorn = SinkPITLossWrapper(pairwise, n_iter=n_iter)
+    loss_hungarian = PITLossWrapper(pairwise, pit_from="pw_mtx")
+
+    # compute loss by sinkhorn
+    loss_sinkhorn.beta = beta
+    mean_loss_sinkhorn = loss_hungarian(est_targets, targets, return_est=False)
```
mpariente, line 86: That's not a test, it's `loss_hungarian`.. On it.

asteroid/losses/sinkpit_wrapper.py
mpariente: You mean remove them right? 
I don't know how that renders in the docs without `>>>`, I'd have to check.

tests/losses/sinkpit_wrapper_test.py
jonashaag: Well you can just do `from .loss_functions_test import ...` but it only works if you properly package all the tests (`__init__.py` files etc). I'll resolve this as "not worth the effort" :) A little duplication is okay

asteroid/losses/sinkpit_wrapper.py
jonashaag: This is Sphinx, right?

You can either do

```
Examples::

	from ... import ...
```

which renders as `Examples: ...`

or

```
Examples ::

	from ... import ...
```

which renders as `Examples ...` (no `:`)

tests/losses/sinkpit_wrapper_test.py
mpariente: Agree. Plus, this file is not supposed to test all the loss functions (it doesn't hurt though).

tests/losses/sinkpit_wrapper_test.py
```diff
@@ -65,7 +65,7 @@ def test_wrapper(batch_size, n_src, time):
         [pairwise_mse, singlesrc_mse, multisrc_mse],
     ],
 )
-def test_proximity_sinkhorn_hungrian(batch_size, n_src, beta, n_iter, function_triplet):
```
jonashaag: grafik

HUNGRIAN

asteroid/losses/sinkpit_wrapper.py
mpariente: I don't know sphinx that well, it's always a mistery why some things work and others don't.

mpariente: If you have a pointer to a succinct and good guide to Sphinx, I'll take it. 
I have a unfinished document called `writing_good_docstrings`, I still don't know how to :roll_eyes:

jonashaag: Haha yeah I was thinking about that too. I also don't have a clue. Maybe sometime we can spend some time researching this and coming up with a guide. Sometime...

jonashaag: Maybe just look at PyTorch or NumPy docstrings

Issue #294, opened 2020-10-29
[CI] Add version compatibility tests for pretrained models


mpariente:
## 🚀 Feature

Include version compat tests for pretrained models (models from asteroid 0.3.0 still work as expected in 0.3.3).

### Motivation

We can break things between versions and still pass CI, this is not ok. 
For example #255 (solved by #258 )

### Questions & remarks
- Only testing against the previous released version seems enough to me.
- This means downloading all the models from Zenodo, it seems a bit heavy. It should probably be in a separate CI build. 
- Probably need to install old version, load the models, separate one file and write to disk. Install current version. Make sure the model in the current version obtains the same output. 
- Other possibility: have a tarball with reference/estimate file pairs for all models, on Zenodo. Download this tarball, test against model prediction in the current version. One _big_ downside of that is that we have to add reference/estimate for each model _by hand_ and update this tarball for each new model.

jonashaag:
Good idea. I actually prefere the reference file solution because it is the most flexible (you decide when to update the reference) and is probably much much easier from an environment/dependency point of view... I don't fancy maintaining installation processes of ancient versions of Asteroid and its dependencies :)

Maybe we can simply add a reference to each model export automatically? So the exported model ships an example along with the weights. (Maybe even add a README etc...) Or we can have the references in a separate folder/repo.

How Huggingface is doing it: https://github.com/huggingface/transformers/blob/master/tests/test_modeling_gpt2.py#L480 Looks like they do it using a reference shipped with the code base, so not tied to a pretrained model per se. I wonder how they do it for user-uploaded models though...

mpariente:
> Maybe we can simply add a reference to each model export automatically? So the exported model ships an example along with the weights. (Maybe even add a README etc...) Or we can have the references in a separate folder/repo in the Asteroid codebase.

I thought about shipping examples with model sharing. We can't upload examples that come from wsj0 though. So random examples probably. 
All the Zenodo records can have a v2 and we could add a mix/estimates folder.  That should weight too much. 

> 
> How Huggingface is doing it: https://github.com/huggingface/transformers/blob/master/tests/test_modeling_gpt2.py#L480 Looks like they do it using a reference shipped with the code base, so not tied to a pretrained model per se. I wonder how they do it for user-uploaded models though...

It was on my todo-list to check how they do. They can afford writing input/output by hand, but for us, it's not feasible.

jonashaag:
> That should weight too much

Typo?

[...]

PR #304, opened 2020-11-02
Refactor demask

### LINKED ISSUE #293: Refactoring of `DeMask` model
### Goal
Hey guys, I was planning to refactor `DeMask` to use `BaseEncoderMaskerDecoder` so that it supports 1D input data when traced with 2D (and the other way round). Currently, a traced `DeMask` model will only work with data that has the same input shape as the example used for tracing because of the way `forward` is written.

Refactoring the class with `BaseEncoderMaskerDecoder` would automatically solve this issue in addition to cleaning the class. However I had some questions before making a whole bunch of changes (see below).

#### in `__init__`

1. When computing `n_feats_input` with `fb_type` set to `stft`, isn't the value 
```python
n_feats_input = (self.encoder.filterbank.n_filters) // 2 + 1
```
only valid for even values of `n_filters`. If it is, wouldn't we need to either support the case of odd values for `n_filters` or raise an error when `n_filters` is odd?
2. When computing `n_feats_output`, there is a check on `self.input_type`. Shouldn't it be on `self.output_type` instead? The error message also mentions "Input type" but shouldn't it be about "Output type"?

#### In `forward`

The current flow of data in the `forward` of `BaseEncoderMaskerDecoder` is as follow (written as plain text as Mermaid diagrams are not supported on GitHub yet):

```
wav -> <input shaping> -> [Encoder] -> [Encoder postprocess] -> [Encoder activation] -> tf_rep

tf_rep ->  [Masker] -> masker_outmasker_out --> [Masker postprocess] -> <mult> -> masked 
       |                                                                  1
       |------------------------------------------------------------------|
masked -> [Decoder] -> [Decoder postprocess] -> <shape output> -> reconstruction
```
where the values in square brackets indicate blocks that can be changed by any inheriting class and the values in `<>` indicate operations that are hard coded in the forward of `BaseEncoderMaskerDecoder`.

As the forward of `DeMask` is written, the input to the mask estimator is not always `tf_rep`. To make it compatible with the `BaseEncoderMaskerDecoder`, we could add a `preprocess_masker_input` between `tf_rep` and `[Masker]` in the base class. Like all the hooks in `BaseEncoderMaskerDecoder`, this would have a default behaviour of returning the data given as input.

mhu-coder:
Fixes #293 

Quick summary of the changes:

- [ | ] Backward compatibility tests for `DeMask`. The tests are commented at the moment while waiting for a solution to #294 
- [x] Clean up of `__init__`
- [x] Fix masker I/O dimensions when either `input_type` or `output_type` are set to `reim`
- [x] Changes to `BaseEncoderMaskerDecoder.forward` to cover what is going on inside `DeMask`:
  - [x] Add `preprocess_masker_input` hook
  - [x] Add `preprocess_product_input` hook, which processes `tf_rep` before application of `masks`
  - [x] Fix `_shape_reconstructed`: we should have tested the length of `size` instead of checking its `ndim` attribute as `ndim` of a `tensor.shape` is always 1
- [x] Make `DeMask` inherit from `BaseEncoderMaskerDecoder`
  - [x] Remove `forward`
  - [x] Rewrite `__init__`
  - [x] Override `preprocess_masker_input`, `preprocess_product_input` and `postprocess_masks`
- [   ] Check that recipe still runs
- [x] Raise error for `STFT` filter bank with odd number of filters.

:hankey: `preprocess_product_input` does not seem right, but I did not find a nicer way of making `DeMask` compatible with the forward of `BaseEncoderMaskerDecoder`. Any suggestion welcome :)!
:hankey: `DeMask` takes a `fb_type` argument while all the other model classes (and `make_enc_dec`) use a `fb_name` argument. Changing it could break backward compatibility though :/

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -399,7 +399,7 @@ def _shape_reconstructed(reconstructed, size):
         torch.Tensor: Reshaped waveform
 
     """
-    if size.ndim == 1:
+    if len(size) == 1:
         return reconstructed.squeeze(0)
     return reconstructed
 
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -305,10 +305,12 @@ class BaseEncoderMaskerDecoder(BaseModel):
         tf_rep = self.postprocess_encoded(tf_rep)
         tf_rep = self.enc_activation(tf_rep)
 
-        est_masks = self.masker(tf_rep)
+        masker_input = self.preprocess_masker_input(tf_rep)
+        est_masks = self.masker(masker_input)
         est_masks = self.postprocess_masks(est_masks)
 
-        masked_tf_rep = est_masks * tf_rep.unsqueeze(1)
+        tf_rep = self.preprocess_product_input(tf_rep)
+        masked_tf_rep = est_masks * tf_rep
         masked_tf_rep = self.postprocess_masked(masked_tf_rep)
 
         decoded = self.decoder(masked_tf_rep)
@@ -330,6 +332,21 @@ class BaseEncoderMaskerDecoder(BaseModel):
         """
         return tf_rep
 
+    def preprocess_masker_input(self, tf_rep):
+        """Transforms time-frequency representation for mask estimation.
+
+        Hook to perform transformations on the encoded, time-frequency domain
+        representation (output of the encoder) after encoder activation. The
+        transformed data are passed as input to the masker
+
+        Args:
+            tf_rep (torch.Tensor): Output of encoder activation
+
+        Return:
+            torch.Tensor: Data to be given to masker
+        """
+        return tf_rep
+
     def postprocess_masks(self, masks):
         """Hook to perform transformations on the masks (output of the masker) before
         masks are applied.
@@ -342,6 +359,21 @@ class BaseEncoderMaskerDecoder(BaseModel):
             Transformed `masks`
         """
         return masks
+
+    def preprocess_product_input(self, tf_rep):
+        """Transforms time-frequency representation for mask application.
+
+        Hook to perform transformation on the encoded, time-frequency domain
+        representation (output of the encoder) after encoder activation. The
+        estimated masks are applied to the transformed data.
+
+        Args:
+            tf_rep (torch.Tensor):  Output of encoder activation
+
+        Return:
+            torch.Tensor: Data that will be masked
+        """
+        return tf_rep.unsqueeze(1)
 
     def postprocess_masked(self, masked_tf_rep):
         """Hook to perform transformations on the masked time-frequency domain
--- a/asteroid/models/demask.py
+++ b/asteroid/models/demask.py
@@ -59,8 +59,8 @@ class DeMask(BaseEncoderMaskerDecoder):
             **fb_kwargs,
         )
 
-        n_masker_in = self._get_n_feats_input(input_type, encoder)
-        n_masker_out = self._get_n_feats_output(output_type, encoder)
+        n_masker_in = self._get_n_feats_input(input_type, encoder.n_feats_out)
+        n_masker_out = self._get_n_feats_output(output_type, encoder.n_feats_out)
         net = _build_masker_nn(
             n_masker_in,
             n_masker_out,
@@ -87,23 +87,23 @@ class DeMask(BaseEncoderMaskerDecoder):
         self.fb_kwargs = fb_kwargs
         self._sample_rate = sample_rate
 
-    def _get_n_feats_input(self, input_type, encoder):
+    def _get_n_feats_input(self, input_type, encoder_n_out):
         if input_type == "reim":
-            return encoder.n_feats_out
+            return encoder_n_out
 
         if input_type not in {"mag", "cat"}:
             raise NotImplementedError("Input type should be either mag, reim or cat")
 
-        n_feats_input = encoder.n_feats_out // 2
+        n_feats_input = encoder_n_out // 2
         if input_type == "cat":
-            n_feats_input += encoder.n_feats_out
+            n_feats_input += encoder_n_out
         return n_feats_input
 
-    def _get_n_feats_output(self, output_type, encoder):
+    def _get_n_feats_output(self, output_type, encoder_n_out):
         if output_type == "mag":
-            return encoder.n_feats_out // 2
+            return encoder_n_out // 2
         if output_type == "reim":
-            return encoder.n_feats_out
+            return encoder_n_out
         raise NotImplementedError("Output type should be either mag or reim")
 
     def forward_masker(self, tf_rep):
--- a/asteroid/models/demask.py
+++ b/asteroid/models/demask.py
@@ -61,7 +61,7 @@ class DeMask(BaseEncoderMaskerDecoder):
 
         n_masker_in = self._get_n_feats_input(input_type, encoder.n_feats_out)
         n_masker_out = self._get_n_feats_output(output_type, encoder.n_feats_out)
-        net = _build_masker_nn(
+        masker = _build_masker_nn(
             n_masker_in,
             n_masker_out,
             norm_type=norm_type,
@@ -70,7 +70,6 @@ class DeMask(BaseEncoderMaskerDecoder):
             dropout=dropout,
             mask_act=mask_act,
         )
-        masker = nn.Sequential(*net)
         super().__init__(encoder, masker, decoder)
 
         self.input_type = input_type
@@ -187,4 +186,4 @@ def _build_masker_nn(
         in_chan = hidden_dim
 
     net.extend([nn.Conv1d(in_chan, n_out, 1), activations.get(mask_act)()])
-    return net
+    return nn.Sequential(*net)
--- a/asteroid/models/demask.py
+++ b/asteroid/models/demask.py
@@ -125,10 +125,6 @@ class DeMask(BaseEncoderMaskerDecoder):  # CHECK-JIT
         # No need for invalid output_types as checked at init
         return est_masks
 
-    @property
-    def sample_rate(self):
-        return self._sample_rate
-
     def get_model_args(self):
         """ Arguments needed to re-instantiate the model. """
         model_args = {
--- a/asteroid/models/demask.py
+++ b/asteroid/models/demask.py
@@ -137,7 +137,7 @@ class DeMask(BaseEncoderMaskerDecoder):
         Returns:
             torch.Tensor: Masked time-frequency representations.
         """
-        if self.output_type != "mag":
+        if self.output_type == "reim":
             tf_rep = tf_rep.unsqueeze(1)
         return est_masks * tf_rep
 
--- a/asteroid/models/demask.py
+++ b/asteroid/models/demask.py
@@ -82,14 +82,14 @@ class DeMask(BaseModel):  # CHECK-JIT
 
     def _get_n_feats_input(self):
         if self.input_type == "reim":
-            return self.encoder.filterbank.n_filters
+            return self.encoder.n_feats_out
 
         if self.input_type not in {"mag", "cat"}:
             raise NotImplementedError("Input type should be either mag, reim or cat")
 
         n_feats_input = self.encoder.n_feats_out // 2
         if self.input_type == "cat":
-            n_feats_input += self.encoder.filterbank.n_filters
+            n_feats_input += self.encoder.n_feats_out
         return n_feats_input
 
     def _get_n_feats_output(self):
--- a/asteroid/models/demask.py
+++ b/asteroid/models/demask.py
@@ -1,12 +1,12 @@
 from torch import nn
-from .base_models import BaseModel
+from .base_models import BaseEncoderMaskerDecoder
 from ..filterbanks import make_enc_dec
 from ..filterbanks.transforms import take_mag, take_cat
 from ..masknn import norms, activations
 from ..utils.torch_utils import pad_x_to_y
 
 
-class DeMask(BaseModel):  # CHECK-JIT
+class DeMask(BaseEncoderMaskerDecoder):  # CHECK-JIT
     """
     Simple MLP model for surgical mask speech enhancement A transformed-domain masking approach is used.
     Args:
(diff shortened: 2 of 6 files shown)


mpariente:
Thanks a lot for the PR !

I'm going to leave the first review round to @JorisCos on this PR.
Just answering this point here. 

>  `DeMask` takes a `fb_type` argument while all the other model classes (and `make_enc_dec`) use a `fb_name` argument. Changing it could break backward compatibility though :/

I know... It was merged in a hurry, never again ! :wink: 
We're going to minorly break BC in 0.4.0. We'll also update the Zenodo records for 0.4.0, so we can change to `fb_name` at that point and the pretrained model will be loadable on both 0.3.4 and 0.4.0, I think it's reasonnable.

asteroid/models/base_models.py
```diff
@@ -305,10 +305,12 @@ def forward(self, wav):
         tf_rep = self.postprocess_encoded(tf_rep)
         tf_rep = self.enc_activation(tf_rep)
 
-        est_masks = self.masker(tf_rep)
```
jonashaag, line 308: To me all of this looks like a special case that ought not to handle in this generic forward() method. We don’t have to make forward() fit to every possible use case, otherwise it will become so generic that the abstraction hasn’t any meaning anymore. 

How about simply moving this entire mask application code into a method `apply_mask(tf_rep, est_masks) -> masked_tf_rep`? Or we can simply not use the parent forward() method at all.

mpariente: I was thinking about `apply_mask` too. This is better. 

I think `preprocess_masker_input` makes sense. If the filterbank is the STFT and you want some phase features, you don't want to mask them. It's pretty common IMO. Alternatively, we could have a `forward_masker` or `apply_masker` function instead of applying the masker directly, and this could contain additional feature extraction in subclasses.

jonashaag: Maybe it would have been better to not implement these hooks like
that in the first place and only split into 3 methods `forward_{endoder,masker,decoder}` from the beginning... not sure

mhu-coder: > Or we can simply not use the parent forward() method at all.

The incentive to stick with the `forward` was to not copy paste the block of code that shapes the input/output.

> split into 3 methods `forward_{endoder,masker,decoder}`

I agree. Having a `forward_{encoder, masker, decoder}` offers more flexibility to the inheriting classes. In addition to these 3 blocks, I would add an `apply_mask` method to keep the mask application step separate from the previous blocks. Along these lines, the forward of `BaseEncoderMaskerDecoder` would look like:

```python
def forward(self, wav):
   shape = jitable_shape(wav)
   wav = _unsqueeze_to_3d(wav)

   tf_rep = self.forward_encoder(wav)
   est_masks = self.forward_masker(tf_rep)
   masked_tf_rep = self.apply_mask(tf_rep, est_masks)
   decode = self.forward_decoder(masked_tf_rep)

   reconstructed = pad_x_to_y(decoded, wav)
   return _shape_reconstructed(reconstructed, shape)

def forward_encoder(self, wav):
    # Computes time-frequency representation of `wav`
    # default: applies encoder and encoder activation
    # ...
    return tf_rep

def forward_masker(self, tf_rep):
    # Estimate masks based on `tf_rep`.
    # default: applies masker
    # ...
    return masks

def apply_masks(self, tf_rep, masks):
    # Applies `masks` on `tf_rep`
    return masks * tf_rep.unsqueeze(1)

def forward_decoder(self, masked):
    # Decodes masked time-frequency representation
    # default: applies decoder
    return decoded
```

With this structure, the only thing that is imposed to any class that inherits from `BaseEncoderMaskerDecoder` is the following diagram (again, in plain text):

```
(wav) --> Encoder --(tf_rep)----------------------> Mask_application ---(masked)---> Decoder ---> (decoded)
                            |                               1
                            |--> Mask_estimation --(masks)--|
```

Pre/post-processing would then be written directly in the inheriting classes.

For backward compatibility with classes that currently implements the `postprocess_*` hooks, we could call these in the `forward_*` and completely remove the hooks for v0.4? We would need to have tests ensuring backward compatibility for each of these models though :stuck_out_tongue:

mpariente: Also not sure but this does look good, and maybe more flexible. I'll need a bit of time to think about it. 

Apart from `DCUNet` and `DCCRN`, no model uses that IIRC, and we have no pretrained models for them, plus `DCCRN` is going to change.

jonashaag: Reminds me of this :) 
convtasnet
(from the Conv TasNet paper)

mpariente: I don't know this, what is ConvTasNet?

JorisCos: Hello ! I agree from what i have seen in `DCUNet` and `DCCRN` the changes are small and straightforward. Having `forward_{endoder,masker,decoder}` seems like the good way to go :)

mpariente:
So, let's make this

```python
def forward(self, wav):
   shape = jitable_shape(wav)
   wav = _unsqueeze_to_3d(wav)

   tf_rep = self.forward_encoder(wav)
   est_masks = self.forward_masker(tf_rep)
   masked_tf_rep = self.apply_mask(tf_rep, est_masks)
   decode = self.forward_decoder(masked_tf_rep)

   reconstructed = pad_x_to_y(decoded, wav)
   return _shape_reconstructed(reconstructed, shape)
```

And call the current hooks in those four methods. 

Who'd like to do it? @mhu-coder ?
We'll block this PR until then.

mhu-coder:
Sure. Quick questions, do I remove the backward compatibility tests since these
* are commented for CI
* would be added in #294 ?

mpariente:
Cool !
Let's leave this PR as is for now, and open a new PR to change `BaseEncoderMaskerDecoder`. Once the other one is merged, we'll come back here.

jonashaag:
Hook part LGTM

asteroid/models/demask.py
```diff
@@ -66,98 +87,59 @@ def __init__(
...
 
-        self.encoder, self.decoder = make_enc_dec(
-            fb_type,
-            kernel_size=kernel_size,
-            n_filters=n_filters,
-            stride=stride,
-            sample_rate=sample_rate,
-            **fb_kwargs,
-        )
+    def _get_n_feats_input(self, input_type, encoder):
```
mpariente, line 90: Take `n_feats_out` as input instead of encoder?

asteroid/models/demask.py
```diff
@@ -66,98 +87,59 @@ def __init__(
...
-        if wav.ndim == 2:
-            wav = wav.unsqueeze(1)
-        # Real forward
-        tf_rep = self.encoder(wav)
+        n_feats_input = encoder.n_feats_out // 2
+        if input_type == "cat":
+            n_feats_input += encoder.n_feats_out
+        return n_feats_input
+
+    def _get_n_feats_output(self, output_type, encoder):
```
mpariente, line 102: Same here.

asteroid/models/demask.py
```diff
@@ -178,3 +160,31 @@ def get_model_args(self):
         }
         model_args.update(self.fb_kwargs)
         return model_args
+
+
+def _build_masker_nn(
```
mpariente, line 165: I wonder if this is the good way to go. 
I know the architecture is really simple, but should this be a hidden function? 
For other models, we have a `nn.Module` that can be used with other features than asteroid filterbanks. 
This might be related to @popcornell 's comments on asteroid's source code organization.

asteroid/models/demask.py
```diff
@@ -178,3 +160,31 @@ def get_model_args(self):
...
+                nn.Conv1d(in_chan, hidden_dim, 1),
+                make_layer_norm(hidden_dim),
+                layer_activation,
+                nn.Dropout(dropout),
+            ]
+        )
+        in_chan = hidden_dim
+
+    net.extend([nn.Conv1d(in_chan, n_out, 1), activations.get(mask_act)()])
+    return net
```
mpariente, line 190: And why return the list and not the `Sequential` of it?

asteroid/models/demask.py
```diff
@@ -178,3 +160,31 @@ def get_model_args(self):
         }
         model_args.update(self.fb_kwargs)
         return model_args
+
```
mpariente, line 163: If we're refactoring we should probably use `self.encoder.get_config()` in `get_model_args`. 
If we make a `DeMaskMasker` module (bad name), with `get_config`, we don't need to write `get_model_args`. This is related to the next comment.

mpariente (commented):
This looks good in general. 
It raises some questions on code organization: where should masker architectures live? 

The current way is to have them in `masknn` and extend them to have encoder/decoder in `models`, is this the right way to do it? 
@popcornell suggests that masker architectures should be moved into `models` and `masknn` renamed to `nnet` or similar. 
What are your takes on this @mhu-coder @jonashaag ?

Even if demask's architecture is very simple, the current code is not consistent with the rest of the codebase and I think we should stay consistent.

asteroid/models/demask.py
popcornell: maybe having it as a method would be better. 

> we have a nn.Module that can be used with other features than asteroid filterbanks.
I must have missed this. What is this module ?

mpariente: > maybe having it as a method would be better.

A method in the class? Or a function outside (like now)? public or private? 

> I must have missed this. What is this module ?

Sorry, it was not clear. I mean the masker architectures e.g `TDConvNet` can be used on anything and is exposed to the user. Here, the masker architecture from `DeMask` is not.

mhu-coder:
> Even if demask's architecture is very simple, the current code is not consistent with the rest of the codebase and I think we should stay consistent.

Agree, it should stay consistent with the rest of the code base.

> @popcornell suggests that masker architectures should be moved into models and masknn renamed to nnet or similar.
What are your takes on this @mhu-coder @jonashaag ?

Looking at the base code, it seems that maskers are never used on their own and only in 1 model. So it would make sense to have them close to the models they define.
OTOH, having them in `asteroid.models` also makes it harder to compose them, if this need ever rises. I guess, for me it comes down to 2 questions:
1. How do you make the distinction between a "masker" architecture (which would go in `asteroid.models`) and a "building-block" architecture that can be used in a "masker" architecture (which would go in `asteroid.nnet` presumably)?
2. What is the need to move them at this point?

asteroid/models/demask.py
popcornell: > A method in the class? Or a function outside (like now)? public or private?

private Method in the class. 

> Sorry, it was not clear. I mean the masker architectures e.g TDConvNet can be used on anything and is exposed to the user. Here, the masker architecture from DeMask is not.

OK, well it is a simple MLP the masker architecture.... i don't know how much is useful for other tasks. 

BTW, can't in principle `BaseENcoderMaskerDecoder` be used ? Or it can't because of difference for STFT mag vs free for example.

popcornell:
> 1. How do you make the distinction between a "masker" architecture (which would go in asteroid.models) and a "building-block" architecture that can be used in a "masker" architecture (which would go in asteroid.nnet presumably)?
> 
> 2. What is the need to move them at this point?

I would say the distinction is dictated by how much that collection of layers is tied to one particular paper/architecture. 
Like the masker from convtasnet is quite binded to the whole convtasnet. Instead the DilateConvs blocks are also used in other architectures for example. 

No need to move for now, but as the models grow it can get confusing maybe.

asteroid/models/demask.py
mpariente: This PR makes that we use `BaseEncoderMaskerDecoder` for `DeMask` now. 

> OK, well it is a simple MLP the masker architecture.... i don't know how much is useful for other tasks.

I know that it's a very simple architecture but I wonder if it makes a difference. 

Too many choices for today for me :joy:

asteroid/models/demask.py
mhu-coder: Actually we do because `get_model_args` of the base class is missing the arguments `fb_kwargs`, `fb_type`, `input_type` and `output_type`. If we break backward compatibility, we can rename `fb_type` in `fb_name` and maybe remove `fb_kwargs` but have to keep the other arguments right?
Also, the `get_model_args` of the base class adds `encoder_activation`, `fb_name`, `in_chan` and `out_chan` arguments.

I just created a (very) simple class for the masker of `DeMask` but, unlike all the other maskers, it has to inherit from `nn.Sequential` if we want to be backward compatible. Otherwise, the state dict cannot load. Shall it be moved to `asteroid.masknn`?

Just waiting for comments before adding docstrings and type hinting.

jonashaag:
FWIW I don’t have an opinion on code structure other than that I feel like it is too early to decide on a structure. At this point, to me it feels like Asteroid is mostly a collection of “random” models, with some common building blocks factored out. (Which doesn’t mean that I don’t see the efforts that are going into keeping the code structured and clean, especially by @mpariente who does the vast majority of reviewing. It’s just that the project itself is relatively young, plus there are almost no examples of similar, more mature code bases that you can copy concepts from, simply because the entire business of collecting lots of deep learning models in a single large code base hasn’t been done many times before.) So, unless there is a reason to work on reorganization right now I suggest to wait for the project to grow a little bigger. I’m sure at some point we will have to actively work on reorganization just because we otherwise can’t get any work done anymore, and at that point usually it is much more clear what needs to be done.

mpariente:
In general, I agree that it's too early but I'm trying to be open to suggestions from the main contributors. Anyway, thinking about it isn't useless even if we don't change anything now.

asteroid/models/demask.py
mpariente: About `DeMask`'s masker. Sorry for being hesitant, and thanks for being so adapting. 
We'll have `build_demask_masker` as you write earlier, but as a public method. Inhereting from `Sequential` is a trick for BC, and I don't like it.  

You're (mostly) right about `get_model_args`. Why _mostly_?
- `in_chan` and `out_chan` are not added by `get_model_args`
- Additional kwargs passed to the filterbanks are handled in `Filterbank.get_config()` which makes me see that the current code is wrong wrt to `fb_kwargs` . 

```python
from asteroid import DeMask

model = DeMask(fb_type="param_sinc", min_low_hz=60)
In [53]: model.get_model_args()
Out[53]: 
{'input_type': 'mag',
 'output_type': 'mag',
 'hidden_dims': [1024],
 'dropout': 0.0,
 'activation': 'relu',
 'mask_act': 'relu',
 'norm_type': 'gLN',
 'fb_type': 'param_sinc',
 'n_filters': 512,
 'stride': 256,
 'kernel_size': 512,
 'fb_kwargs': {'min_low_hz': 60},
 'sample_rate': 16000,
 'min_low_hz': 60}
```

Again, thanks for the flexibility. I'll take it from here.

mpariente:
Sorry for force-pushing, but the commit revert didn't work and I didn't spend time to find out why. 

I kept the `fb_type` in the tests on purpose to attest BC. A follow up PR will change to `fb_name`.

Issue #279, opened 2020-10-15
Add OLA support to separate() method


jonashaag:
## 🚀 Feature

The `.separate()` method of `BaseModel` is very convenient but doesn't handle long audio files with OLA.

### Motivation

Make it easy for people to use `separate()` with audio files that are longer than the models

### Ideas

Two ideas:

1) Add OLA support to `.separate()` explicitly, i.e. params `ola_window_size`, `ola_hop_length` etc. and have it deal with OLA.

2) Move `.separate()` out of `BaseModel` to a function and have it accept any type of model, then you can just pass an OLA wrapper like this: `separate(LambdaOverlapAdd(model, window_size=...))`. This seem to work great (I have been doing that for some time now), and also means that the `.separate()` method probably shouldn't be part of `BaseModel` in the first place, as it is more generally applicable.

mpariente:
If the method is useful by itself, it makes sense to expose it outside of `BaseModel`.
Let's say we move `separate` outside `BaseModel` then, we could still have something like this. 

```python 
class BaseModel:
    def separate(self, *args, ola_window_size=None, ola_hop_length=None):
        net = self
        if ola_window_size is not None:
            net = LambdaOverlapAdd(self, ...)
        return separate(net, *args, *kwargs)
```

One thing to consider though: for now we have only waveform-based models under `BaseModel` but some other models can require preprocessing of the input wav and postprocessing of the net output to separate input data.  This is what motivated me to bind the `separate` method to the model.

If you'd like to make a PR to implement the second option, you can go ahead and we'll solve the problems as we go.

jonashaag:
Note to self: Also fix this https://github.com/mpariente/asteroid/blob/master/asteroid/models/base_models.py#L144 (should be before exists check)

mpariente:
Oh yes, thanks for the catch

PR #305, opened 2020-11-03
#279 OLA support for separate() and asteroid-infer

jonashaag:
Example:

```
asteroid-infer "mpariente/ConvTasNet_WHAM!_sepclean" --files ... --ola-window 8000 --ola-hop 4000
```

```py
from asteroid.foo import separate

ola_model = LambdaOverlapAdd(model, ...)
separate(ola_model, ...)
```


asteroid/dsp/overlap_add.py
```diff
@@ -10,9 +10,15 @@ class LambdaOverlapAdd(torch.nn.Module):
...
+    `LambdaOverlapAdd` works with `foo.separate` and the `asteroid-infer` CLI if
+    the given `nnet` implements `foo.Separatable`. (Both `asteroid.models.BaseModel`
+    and `asteroid.models.BaseEncoderMaskerDecoder` do.)
+
     Args:
         nnet (callable): Function to apply to each segment.
-        n_src (int): Number of sources in the output of nnet.
+        n_src (Optional[int]): Number of sources in the output of nnet.
+            If None, the number of sources is determined by the network's output,
+            but some correctness checks cannot be performed.
```
jonashaag, line 21: This was the easiest (but maybe not the cleanest) way to make OLA work with any/unknown models, like in asteroid-infer. The alternative would be to add `n_src` to the `Separatable` interface (ie. the model)

asteroid/foo.py
```diff
@@ -0,0 +1,162 @@
+import os
```
jonashaag, line 1: I didn't know where to stick all of this, so it lives in `asteroid.foo` for now :-D

asteroid/foo.py
```diff
@@ -0,0 +1,162 @@
...
+import torch
+import numpy as np
+import warnings
+
+try:
+    from typing import Protocol
```
jonashaag, line 7: Protocols are a new feature in Python 3.8 that define an interface without requiring the implementing classes to inherit some abstract base class (ie. duck typing). It is used only for documentation right now

asteroid/foo.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+from asteroid.utils import get_device
+
+
+class Separatable(Protocol):
+    """Things that are separatable.
+
+    In addition to supporting the protocol specified below, implementations must
+    also work with `asteroid.utils.get_device` (eg. any `torch.nn.Module`).
```
jonashaag, line 21: Wasn't worth adding this to the protocol. 99% of the time a Separatable instance will be a `nn.Module`. For `LambdaOverlapAdd`, `get_device` infers the correct device from `.parameters()` (wich is identical to `.nnet.parameters()`)

asteroid/foo.py
```diff
@@ -0,0 +1,162 @@
...
+                f"Received a signal with a sampling rate of {fs}Hz for a model "
+                f"of {model.sample_rate}Hz. You can pass `resample=True` to resample automatically."
+            )
+    to_save = numpy_separate(model, wav, **kwargs)
+
+    # Save wav files to filename_est1.wav etc...
+    for src_idx, est_src in enumerate(to_save):
+        base = ".".join(filename.split(".")[:-1])
+        save_name = base + "_est{}.".format(src_idx + 1) + filename.split(".")[-1]
+        if output_dir is not None:
```
jonashaag, line 150: Swapped order as noted in #279

asteroid/foo.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class Separatable(Protocol):
+    """Things that are separatable.
+
+    In addition to supporting the protocol specified below, implementations must
+    also work with `asteroid.utils.get_device` (eg. any `torch.nn.Module`).
+    """
+
+    def _separate(self, wav, **kwargs):
```
jonashaag, line 24: Having one method start with `_` is a bit ugly but I didn't know how to change it without being backwards incompatible

asteroid/scripts/asteroid_cli.py
```diff
@@ -17,24 +20,19 @@
...
-        publish_dir (str): Path to the publishing directory.
-            Usually under exp/exp_name/publish_dir
-        uploader (str): Full name of the uploader (Ex: Manuel Pariente)
-        affiliation (str, optional): Affiliation (no accent).
-        git_username (str, optional): GitHub username.
-        token (str): Access token generated to upload depositions.
-        force_publish (bool): Whether to directly publish without
-            asking confirmation before. Defaults to False.
-        use_sandbox (bool): Whether to use Zenodo's sandbox instead of
-            the official Zenodo.
```
jonashaag, line 33: Removed these docstrings since these aren't actually function params

asteroid/scripts/asteroid_cli.py
```diff
@@ -124,20 +112,59 @@ def infer():
...
     file_list = _process_files_as_list(args.files)
 
     for f in file_list:
-        model.separate(f, force_overwrite=args.force_overwrite, output_dir=args.output_dir)
+        separate(
+            model,
+            f,
+            force_overwrite=args.force_overwrite,
+            output_dir=args.output_dir,
+            resample=args.resample,
```
jonashaag, line 162: Fixed resample not being passed to separate

asteroid/scripts/asteroid_cli.py
```diff
@@ -90,25 +88,15 @@ def upload():
...
-            of filenames, directory names and globs.
-        force_overwrite (bool): Whether to overwrite output wav files.
-        resample (bool): Whether to resample wrong sample rate input files.
-        output_dir (str): Output directory to save files.
-    """
-    import argparse
-
+    """CLI function to run pretrained model inference on wav files."""
     parser = argparse.ArgumentParser()
     parser.add_argument("url_or_path", type=str, help="Path to the pretrained model.")
```
jonashaag, line 93: Unrelated to this PR, but I think this is not very good design. You will always pass exactly 1 model, but N files. Usually you'd use a `--named-flag` for the model here, and unnamed/positional arguments for the files: `asteroid-infer -m mymodel file1.wav file2.wav ...`.

asteroid/dsp/overlap_add.py
```diff
@@ -116,7 +128,7 @@ def ola_forward(self, x):
             padding=(self.window_size, 0),
             stride=(self.hop_size, 1),
         )
-        return out.squeeze(-1).reshape(batch, self.n_src, -1)
+        return out.squeeze(-1).reshape(batch, n_src, -1)
 
     def forward(self, x):
```
mpariente, line 133: A bit unrelated but we should be able to pass kwargs from here to the network.
With a warning that we don't segment the kwargs so they are shared between chunks.

asteroid/foo.py
mpariente: Also not clear to me where to put this yet ^^

mpariente (commented):
This is a very cool PR. Having a functional `separate` is more powerful that tied to a class!
There is one thing which looks wrong to me, few things to be fixed and others to be discussed. 

Thanks again

asteroid/dsp/overlap_add.py
jonashaag: No because I didn’t want to break the interface

asteroid/scripts/asteroid_cli.py
jonashaag: Oh... Didn’t think about that. I will check if there is a better way.

asteroid/dsp/overlap_add.py
```diff
@@ -132,6 +144,15 @@ def forward(self, x):
...
             return olad
 
+    # Implement `foo.Separatable` (separation support)
+
+    @property
+    def sample_rate(self):
+        return self.nnet.sample_rate
+
+    def _separate(self, *args, **kwargs):
+        return self.nnet._separate(*args, **kwargs)
```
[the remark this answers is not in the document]
jonashaag: Oops

asteroid/dsp/overlap_add.py
mpariente: Ok.

mpariente:
Should we fix this and merge it?

asteroid/foo.py
```diff
@@ -0,0 +1,162 @@
...
+    # Python < 3.8
+    class Protocol:
+        pass
+
+
+from asteroid.utils import get_device
+
+
+class Separatable(Protocol):
+    """Things that are separatable.
```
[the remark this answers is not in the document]
jonashaag: Oh think "separatable" would be more common, but from a quick Google it seems the more appropriate term would be "separable". Shall we change to that?

jonashaag:
Addressed most of the suggestions. Did not change the `_separate` name so far. Will continue working on it later

mpariente:
The `_separate` can be change in a follow up PR, it'll be better that way

asteroid/scripts/asteroid_cli.py
jonashaag: Bildschirmfoto 2020-11-19 um 13 10 34

What do you think about this? It is autogenerated from `--help`.

Or do you want actual Python code docs for people that want to work on the CLI code itself?

mpariente: I think this is reasonable, but it's a bit ugly. 
Anyway, we should probably change the way we make CLI (using `click`, `typer` or `fire` which I discovered only recently), so yes, it does look fine for now.

asteroid/foo.py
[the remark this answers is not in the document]
jonashaag: @mpariente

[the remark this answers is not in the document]
mpariente: I thought that this was a naming convention. That because the method was `separate`, we called it `Separatable`. Isn't this the case?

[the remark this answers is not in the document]
jonashaag: Yes but seems like to proper name would be "separable".

We could also call the interface `Separate`, which is common eg. in Rust (eg. https://doc.rust-lang.org/reference/special-types-and-traits.html#copy instead of Copyable).

Let's just pick any name, I don't really care :D

jonashaag:
We still need a name for all of this that's better than `asteroid.foo`! Maybe @popcornell you have an idea?

popcornell:
Should be in engine maybe ? 
Like` engine.inference`. 
Then you can put into the asteroid` __init__` and maybe have `from asteroid import separate`.

asteroid/scripts/asteroid_cli.py
mpariente: Ok, thanks. I haven't been using UNIX for so long so I don't know much.
Note that we support globs in --files, but I understand the point.

mpariente:
`engine` is only related to training for now. I would keep it out of there. 
What about simply `asteorid/separate.py`?

jonashaag:
`asteroid.separate` sounds ok

popcornell:
Also good. If we plan to use `engine` only for training in 1.0 we should rename to `training` probably.

mpariente:
Haha, yes ^^

mpariente:
This mainly looked ready right? 
Let's rename to `asteroid.separate` and review again.

jonashaag:
yes I'll do it

jonashaag:
Oh, it already ran the tests, but didn't refresh the page. Well, then this seems good to go :)

asteroid/separate.py
```diff
@@ -0,0 +1,169 @@
...
+try:
+    from typing import Protocol
+except ImportError:  # noqa
+    # Python < 3.8
+    class Protocol:
+        pass
+
+
+from asteroid.dsp.overlap_add import LambdaOverlapAdd
+from asteroid.utils import get_device
```
mpariente, line 15:
```suggestion
from .dsp.overlap_add import LambdaOverlapAdd
from .utils import get_device
```

mpariente (approved):

PR #306, opened 2020-11-03
FasNet TAC integration

popcornell:
As the title implies. It was about time. 

Doubts I have: 
1. where to put the model and its components (especially the NCC module probably is better to move it in `dsp`)
   (IMO in future we will have to refactor and re-organize the whole masknn ).
2. the .pkl for the dataset generation should be moved elsewhere probably or fetched directly by cloning luo's repo.
3. eval script missing right now (TODO)
4. also docstrings (TODO)


mpariente:
Cool, thanks !

> IMO in future we will have to refactor and re-organize the whole masknn

Can you explain why, in your opinion, please?

mpariente:
Another thing: I know that if we merge this without squashing, the binary files will be in the git history,  even if removed from the PR. This would be problematic because they are quite heavy. 

But, I'm not sure what happens if we squash. My intuition is that they won't be in history. I know @iver56 has some knowledge about this, do you know about this particular case?

iver56:
> 
> 
> Another thing: I know that if we merge this without squashing, the binary files will be in the git history, even if removed from the PR. This would be problematic because they are quite heavy.
> 
> But, I'm not sure what happens if we squash. My intuition is that they won't be in history. I know @iver56 has some knowledge about this, do you know about this particular case?

I'm not an expert on this topic, but I think if you squash the commits (where one of them added a file and a later one removed it again), the squashed commit does not include the file. So I think if you include only the squashed commit, and discard the original, offending commits, the repository won't contain the big file. Maybe `git gc` is needed for pruning loose objects. You can check the repository size at https://api.github.com/repos/mpariente/asteroid

mpariente:
Thanks a lot for the feedback and resources Iver !

popcornell:
Your call really IMO if we squash there should not be in the history as long as I remove them.  If we wanna be sure i can make another pull request. 

Sure,  IMO we have to be more modular and distinguish between models and building blocks. 
For example: in `attention.py` there is DPTNet and is imported and wrapped in `models`. 
I think in attention.py only building blocks should be included. And i would rename masknn to nnet or  blocks or smthing like that. 
I would put both DPTNet and DPTAsNet in models.

mpariente:
> Your call really IMO if we squash there should not be in the history as long as I remove them. If we wanna be sure i can make another pull request.

Let's keep it that way. I think it'll be ok. 

You say that mask nets should be considered as models, it makes sense. 
I don't have a clear idea of how to organize it better, or why it is currently limiting. I'm happy to discuss about that.

asteroid/masknn/tac.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TAC(nn.Module):
+    def __init__(self, input_dim, hidden_dim=384, activation="prelu", norm_type="gLN"):
+        super().__init__()
+
+        self.hidden_dim = hidden_dim
+        self.input_tf = nn.Sequential(
+            nn.Linear(input_dim, hidden_dim), activations.get(activation)()
+        )
```
JorisCos, line 13: Adding a `self.activation = activations.get(activation)`  and `self.norm = norm_type` could save space and makes things a bit clearer

JorisCos (commented):
Thanks for the PR ! Looks great !  I will take a look again once everything is added :)

popcornell:
What do you think about the NCC module we move it to DSP ?

JorisCos:
> What do you think about the NCC module we move it to DSP ?

IMO yes

popcornell:
Also keep in mind this is the first step towards Multi-Channel support so we may want also to start to think about how to organize the toolkit also to support multi-channel separation and enhancement algorithms and building blocks.

mpariente:
@popcornell do you feel like writing simple docstrings on the parts in the source code? Reviewing will be easier from there.

popcornell:
Yes but i have limited time available this week

asteroid/models/fasnet.py
```diff
@@ -0,0 +1,298 @@
...
+            "dropout": self.dropout,
+            "use_tac": self.use_tac,
+        }
+
+        return config
+
+    @property
+    def sample_rate(self):
+        """Operating sample rate of the model (float)."""
+        raise self.samplerate
```
iver56, line 298:
```suggestion
        return self.samplerate
```

mpariente: Thanks. We'll have to pass `sample_rate` to `BaseModel` and remove that part anyway, but in the mean time, at least it makes sense, thanks.

asteroid/masknn/tac.py
popcornell: but activation must be different for each MLP block

popcornell:
Added docstrings, removed .pkl config files. Added comments, added a dsp.spatial module with fast multi-channel cross-correlation.

mpariente (commented):
Just an empty review so that you can require a new review when the PR is ready.

mpariente:
Great! 
Let's merge master into the branch because this is a mess to review.

asteroid/dsp/spatial.py
```diff
@@ -0,0 +1,65 @@
+import torch
+import torch.nn.functional as F
+
+
+def xcorr(ref, input, normalized=True, eps=1e-8):
```
JorisCos, line 5: Named after the Matlab function ?

asteroid/dsp/spatial.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Efficient cross-correlation computation between multi-channel signals.
+    Input tensors must be of shape: mic_channels, batch, seq_len.
```
JorisCos, line 8: I wonder if it should expect batch, mic_channels, seq_len to keep having the same input format everywhere

egs/TAC/run.sh
```diff
@@ -0,0 +1,123 @@
...
+
+  if ! test -e $librispeech_dir/dev-clean; then
+    echo "Downloading LibriSpeech/dev-clean into $storage_dir"
+	  wget -c --tries=0 --read-timeout=20 http://www.openslr.org/resources/12/dev-clean.tar.gz -P $storage_dir
+	  tar -xzf $storage_dir/dev-clean.tar.gz -C $storage_dir
+	  rm -rf $storage_dir/dev-clean.tar.gz
+	fi
+
+  if ! test -e $librispeech_dir/test-clean; then
+    echo "Downloading LibriSpeech/dev-clean into $storage_dir"
```
JorisCos, line 60: LibriSpeech/test-clean

JorisCos (commented):
Seeing multichannel integration into Asteroid is super exciting, thanks for this PR. We need tests for the model and the xcor function

asteroid/dsp/spatial.py
mpariente: Yes, I think that would be much better!

asteroid/dsp/spatial.py
popcornell: maybe, i named that way only because it sounds cooler i did not think about MATLAB

asteroid/dsp/spatial.py
popcornell: Agree. What do you think that right now it does not compute all possible pairs when both inputs are multi-channel ? We do not need that functionality right now but we might need it in future.

popcornell:
> We need tests for the model and the xcor function

it is at -9.3 on the multichannel dataset. epoch 36
Both are working quite well

mpariente:
Joris meant unit tests, to make sure we don't break things when we edit the files next time.

asteroid/dsp/spatial.py
mpariente: I guess that's fine for now. Probably by using broadcasting it can be achieved quite cheap.

asteroid/dsp/spatial.py
mpariente: It sounds cool! :smiley:

asteroid/masknn/tac.py
```diff
@@ -0,0 +1,80 @@
...
+        self.avg_tf = nn.Sequential(
+            nn.Linear(hidden_dim, hidden_dim), activations.get(activation)()
+        )
+        self.concat_tf = nn.Sequential(
+            nn.Linear(2 * hidden_dim, input_dim), activations.get(activation)()
+        )
+        self.norm = norms.get(norm_type)(input_dim)
+
+    def forward(self, x, valid_mics):
```
mpariente, line 40: Can you add a docstring to say what valid_mics is please?

asteroid/models/fasnet.py
```diff
@@ -0,0 +1,284 @@
...
+
+        n_chunks = unfolded.size(-1)
+        unfolded = unfolded.reshape(batch_size, nmic, window + 2 * context, n_chunks)
+
+        return (
+            unfolded[:, :, context : context + window].transpose(2, -1),
+            unfolded.transpose(2, -1),
+        )
+
+    def forward(self, x, valid_mics):
```
mpariente, line 142: Same here, I don't know what is `valid_mics`.

asteroid/models/fasnet.py
```diff
@@ -0,0 +1,284 @@
...
+                bf_signal[b, : valid_mics[b]].mean(0).unsqueeze(0) for b in range(batch_size)
+            ]
+            bf_signal = torch.cat(bf_signal, 0)
+
+        return bf_signal
+
+    def get_model_args(self):
+        import ipdb
+
+        ipdb.set_trace()
```
mpariente, line 263: Let's remove that.

egs/TAC/local/tac_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+import numpy as np
+from pathlib import Path
+from asteroid.data.librimix_dataset import librispeech_license
+
+
+class TACDataset(Dataset):
+
+    dataset_name = "Multi-channel Librispeech-derived dataset used in Transform Average Concatenate"
+
+    def __init__(self, json_file, segment=None, samplerate=16000, max_mics=6, train=True):
```
mpariente, line 14: We'll need docs here.

egs/TAC/local/tac_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+import soundfile as sf
+import torch
+import numpy as np
+from pathlib import Path
+from asteroid.data.librimix_dataset import librispeech_license
+
+
+class TACDataset(Dataset):
+
+    dataset_name = "Multi-channel Librispeech-derived dataset used in Transform Average Concatenate"
```
mpariente, line 12: Maybe this description goes in the docstring, and we call it something smaller? 
The name will be used as a tag in the HF's hub.

egs/TAC/run.sh
```diff
@@ -0,0 +1,123 @@
...
+
+  #git clone https://github.com/yluo42/TAC ./local/TAC
+  cd local/TAC/data
+  #$python_path create_dataset.py \
+   #             --output-path=$storage_dir \
+#		--dataset=$dataset_type \
+#		--libri-path=$librispeech_dir \
+#		--noise-path=$noise_dir
+cd ../../../	
+fi
```
mpariente, line 86: So, we're missing this, right?

egs/TAC/run.sh
```diff
@@ -0,0 +1,123 @@
...
+# Exit on error
+set -e
+set -o pipefail
+
+# Main storage directory where dataset will be stored
+storage_dir=$(readlink -m ./datasets)
+
+librispeech_dir=$storage_dir/LibriSpeech
+noise_dir=$storage_dir/Nonspeech
+CONF_FILE=./local/conf.yml
```
mpariente, line 12: This is unused. 
I understand that we would want to limit the confusion between conf args and bash args, but here, there is no way to have several config files.

Let's add an argument for that in the train file.

egs/TAC/train.py
```diff
@@ -0,0 +1,156 @@
...
+
+if __name__ == "__main__":
+    import yaml
+    from pprint import pprint as print
+    from asteroid.utils import prepare_parser_from_dict, parse_args_as_dict
+
+    # We start with opening the config file conf.yml as a dictionary from
+    # which we can create parsers. Each top level key in the dictionary defined
+    # by the YAML file creates a group in the parser.
+    with open("local/conf.yml") as f:
```
mpariente, line 145: Pass conf file from bash and load here.

mpariente (commented):
It's very cool! 
The comments of CodeCov make it reading the code quite hard, so I disabled them. From the next changes, it should be better. 

Some comments: 
- Dataset creation is not ready right? 
- We would need docs for the Dataset part. 
- Let's add comments on the `valid_mics`. I wonder if the padding to `max_mics` couldn't be done in the network instead of in the dataloader? WDYT? 
- We need unit tests for xcorr and the architectures. 

I think it makes sense to experiment with a conf-file only recipe, for perspective, but let's add a comment about that in the beginning of run.sh.

egs/TAC/train.py
popcornell: good catch thanks

egs/TAC/run.sh
popcornell: no i forgot to re-comment it back. It is working

egs/TAC/local/tac_dataset.py
popcornell: TACDataset ?

asteroid/models/fasnet.py
popcornell: I need a linter which checks for ipdb

popcornell:
> Joris meant unit tests, to make sure we don't break things when we edit the files next time.

yeah i know but I hate them

egs/TAC/run.sh
popcornell: added

popcornell:
> I wonder if the padding to max_mics couldn't be done in the network instead of in the dataloader? WDYT?
You can't because you have to pad in order to batch examples with different number of mics.

mpariente:
Yeah, I realized that after I asked the question, thanks.

mpariente:
Re-request my review when you're done please

asteroid/dsp/spatial.py
```diff
@@ -2,64 +2,56 @@
...
 
     """
-    # input has size batches, nmics2, seq_len2
-    # ref has size  batches, nmics1, seq_len1
-
-    assert input.size(0) == ref.size(0), "ref and input signals should have same batch size."
-    assert input.size(2) >= ref.size(2), "Input signal should not be shorter than the ref signal."
+    # inp: batch, nmics2, seq_len2 || ref: batch, nmics1, seq_len1
+    assert inp.size(0) == ref.size(0), "ref and inp signals should have same batch size."
+    assert inp.size(2) >= ref.size(2), "Input signal should be shorter than the ref signal."
```
mpariente, line 30: Do you agree that  "Input signal should not be shorter than the ref signal." was wrong?

asteroid/dsp/spatial.py
```diff
@@ -2,64 +2,56 @@
 import torch.nn.functional as F
 
 
-def xcorr(input, ref, normalized=True, eps=1e-8):
-    """
-    Efficient cross-correlation computation between multi-channel signals.
-    Input tensors must be of shape: batch, mic_channels, seq_len.
-    The two signals can have different seq_lens but the reference signal should not be shorter than the input signal.
+def xcorr(inp, ref, normalized=True, eps=1e-8):
```
mpariente, line 5: `input` is a built-in so I replaced by `inp`

egs/TAC/local/tac_dataset.py
```diff
@@ -0,0 +1,145 @@
+from torch.utils.data import Dataset
```
mpariente, line 1: I wonder why not adding this is asteroid/data? 
Is there a reason? Do you feel it's too recipe specific?

egs/TAC/train.py
```diff
@@ -0,0 +1,156 @@
...
+    import yaml
+    from pprint import pprint as print
+    from asteroid.utils import prepare_parser_from_dict, parse_args_as_dict
+
+    # We start with opening the config file conf.yml as a dictionary from
+    # which we can create parsers. Each top level key in the dictionary defined
+    # by the YAML file creates a group in the parser.
+    with open("./local/conf.yml") as f:
+        def_conf = yaml.safe_load(f)
+    parser = prepare_parser_from_dict(def_conf, parser=parser)
```
mpariente, line 149: I misunderstood something. 
I thought that you wanted to pass config files directly, instead of the usual way, and that the conf file could be passed in run.sh?

mpariente (approved):
We can almost merge. Things to sort out before: 
- [x] Location of `TACDataset`
- [x] Clarification on the config etc..
- [x] Rebase for the multichannel support, and tests serialization. 
- [x] Test the broadcasting of xcorr.

mpariente:
\rebase

asteroid/masknn/tac.py
```diff
@@ -52,6 +52,8 @@ def forward(self, x, valid_mics):
         """
         # Input is 5D because it is multi-channel DPRNN. DPRNN single channel is 4D.
         batch_size, nmics, channels, chunk_size, n_chunks = x.size()
+        if valid_mics is None:
+            valid_mics = torch.LongTensor([nmics] * batch_size)
```
mpariente, line 56: If we don't have a default value, we cannot use it in the inference CLI. 

If there is no `valid_mics`, we consider all recordings to have the max number of mics. 
@popcornell , can you validate that this is ok?

popcornell: Yes this is a sensible default

mpariente: Cool thanks.

egs/TAC/local/tac_dataset.py
popcornell: This dataset is problematic if you remember due to lack of spatial diversity in mics and the limited noise types.
I would not add it to the toolkit IMO.

mpariente:
Also, the model currently only supports 3D inputs, which won't work with inference. 
I don't know what to do about that. 
In `BaseEncoderMaskerDecoder`, we have these three funcs to handle input shapes in a jitable way, should we also integrate in `FasNetTAC`?
IMO, we can merge like this now and see if somebody can work on the jitability afterwards, maybe then, move these shape related stuff into `BaseModel` which can be useful in other models.

asteroid/dsp/spatial.py
popcornell: yes missing the not

mpariente:
I think there is a confusion in the `enc_dim` and `feature_dim` because if I change one without the other, the forward doesn't work.

egs/TAC/local/tac_dataset.py
mpariente: Ok, cool

mpariente:
Great, the problem seems to be solved for `feature_dim`, there is still a problem with `enc_dim` though..

popcornell:
Now should be solved

mpariente:
GREAT !

PR #307, opened 2020-11-04
Simplify base encoder masker decoder

mhu-coder:
Simplifies structure of `BaseEncoderMaskerDecoder.forward` although the hooks are still called in the `forward_*` and `apply_masks` methods.

I am not sure about the shapes given the docstrings when they are indicated, and when they are not, it's because I didn't know if there was supposed to be an expected shape x).

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -302,9 +302,7 @@ class BaseEncoderMaskerDecoder(BaseModel):
 
         # Real forward
         tf_rep = self.forward_encoder(wav)
-
-        est_masks = self.masker(tf_rep)
-        est_masks = self.postprocess_masks(est_masks)
+        est_masks = self.forward_masker(tf_rep)
 
         masked_tf_rep = est_masks * tf_rep.unsqueeze(1)
         masked_tf_rep = self.postprocess_masked(masked_tf_rep)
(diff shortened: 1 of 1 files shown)


mpariente:
LGTM

jonashaag:
Looks ok, but personally I would remove the hooks entirely now that we have a more powerful concept. Otherwise, if someone suggests new places to add hooks in this pipeline, where do we draw the line between hooks that we include or exclude?

mpariente:
How do you remove the hooks without breaking BC?

mhu-coder:
Removing the hooks here without warning will give a rather poor experience to people that have forked the project and written classes overriding the hooks...

Since there are a couple of things that will not be backward compatible in 0.4, would it be possible to add the removal of the hooks to this list? We could add this change as well as renaming `fb_type` to `fb_name` in `DeMask` to the roadmap.

I agree with Jonas, if we leave this as is for too long, it may become really hard to switch in the future.

jonashaag:
> How do you remove the hooks without breaking BC?

Two ways:

1. Just break BC as you suggested in the other thread (if I understood correctly). It's only been out for a month, and migrating code to the new API is simple.
2. Merge this PR as-is but raise deprecation warning in the hooks

mpariente:
I don't suggest we break BC. For scriptability, it was not an option.

Ok to break BC on this. So
- Merge this as is.
- Rewrite asteroid code that used the former hooks.
- Remove the hooks. 

Right?

mpariente (approved):

PR #308, opened 2020-11-04
DCUNet: Replace old hooks by new one

mpariente:
Another step towards removing the old hooks.
Ref #307

### DIFF
--- a/asteroid/models/dcunet.py
+++ b/asteroid/models/dcunet.py
@@ -42,10 +42,12 @@ class BaseDCUNet(BaseEncoderMaskerDecoder):  # CHECK-JIT
         masker = self.masknet_class.default_architecture(architecture, **(masknet_kwargs or {}))
         super().__init__(encoder, masker, decoder)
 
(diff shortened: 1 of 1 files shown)


asteroid/models/dcunet.py
```diff
@@ -42,10 +42,12 @@ def __init__(
         masker = self.masknet_class.default_architecture(architecture, **(masknet_kwargs or {}))
         super().__init__(encoder, masker, decoder)
 
-    def postprocess_encoded(self, tf_rep):
+    def forward_encoder(self, wav):
+        tf_rep = self.encoder(wav)
```
jonashaag, line 46: Yes, or `return complex_nn.as_torch_complex(super().forward_encoder(wav))` and below `from_torchaudio(torch.view_as_real(super().apply_masks(tf_rep, est_masks))`

I have no preference

mpariente: I prefer the current way, it's clearer.

Issue #315, opened 2020-11-05
Getting started/inference docs


jonashaag:
I was missing 1) documentation about `asteroid-infer`, 2) a simple pretrained-inference-only tutorial, 3) a reference to your nice Interspeech YouTube video, so here is a draft of a very very simple tutorial for absolute beginners on how to use a pretrained model.

Feedback welcome :) I don't know how to actually collaboratively work on these notebooks, so here is a static link.

https://colab.research.google.com/gist/jonashaag/1cc34570b0e6e26dc35e5dc843cbeef3/01-asteroidgettingstarted.ipynb

mpariente:
I love it ! 

I also don't know how to do that either haha 

Feedback:
- This is very concise and straight to the point, it's perfect!
- I'd like to see a pretrained model loading in python as well, like this. 

```python
from asteroid import DPRNNTasNet

model = DPRNNTasNet.from_pretrained("mpariente/DPRNNTasNet(ks=16)_WHAM!_sepclean")
model.separate("female-female-mixture.wav")
```

- I'd use `%%capture` above the pip and wget cells to focus the user on the output that matters. 
- The plotting at the end makes me think that we should probably have a similar feature :smiley: 
I don't know if the presentation video is very informative, but if it is, we could add it to the docs somewhere.

mpariente:
And, we should rename the 01_AsteroidGettingStarted.ipynb notebook, because it's not clear.

jonashaag:
Incorporated the feedback. Also added training section.

I didn't have the patience to let it run for an hour or so, maybe next week.

I think having a video is good. Some people learn a lot better using videos and audio than reading stuff.

mpariente:
It looks great, you can open a PR !

We cannot expect the model to sound well, even after 200 epochs, MiniLibriMix is really small. 
So this should only be for demonstration purpose, and we don't need to let it train for 20 epochs.

mpariente:
Closed by #319

PR #319, opened 2020-11-07
Add an actual Getting Started notebook

jonashaag:
See #315 

Feel free to change names etc. (or anything else in this PR :))

Maybe we should move the "Notebooks and Tutorials" section above the "Asteroid" section in the docs?


mpariente:
I moved the sections, and corrected the link to the 00 notebook. 
This is a great notebook, much better than the previous "getting started"

PR #320, opened 2020-11-09
Mixture Invariant Training

giorgiacantisani:
Hi!

I implemented a mixture invariant training loss wrapper inspired by the paper "Unsupervised sound separation using mixtures of mixtures." arXiv preprint arXiv:2006.12701 (2020). 

Specifically, there are two new options for PITLossWrapper:
1. `mix_it`: Find the best partition of the estimated sources that gives the minimum loss for the MixIT training paradigm. Valid for any number of mixtures as soon as they contain the same number of sources.
2. `mix_it_gen`: Find the best partition of the estimated sources that gives the minimum loss for the MixIT training paradigm. Valid only for two mixtures, but those mixtures do not necessarily have to contain the same number of sources. It is allowed the case where one mixture is silent.

### DIFF
--- a/asteroid/losses/pit_wrapper.py
+++ b/asteroid/losses/pit_wrapper.py
@@ -8,7 +8,7 @@ class PITLossWrapper(nn.Module):
     r"""Permutation invariant loss wrapper.
 
     Args:
-        loss_func: function with signature (targets, est_targets, **kwargs).
+        loss_func: function with signature (est_targets, targets, **kwargs).
         pit_from (str): Determines how PIT is applied.
 
             * ``'pw_mtx'`` (pairwise matrix): `loss_func` computes pairwise
@@ -135,7 +135,7 @@ class PITLossWrapper(nn.Module):
         for a given loss function.
 
         Args:
-            loss_func: function with signature (targets, est_targets, **kwargs)
+            loss_func: function with signature (est_targets, targets, **kwargs)
                 The loss function to get pair-wise losses from.
             est_targets: torch.Tensor. Expected shape [batch, nsrc, *].
                 The batch of target estimates.
@@ -164,7 +164,7 @@ class PITLossWrapper(nn.Module):
         """Find best permutation from loss function with source axis.
 
         Args:
-            loss_func: function with signature (targets, est_targets, **kwargs)
+            loss_func: function with signature (est_targets, targets, **kwargs)
                 The loss function batch losses from.
             est_targets: torch.Tensor. Expected shape [batch, nsrc, *].
                 The batch of target estimates.
--- a/asteroid/losses/pit_wrapper.py
+++ b/asteroid/losses/pit_wrapper.py
@@ -4,7 +4,7 @@ from torch import nn
 
 
 class PITLossWrapper(nn.Module):
-    """ Permutation invariant loss wrapper.
+    r""" Permutation invariant loss wrapper.
 
     Args:
         loss_func: function with signature (targets, est_targets, **kwargs).
--- a/asteroid/losses/pit_wrapper.py
+++ b/asteroid/losses/pit_wrapper.py
@@ -1,11 +1,10 @@
-from itertools import permutations
+from itertools import permutations, combinations
 import torch
 from torch import nn
-from scipy.optimize import linear_sum_assignment
 
 
 class PITLossWrapper(nn.Module):
-    r"""Permutation invariant loss wrapper.
+    """ Permutation invariant loss wrapper.
 
     Args:
         loss_func: function with signature (targets, est_targets, **kwargs).
@@ -24,6 +23,18 @@ class PITLossWrapper(nn.Module):
               average loss for a given permutations of the sources and
               estimates. Output shape : :math:`(batch)`.
               See :meth:`~PITLossWrapper.best_perm_from_perm_avg_loss`.
+            * ``'mix_it'``(mixture invariant): `loss_func` computes the
+              loss for a given partition of the sources. Valid for any
+              number of mixtures as soon as they contain the same number
+              of sources.
+              Output shape : :math:`(batch)`.
+              See :meth:`~PITLossWrapper.best_part_mix_it`.
+            * ``'mix_it_gen'``(mixture invariant generalized): `loss_func`
+              computes the loss for a given partition of the sources.
+              Valid only for two mixtures, but those mixtures do not
+              necessarly have to contain the same number of sources.
+              Output shape : :math:`(batch)`.
+              See :meth:`~PITLossWrapper.best_part_mix_it_gen`.
 
             In terms of efficiency, ``'perm_avg'`` is the least efficicient.
 
@@ -38,7 +49,7 @@ class PITLossWrapper(nn.Module):
     For each of these modes, the best permutation and reordering will be
     automatically computed.
 
-    Examples
+    Examples:
         >>> import torch
         >>> from asteroid.losses import pairwise_neg_sisdr
         >>> sources = torch.randn(10, 3, 16000)
@@ -58,20 +69,18 @@ class PITLossWrapper(nn.Module):
         >>> loss_val = loss_func(est_sources, sources,
         >>>                      reduce_kwargs=reduce_kwargs)
     """
-
-    def __init__(self, loss_func, pit_from="pw_mtx", perm_reduce=None):
+    def __init__(self, loss_func, pit_from='pw_mtx', perm_reduce=None):
         super().__init__()
         self.loss_func = loss_func
         self.pit_from = pit_from
         self.perm_reduce = perm_reduce
-        if self.pit_from not in ["pw_mtx", "pw_pt", "perm_avg"]:
-            raise ValueError(
-                "Unsupported loss function type for now. Expected"
-                "one of [`pw_mtx`, `pw_pt`, `perm_avg`]"
-            )
+        if self.pit_from not in ['pw_mtx', 'pw_pt', 'perm_avg', 'mix_it', 'mix_it_gen']:
+            raise ValueError('Unsupported loss function type for now. Expected'
+                             'one of [`pw_mtx`, `pw_pt`, `perm_avg`, `mix_it`, `mix_it_gen`]')
 
-    def forward(self, est_targets, targets, return_est=False, reduce_kwargs=None, **kwargs):
-        """Find the best permutation and return the loss.
+    def forward(self, est_targets, targets, return_est=False,
+                reduce_kwargs=None, **kwargs):
+        """ Find the best permutation and return the loss.
 
         Args:
             est_targets: torch.Tensor. Expected shape [batch, nsrc, *].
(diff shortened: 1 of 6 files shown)


popcornell:
Thank you very much for your pull request !

I have also a very naive implementation of MixIT on my local machine but you beat me on time ;) ! I think we can merge the two together (especially for the mixing matrix generation part). 
Did you also had time to test it on some dataset ? Do you mind if I push some examples using mixit here ? 

Also @mpariente  how do you feel having mixit as an extension of current `PITLossWrapper` ? Theoretically it is sound but maybe could be confusing for some users.

giorgiacantisani:
Yes sure we can definitely merge them! I didn't focus to much on optimizing it and maybe you've a faster implementation. 

I tested it with the MUSDB18 dataset and seems to work!

mpariente:
Thank you very much for the implementation, that's great !

PIT and MixIT are essentially different training paradigm (though related), so I also think making a separate class would be a good idea. 

It seems that the first commit in this PR reverts recent changes made to `PITLossWrapper` (black compliance, doc rendering and others). We want to keep those changes, not revert them, plus it makes it more difficult to review the proposed changes.
Could you start again from `master` and only apply changes related to MixIT please? 
(No need to make a separate class for now, we'll converge on that afterwards)

Note to self: we start to have several extensions of `PITLossWrapper`, start to think about factoring out common logics/method.

giorgiacantisani:
Yes, I am sorry I was working on an old version of the file and didn't realize it until the PR. I'll fix this

tests/losses/pit_wrapper_test.py
```diff
@@ -114,19 +152,3 @@ def reduce_func(perm_losses, class_weights=None):
...
-    min_loss, min_idx = PITLossWrapper.find_best_perm_factorial(pwl)
-    min_loss_hun, min_idx_hun = PITLossWrapper.find_best_perm_hungarian(pwl)
-
-    assert_allclose(min_loss, min_loss_hun)
-    assert_allclose(min_idx, min_idx_hun)
-
-
-def test_raises_wrong_pit_from():
-    with pytest.raises(ValueError):
-        PITLossWrapper(lambda x: x, pit_from="unknown_mode")
```
mpariente, line 132: Re-add those tests please.

asteroid/losses/pit_wrapper.py
```diff
@@ -191,6 +221,169 @@ def best_perm_from_perm_avg_loss(loss_func, est_targets, targets, **kwargs):
...
+        # check input dimensions
+        assert est_targets.shape[0] == targets.shape[0]
+        assert est_targets.shape[2] == targets.shape[2]
+
+        # get L, N and M
+        L = targets.shape[1]        # number of mixtures 
+        N = est_targets.shape[1]    # number of estimated sources
+        if N % L != 0:
+            raise ValueError('The mixtures are assumed to contain the same number of sources')
+        M = N // L                  # number of sources in each mixture
```
mpariente, line 265: We try to avoid variable names like that. 
Using `n_mixtures` or `n_est` would be better.

asteroid/losses/pit_wrapper.py
```diff
@@ -191,6 +221,169 @@ def best_perm_from_perm_avg_loss(loss_func, est_targets, targets, **kwargs):
...
+            all_combinations = []
+            for k in range(len(lst) + 1):
+                for c in combinations(lst, k):
+                    rest = [x for x in lst if x not in c]
+                    all_combinations.append([list(c), rest]) 
+            return all_combinations
+
+        # Generate all the possible partitions
+        parts = all_combinations(range(N))    
+        for p, partition in enumerate(parts):
```
mpariente, line 362: I guess this could be factored out between the two methods?

asteroid/losses/pit_wrapper.py
```diff
@@ -191,6 +221,169 @@ def best_perm_from_perm_avg_loss(loss_func, est_targets, targets, **kwargs):
...
+            assert len(partition) == L
+        
+            # sum the sources according to the given partition
+            est_mixes = torch.stack([torch.sum(est_targets[:, indexes, :], axis=1) for indexes in partition], axis=1)
+
+            # get loss for the given partition
+            if p == 0:
+                loss_set = loss_func(est_mixes, targets, **kwargs)[:, None]
+            else:
+                loss_set = torch.cat([loss_set, loss_func(est_mixes, targets, **kwargs)[:, None]], dim=1)
```
mpariente, line 372: It's more efficient make a list and use `cat`/`stack` than concatenating repeatedly.

mpariente (commented):
This is much better now, thank you! 

There are few things to make before merging it
- Re-add the tests that were deleted,
- Making a separate class,
- Checking with @popcornell for feedback/examples/common code etc.. 
- Factoring out common code between both methods.

Also, it would be amazing to have a notebook to show the mixing matrices (partitions as you call it?)

Would you be ok to do that? 
It would be amazing !

asteroid/losses/pit_wrapper.py
giorgiacantisani: not really because the two functions are different...

giorgiacantisani: If you run this little example the difference will be clearer I guess:

```
# Exmple
n_mix = 2
n_est = 4
n_src = 4 // 2

# Mixit
def combs(lst, k, l):
    if l == 0:
        yield []
    else:
        for c in combinations(lst, k):
            rest = [x for x in lst if x not in c]
            for r in combs(rest, k, l-1):
                yield [list(c), *r]

parts = list(combs(range(n_est), n_src, n_mix))    
print('Mixit') 
for partition in parts:
    print(partition)

# Mixit_gen
def all_combinations(lst):
    all_combinations = []
    for k in range(len(lst) + 1):
        for c in combinations(lst, k):
            rest = [x for x in lst if x not in c]
            all_combinations.append([list(c), rest]) 
    return all_combinations

parts = all_combinations(range(n_est))  
print('Mixit_gen')  
for partition in parts:
    print(partition)
```

mpariente: Thanks for the example. 
I meant, the code under it. 
Yes, the mixing matrices/partitions are different, but the loop seems quite similar right?

giorgiacantisani: yes then the loop it's the same!

popcornell: also the number of sources and number of mixtures are pretty much always known beforehand, maybe we can cache the mixing matrix at beginning.

mpariente: Yes, caching the mixing matrix will be useful!

mpariente:
Let me know if you have any question on what's left to do, or if you need help for any of these tasks, we'll be happy to helps.

giorgiacantisani:
Now there is a new class `MixITLossWrapper` with his own test!

mpariente:
Cool, thanks !
I'll have a look soon.

mpariente:
Hey @giorgiacantisani, I made some changes to factor out the loop and fix the tests so that it passes. 
I checked and the partition generation is negligible compared to the actual loss computation, so I don't think we need to cache it. 
OTOH, I think this could be optimized further 
For example in the `n_mix=3, n_src=3`, there are 1680 partitions, and they look like this 
```python
 [[[0, 1, 2], [3, 4, 5], [6, 7, 8]],
 [[0, 1, 2], [3, 4, 6], [5, 7, 8]],
 [[0, 1, 2], [3, 4, 7], [5, 6, 8]],
 [[0, 1, 2], [3, 4, 8], [5, 6, 7]],
 [[0, 1, 2], [3, 5, 6], [4, 7, 8]],
 [[0, 1, 2], [3, 5, 7], [4, 6, 8]],
 [[0, 1, 2], [3, 5, 8], [4, 6, 7]],
 [[0, 1, 2], [3, 6, 7], [4, 5, 8]],
 [[0, 1, 2], [3, 6, 8], [4, 5, 7]],
 [[0, 1, 2], [3, 7, 8], [4, 5, 6]],
 [[0, 1, 2], [4, 5, 6], [3, 7, 8]],
 [[0, 1, 2], [4, 5, 7], [3, 6, 8]],
 [[0, 1, 2], [4, 5, 8], [3, 6, 7]],
 [[0, 1, 2], [4, 6, 7], [3, 5, 8]],
 [[0, 1, 2], [4, 6, 8], [3, 5, 7]],
 [[0, 1, 2], [4, 7, 8], [3, 5, 6]],
 [[0, 1, 2], [5, 6, 7], [3, 4, 8]],
 [[0, 1, 2], [5, 6, 8], [3, 4, 7]],
 [[0, 1, 2], [5, 7, 8], [3, 4, 6]],
 [[0, 1, 2], [6, 7, 8], [3, 4, 5]],
 [[0, 1, 3], [2, 4, 5], [6, 7, 8]],
...
```
So we compute the sum of ests (0, 1, 2), and the loss compared to mix[:, 0] 20 times, right? 
I don't think I'll spend time optimizing it now but let's keep it in mind :wink: 

@giorgiacantisani @popcornell : I'm waiting for your approval of how it looks now and I'll merge

popcornell:
Can you wait till i test it on WSJ0-2mix (saturday) ?

mpariente:
Ok, sure.

mpariente:
Maybe @etzinis would like to review actually? ^^

giorgiacantisani:
> Hey @giorgiacantisani, I made some changes to factor out the loop and fix the tests so that it passes.
> I checked and the partition generation is negligible compared to the actual loss computation, so I don't think we need to cache it.
> OTOH, I think this could be optimized further
> For example in the `n_mix=3, n_src=3`, there are 1680 partitions, and they look like this
> 
> ```python
>  [[[0, 1, 2], [3, 4, 5], [6, 7, 8]],
>  [[0, 1, 2], [3, 4, 6], [5, 7, 8]],
>  [[0, 1, 2], [3, 4, 7], [5, 6, 8]],
>  [[0, 1, 2], [3, 4, 8], [5, 6, 7]],
>  [[0, 1, 2], [3, 5, 6], [4, 7, 8]],
>  [[0, 1, 2], [3, 5, 7], [4, 6, 8]],
>  [[0, 1, 2], [3, 5, 8], [4, 6, 7]],
>  [[0, 1, 2], [3, 6, 7], [4, 5, 8]],
>  [[0, 1, 2], [3, 6, 8], [4, 5, 7]],
>  [[0, 1, 2], [3, 7, 8], [4, 5, 6]],
>  [[0, 1, 2], [4, 5, 6], [3, 7, 8]],
>  [[0, 1, 2], [4, 5, 7], [3, 6, 8]],
>  [[0, 1, 2], [4, 5, 8], [3, 6, 7]],
>  [[0, 1, 2], [4, 6, 7], [3, 5, 8]],
>  [[0, 1, 2], [4, 6, 8], [3, 5, 7]],
>  [[0, 1, 2], [4, 7, 8], [3, 5, 6]],
>  [[0, 1, 2], [5, 6, 7], [3, 4, 8]],
>  [[0, 1, 2], [5, 6, 8], [3, 4, 7]],
>  [[0, 1, 2], [5, 7, 8], [3, 4, 6]],
>  [[0, 1, 2], [6, 7, 8], [3, 4, 5]],
>  [[0, 1, 3], [2, 4, 5], [6, 7, 8]],
> ...
> ```
> 
> So we compute the sum of ests (0, 1, 2), and the loss compared to mix[:, 0] 20 times, right?
> I don't think I'll spend time optimizing it now but let's keep it in mind 😉
> 

Yes, that's correct.  I didn't even try to optimize it because normally the number of sources and mixes isn't too high...

giorgiacantisani:
But I agree that it would be nice to optimize it! For the rest, all the new commits look good to me :)

etzinis:
Seems good to me btw. However, I do not know why there is all this code complexity going around for generatign the combinations which can also be done in a single line:

for comb_ind in itertools.product(range(num_mixtures), repeat=num_estimated_sources)

E.g. for 2 mixtures and 4 estimated sources
`In [2]: [i for i in itertools.product(range(2), repeat=4)]
Out[2]: 
[(0, 0, 0, 0),
 (0, 0, 0, 1),
 (0, 0, 1, 0),
 (0, 0, 1, 1),
 (0, 1, 0, 0),
 (0, 1, 0, 1),
 (0, 1, 1, 0),
 (0, 1, 1, 1),
 (1, 0, 0, 0),
 (1, 0, 0, 1),
 (1, 0, 1, 0),
 (1, 0, 1, 1),
 (1, 1, 0, 0),
 (1, 1, 0, 1),
 (1, 1, 1, 0),
 (1, 1, 1, 1)]`

etzinis:
Building on top of my previous comment: Each estimated source can only correspond to 1 mixture. So first you generate all possible assignments, compute the loss in each one of them and finally compute the min across all combinations.

mpariente:
Thanks for your feedback!

> Each estimated source can only correspond to 1 mixture. So first you generate all possible assignments, compute the loss in each one of them and finally compute the min across all combinations.

This, I understand. But I don't understand the role of the output you showed in the previous snippet..
Can you please elaborate?

mpariente:
Ok, it's the mixture assignment of the source ^^
I didn't spend much time on this so I'm not sure but you usually consider all possible assignments? Or you assume that mixtures have the same number of sources? 
What is the max num of sources and mixtures you tried MixIT with?

popcornell:
> This, I understand. But I don't understand the role of the output you showed in the previous snippet..
> Can you please elaborate?

I think I understand it, in my implementation i generate the combinations as @etzinis suggests and get the other row by using not ( the mixing matrxi is 2xM where M is number of max sources). Then I use it  as a mixing matrix i simply multiply it by the estimated sources. No fancy indexing but lots of useless multiplications by zeros. Probably a masking approach gets the best of both worlds.

etzinis:
@mpariente Everytime you need to obtain all possible assignments. So if your model outputs M sources then you need to have all possible assignments of estimated sources to mixtures, namely, num_mixtures^M. Btw generating the mixing matrices is exactly the same as generating the index assignments (I would say a bit faster since you need to do it only once in __init__). @popcornell has a nice implementation imho.

mpariente:
@etzinis thanks for the feedback because the current way, without `generalized=True`, it doesn't consider all assignments, but the ones that assign equal number of sources to each mixture (if I understood correctly). 

So maybe, we should switch to computing all the assignements all the time, in a more efficient manner.

mpariente:
@popcornell Cool example, this works for two mixtures, does it generalize well for more? 

I currently don't have a lot of time to put into rewriting this sadly

etzinis:
I would suggest this: In most of the cases you are going to use 2 mixtures, so there is no need to implement it for more (you can always do that later if you need to). Assuming that each mixture has the same number of sources, restrict the model a little bit in finding a good optimum from my experience. However, if the number of sources is quite large, then it would not mean much (you just compute some less combinations).

mpariente:
The current code can compute the loss for any number of mixtures, but fixed number of sources assigned. 
I'm not sure that I'd like to limit the users to two sources only, specially if it's already implemented..

popcornell:
I think what @giorgiacantisani  is really good and can be merged (once tested on a dataset which i ll do). 
Let's keep these observations for optimizing the loss in the future with a second pull request.

mpariente:
I agree that this is very good ! 
I checked again and the current code approaches the problem differently: it produces the indices the other way around, from the mixture to the sources. 

Maybe we should make the `generalized` the default (where possible), WDYT?

popcornell:
In my opinion yes, the current implementation is also speedy enough with generalized=True
I've tested it and it works so I think we can merge. 
I have a recipe ready I can merge it afterwards.

mpariente:
Thanks to @giorgiacantisani and everyone involved! I'll wait for the tests and merge it. 

I made the `generalized` the default because this seems to be the intended behavior from the original authors, hope you're ok with this @giorgiacantisani

mpariente:
@popcornell Can you open the PR for the MixIT recipe :smiley:

giorgiacantisani:
>Thanks to @giorgiacantisani and everyone involved! I'll wait for the tests and merge it.

you're welcome :) thank you for reviewing my code!

> I made the generalized the default because this seems to be the intended behavior from the original authors, hope you're ok with this @giorgiacantisani

It's perfectly fine for me, `generalized` was not the default because I did it in a second moment and not because of a specific reason!

Issue #322, opened 2020-11-10
have trouble about resume from ckpt


SouppuoS:
I try to train DPRNN and DPTNet on wsj0-3mix from a checkpoint. and val_loss look like this (blue line is train from a ckpt). 
I think it should be a curve much smoother.
![snap](https://user-images.githubusercontent.com/11723579/98638931-b11b7300-2364-11eb-93f6-dcc57058616e.png)

Scripts were copied from wham folder. to resume from ckpt I add some code in train.py like this:
```
state_dict = torch.load(os.path.join(exp_dir, ckpt_name))   
system.load_state_dict(state_dict['state_dict'])
...
trainer = pl.Trainer(..., resume_from_checkpoint=os.path.join(exp_dir, ckpt_name))
```
And to train DPRNN & DPTNet on wsj0-3mix, I replace wham dataset & dataloader code in train.py to this:
```
train_loader, val_loader = make_dataloaders(n_src=3, **conf["data"], **conf["training"])
```
no change in conf.yml.
I also tried deepclustering on wsj0-3mix, dprnn & dptnet on wham and my own dataset, did not find this problem.

what did I miss? Need help plz.

mpariente:
Thanks for the detailed issue. 

First guess: this might be related to the optimizer and scheduler. I don't know if the optimizer state is resumed from the checkpoint. Also don't know if the learning rate was halved, does it start from half again? 

I didn't encounter this problem so I'm not sure what's the solution. Can you check if the optimizer's state is loaded? (or even present in the checkpoint?)

jonashaag:
I am seeing this regularly when interrupting and then continuing a training, with all sorts of optimizers. But what I haven't seen so far is that before the interruption training essentially stagnated when compared to what happens to the training after continuation.

Can you post a validation loss plot with less smoothing? Maybe it has something to do with how Tensorboard smoothes.

SouppuoS:
hi, mpariente 
I load the ckpt (generated from DPTNet) and there is nothing in `lr_schedulers`. I dont know if it is right.
![snap2](https://user-images.githubusercontent.com/11723579/98671223-e341ca80-238e-11eb-82df-7c9b6794bac2.jpg)
and after `system.load_state_dict()`, system.scheduler seems did not change, as is shown below. I think the epoch should be 83 (83 can be found in ckpt's state_dict['epoch']) instead of 6. 
[...]

PR #326, opened 2020-11-10
schedulers not saved when they are dict 

popcornell:
Should fix https://github.com/mpariente/asteroid/issues/322. 
The changes are minimal.

### DIFF
--- a/asteroid/engine/schedulers.py
+++ b/asteroid/engine/schedulers.py
@@ -30,7 +30,9 @@ class _BaseScheduler(object):
         self._set_lr(lr)
 
     def load_state_dict(self, state_dict):
-        self.__dict__.update(state_dict)
+
+        self.__dict__.update({k : v for k, v in state_dict.items()
+                              if k != "optimizer"})
 
     def state_dict(self):
         return {key: value for key, value in self.__dict__.items()}
--- a/asteroid/engine/schedulers.py
+++ b/asteroid/engine/schedulers.py
@@ -30,12 +30,10 @@ class _BaseScheduler(object):
         self._set_lr(lr)
 
     def load_state_dict(self, state_dict):
-
-        self.__dict__.update({k : v for k, v in state_dict.items()
-                              if k != "optimizer"})
+        self.__dict__.update({k: v for k, v in state_dict.items()})
 
     def state_dict(self):
-        return {key: value for key, value in self.__dict__.items()}
+        return {key: value for key, value in self.__dict__.items() if key != "optimizer"}
 
     def plot(self, start=0, stop=100_000):  # noqa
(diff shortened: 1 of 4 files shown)


asteroid/engine/schedulers.py
```diff
@@ -30,7 +30,9 @@ def step(self):
         self._set_lr(lr)
 
     def load_state_dict(self, state_dict):
-        self.__dict__.update(state_dict)
+
+        self.__dict__.update({k : v for k, v in state_dict.items()
+                              if k != "optimizer"})
```
mpariente, line 35: I would put that on the `state_dict` rather than `load_state_dict`. I'll do it.

mpariente:
From lightning's docs: 
```python
                # example with step-based learning rate schedulers
                def configure_optimizers(self):
                    gen_opt = Adam(self.model_gen.parameters(), lr=0.01)
                    dis_opt = Adam(self.model_disc.parameters(), lr=0.02)
                    gen_sched = {'scheduler': ExponentialLR(gen_opt, 0.99),
                                 'interval': 'step'}  # called after each training step
                    dis_sched = CosineAnnealing(discriminator_opt, T_max=10) # called every epoch
                    return [gen_opt, dis_opt], [gen_sched, dis_sched]
```
so this is already fully supported in lightning in 0.9.0 ..

asteroid/engine/system.py
```diff
@@ -163,23 +163,23 @@ def validation_epoch_end(self, outputs):
...
+        epoch_schedulers = []
+        for sched in self.scheduler:
+            if not isinstance(sched, dict):
+                epoch_schedulers.append(sched)
+            else:
+                assert sched["interval"] in [
+                    "batch",
+                    "epoch",
+                ], "Scheduler interval should be either batch or epoch"
+                epoch_schedulers.append(sched)
```
mpariente, line 181: Sorry, the diff is not easily readable..
Only thing which was changed: sched is added to `epoch_schedulers` whatever happens (so that pl handles the saving).

popcornell:
If is supported now in lightning, we should use lightning directly IMO. Let me test it.

popcornell:
Ok now instead of batch we use step. Everything is handled in the background by lightning.

mpariente:
Cool, thanks !
I made sure the changes were BC and fixed some docs and tests.

jonashaag:
We should add a test so that we don’t reintroduce the bug

popcornell:
Do we have some dummy training test already ? Because we need to load and unload from a checkpoint the whole trainer I think in order to be sure.

mpariente:
Yes we have dummy sets in test_utils.py

PR #327, opened 2020-11-12
improve: improve Filterbank.forward error message

### LINKED ISSUE #325: Provide better error message for impatient people like me
Because I did not read the documentation before playing with `asteroid`, I tried this:

```python
filters = ParamSincFB(...)
filters(waveforms)
```
instead of 
```python
filters = Encoder(ParamSincFB(...))
filters(waveforms)
```

Therefore, I got a a not-so-nice `NotImplementedError:` exception without any details.

Would you accept a PR that adds a `forward`  in `Filterbank` base class that raises a `NotImplementedError` that tells impatient people like me that they should wrap filter banks in an `Encoder` or a `Decoder`?

hbredin:
Closes #325

### DIFF
--- a/asteroid/filterbanks/enc_dec.py
+++ b/asteroid/filterbanks/enc_dec.py
@@ -47,6 +47,9 @@ class Filterbank(nn.Module):
         }
         return config
 
+    def forward(self, waveform):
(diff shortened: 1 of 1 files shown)


mpariente:
Thanks !

PR #328, opened 2020-11-16
Mixit recipe

popcornell:
Simple MixIT recipe on WHAM clean using always MoMs with 4 speakers.


egs/wham/MixIT/utils/parse_options.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
mpariente, line 1: Please `ln -s ../../wham/ConvTasNet/utils .` instead.

popcornell: oops sorry

egs/wham/MixIT/train.py
```diff
@@ -0,0 +1,180 @@
...
+        moms = mix1 + mix2
+
+        est_sources = self(moms)
+        loss = self.loss_func["mixit"](est_sources, torch.stack((mix1, mix2)).transpose(0, 1))
+
+        with torch.no_grad():
+            # we cheat during training and check oracle performance ^^
+            loss_oracle = self.loss_func["pit"](
+                est_sources, torch.cat((oracle[bsz // 2 :], oracle[: bsz // 2]), 1)
+            )
```
mpariente, line 43: Actually, `MixITWrapper` could be used as `PITLossWrapper` when the number of sources is the same as the number of mixtures. Maybe we should make that, it would be a bit cleaner.

popcornell: Also on validation then. But I guess it could be also confusing for someone. That is why we choose to have two separate classes even if MixIT is a generalization of PIT

popcornell:
One thing could be confusing is having this in WHAM. We use WHAM clean and MixIT works only on WHAM clean.
Because WHAM clean is same as WSJ0-2mix we might want to move this there.

egs/wham/MixIT/train.py
mpariente: Haha, yes. I don't know then..

mpariente:
> One thing could be confusing is having this in WHAM. We use WHAM clean and MixIT works only on WHAM clean.
> Because WHAM clean is same as WSJ0-2mix we might want to move this there.

Didn't think about that.
We can hard code the task in the train file.. And the README is pretty clear so I think it's fine.

mpariente:
It's very frustrating to copy paste all this content for a "simple" MixIT example though.

popcornell:
I agree, I can use symlink maybe for the whole local folder. 
But conf.yml has to be different so it has to be moved somewhere else. Like in a conf directory as Kaldi does.

mpariente (approved):
Looks great, thanks !

egs/wham/MixIT/eval.py
```diff
@@ -0,0 +1,144 @@
...
+    series_list = []
+    torch.no_grad().__enter__()
+    for idx in tqdm(range(len(test_set))):
+        # Forward the network on the mixture.
+        mix, sources = tensors_to_device(test_set[idx], device=model_device)
+        est_sources = model(mix[None, None])
+        _, indxs = torch.sort(torch.sqrt(torch.mean(est_sources ** 2, dim=-1)), descending=True)
+        indxs = indxs[:, :2]
+
+        est_sources = est_sources.gather(1, indxs.unsqueeze(-1).repeat(1, 1, est_sources.shape[-1]))
```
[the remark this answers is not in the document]
popcornell: yeah you are right

egs/wham/MixIT/eval.py
```diff
@@ -41,7 +41,7 @@
 
 def main(conf):
     model_path = os.path.join(conf["exp_dir"], "best_model.pth")
-    model = ConvTasNet.from_pretrained(model_path)
+    model = DPRNNTasNet.from_pretrained(model_path)
```
popcornell, line 44: thank you i totally forgot about this !

popcornell:
Probably can be improved a lot by using a custom dataloader and do the "magic" there by creating MoMs with not always 2 speakers. But this recipe it is just a proof of concept and should be enough to show MIxIT.

mpariente:
> But this recipe it is just a proof of concept and should be enough to show MIxIT.

Completely agree. This is enough IMO

I'm waiting for the comment in the eval file (and the pl module maybe) and I'll merge.

popcornell:
Added comments on both eval.py and validation_step

PR #329, opened 2020-11-16
ESPNet integration: v0

mpariente:
So that's a first iteration for ASR evaluation in the LibriMix recipe. 

Thanks to @JorisCos  for some of the ground work. 

I'll comment in the PR about things I'm unsure about and where I'd like help hopefully.
None of the code has been ran yet, things will likely change place but this is just a rough draft to see how easy it would be to do it. 

@raikarsagar @JorisCos @popcornell @mhu-coder

### DIFF
--- a/egs/librimix/ConvTasNet/eval.py
+++ b/egs/librimix/ConvTasNet/eval.py
@@ -121,7 +121,7 @@ def main(conf):
             **wer_tracker(
                 mix=mix_np,
                 clean=sources_np,
-                est_sources=est_sources_np,
+                estimate=est_sources_np,
                 wav_id=ids,
                 sample_rate=conf["sample_rate"],
             )
@@ -162,9 +162,15 @@ def main(conf):
     print("Overall metrics :")
     pprint(final_results)
     if conf["compute_wer"]:
+        # Print WER report
         print("\nWER report")
-        wer_card = wer_tracker.final_report()
-        pprint(wer_card)
+        wer_card = wer_tracker.final_report().to_markdown(index=False,
+                                                          tablefmt='github')
+        print(wer_card)
+
+        # Save the report
+        with open(os.path.join(eval_save_dir, "final_wer.md"), "w") as f :
+            f.write(wer_card)
 
     with open(os.path.join(eval_save_dir, "final_metrics.json"), "w") as f:
         json.dump(final_results, f, indent=0)
--- a/egs/librimix/ConvTasNet/eval.py
+++ b/egs/librimix/ConvTasNet/eval.py
@@ -180,6 +180,32 @@ class WERTracker:
     def _df_to_dict(self, df):
         return {k: v for k, v in zip(df["utt_id"].to_list(), df["text"].to_list())}
 
+    def final_report(self):
+        """Generate a MarkDown table, as done by ESPNet."""
+        mix_n_word = sum(self.mix_counter[k] for k in ["hits", "substitutions", "deletions"])
+        est_n_word = sum(self.est_counter[k] for k in ["hits", "substitutions", "deletions"])
+        mix_wer = self.wer_from_hsdi(**dict(self.mix_counter))
+        est_wer = self.wer_from_hsdi(**dict(self.est_counter))
+
+        mix_hsdi = [
+            self.mix_counter[k] for k in ["hits", "substitutions", "deletions", "insertions"]
+        ]
+        est_hsdi = [
+            self.est_counter[k] for k in ["hits", "substitutions", "deletions", "insertions"]
+        ]
+        #                   Snt               Wrd         HSDI       Err     S.Err
+        for_mix = [len(self.mix_counter), mix_n_word] + mix_hsdi + [mix_wer, "-"]
+        for_est = [len(self.est_counter), est_n_word] + est_hsdi + [est_wer, "-"]
+
+        line_list = [
+            "| dataset | Snt | Wrd | Corr | Sub | Del | Ins | Err | S.Err |",
+            "| --- | --- | --- | --- | --- | --- | --- | --- | --- |"
+            f"| decode_asr_lm / test_clean / mixture   |" + " | ".join(map(str, for_mix)) + "|",
+            f"| decode_asr_lm / test_clean / separated |" + " | ".join(map(str, for_est)) + "|",
+        ]
+        result_card = "\n".join(line_list)
+        return result_card
+
 
 def main(conf):
     compute_metrics = update_compute_metrics(conf["compute_wer"], COMPUTE_METRICS)
@@ -262,13 +288,21 @@ def main(conf):
 
     # Print and save summary metrics
     final_results = {}
+    if conf["compute_wer"]:
+        compute_metrics.append("wer")  # Just for the final .json.
     for metric_name in compute_metrics:
         input_metric_name = "input_" + metric_name
         ldf = all_metrics_df[metric_name] - all_metrics_df[input_metric_name]
         final_results[metric_name] = all_metrics_df[metric_name].mean()
         final_results[metric_name + "_imp"] = ldf.mean()
+
     print("Overall metrics :")
     pprint(final_results)
+    if conf["compute_wer"]:
+        print("\nWER report")
+        wer_card = wer_tracker.final_report()
+        pprint(wer_card)
+
     with open(os.path.join(eval_save_dir, "final_metrics.json"), "w") as f:
         json.dump(final_results, f, indent=0)
 
--- a/egs/librimix/ConvTasNet/eval.py
+++ b/egs/librimix/ConvTasNet/eval.py
@@ -1,5 +1,6 @@
 import os
 import random
+from typing import List
 import soundfile as sf
 import torch
 import yaml
@@ -17,6 +18,17 @@ from asteroid import ConvTasNet
 from asteroid.models import save_publishable
 from asteroid.utils import tensors_to_device
 
+
+def import_wer():
+    try:
+        from jiwer import wer
+    except ModuleNotFoundError:
+        return None
+    else:
+        return wer
+
+
+wer = import_wer()
 
 parser = argparse.ArgumentParser()
 parser.add_argument(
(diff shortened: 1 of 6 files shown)


egs/librimix/ConvTasNet/eval.py
```diff
@@ -18,6 +19,17 @@
 from asteroid.utils import tensors_to_device
 
 
+def import_wer():
```
mpariente, line 22: This function can live somewhere else.

egs/librimix/ConvTasNet/eval.py
```diff
@@ -41,11 +53,101 @@
 parser.add_argument(
     "--n_save_ex", type=int, default=10, help="Number of audio examples to save, -1 means all"
 )
+parser.add_argument(
+    "--compute-wer", type=int, default=0, help="Compute WER using ESPNet's pretrained model"
+)
+
+COMPUTE_METRICS = ["si_sdr", "sdr", "sir", "sar", "stoi"]
+ASR_MODEL_PATH = (
```
mpariente, line 61: Will need to be a variable probably.

egs/librimix/ConvTasNet/eval.py
```diff
@@ -41,11 +53,101 @@
...
+        from espnet_model_zoo.downloader import ModelDownloader
+    except ModuleNotFoundError:
+        import warnings
+
+        warnings.warn("Couldn't find espnet installation. Continuing without.")
+        return metric_list
+    return metric_list + ["wer"]
+
+
+class MockTracker:
```
mpariente, line 80: Will live somewhere else (follows `WERTracker`). It's to enable the WER/no WER without too many if and code changes.

egs/librimix/ConvTasNet/eval.py
```diff
@@ -41,11 +53,101 @@
...
+    def predict_hypothesis(self, wav):
+        nbests = self.asr_model(wav)
+        text, *_ = nbests[0]
+        return text
+
+    def compute_wer(self, transcriptions):
+        # Is average WER the average over individual WER?
+        # Or computed overall with all S, D, I?
+        # Is it different than the average from call (probably)
+        return
```
mpariente, line 132: Would love an answer to that

egs/librimix/ConvTasNet/eval.py
```diff
@@ -41,11 +53,101 @@
...
+        if sample_rate != self.sample_rate:
+            mix, est_sources = self.resample(
+                mix, est_sources, fs_from=sample_rate, fs_to=self.sample_rate
+            )
+        input_wer = output_wer = 0.0
+        # Count the mixture output for each speaker
+        txt = self.predict_hypothesis(mix)
+        for tmp_id in wav_id:
+            input_wer += wer(truth=self.trans_dic[tmp_id], hypothesis=txt) / len(wav_id)
+            self.input_txt_list.append(dict(utt_id=tmp_id, text=txt))
```
mpariente, line 115: Compute the WER from mixture's hypothesis to each ground truth. Might not be ideal. IDK how it's done in ESPNet.

egs/librimix/ConvTasNet/eval.py
```diff
@@ -58,6 +160,8 @@ def main(conf):
         sample_rate=conf["sample_rate"],
         n_src=conf["train_conf"]["data"]["n_src"],
         segment=None,
+        return_id=True,
+        # FIXME: ensure max mode for eval.
```
mpariente, line 164: Related to another FIXME in the run.sh

egs/librimix/ConvTasNet/local/get_text.py
```diff
@@ -0,0 +1,28 @@
...
+    f = open(name, "r")
+    for line in f:
+        dict1 = {}
+        split_line = line.split(" ", maxsplit=1)
+        dict1["utt_id"] = split_line[0]
+        dict1["text"] = split_line[1].replace("\n", "").replace("\r", "")
+        row_list.append(dict1)
+
+df = pd.DataFrame(row_list)
+df.to_csv(os.path.join(args.outdir, "annotations.csv"), index=False)
```
mpariente, line 28: Might want to change the name to match ESPNet's conventions.

egs/librimix/ConvTasNet/local/prepare_data.sh
```diff
@@ -17,3 +17,9 @@ cd LibriMix
 cd $current_dir
 $python_path local/create_local_metadata.py --librimix_dir $storage_dir/Libri$n_src"Mix"
 
+
+# TODO: check folders and outdir
+#dev-clean  test-clean  train-clean-100  train-clean-360
+for split in train-360 train-100 dev-clean test-clean; do
+  $python_path local/get_text.py --libridir $storage_dir/LibriSpeech --split $split --outdir data/$split
+done
```
mpariente, line 25: The output directories don't have the `clean` in the name.. We might either want to call this 4 times without the loop, or loop over directories in the .py file.

egs/librimix/ConvTasNet/local/prepare_data.sh
```diff
@@ -17,3 +17,9 @@ cd LibriMix
 cd $current_dir
 $python_path local/create_local_metadata.py --librimix_dir $storage_dir/Libri$n_src"Mix"
 
+
+# TODO: check folders and outdir
```
mpariente, line 21: Here, we retrieve the txt files for each split and gather them in .csv files that we'll store in the local `data` folder.

egs/librimix/ConvTasNet/run.sh
```diff
@@ -40,6 +40,7 @@ optimizer=adam
 lr=0.001
 weight_decay=0.
 # Data config
+# FIXME: this is wrong to have it written in plain
 train_dir=data/wav8k/min/train-360
 valid_dir=data/wav8k/min/dev
 test_dir=data/wav8k/min/test
```
mpariente, line 46: These path should depend on the sample rate, not be hardcoded. From WHAM's recipes!
```
sr_string=$(($sample_rate/1000))
suffix=wav${sr_string}k/$mode
```

egs/librimix/ConvTasNet/eval.py
mpariente, line 164: New problem: we need the "max" version for ASR, but the other metrics should still be computed with the "min" mode so that the results are consistent with what we had before. 
We'll probably need to change the dataloader to return the lengths, other solutions?

mpariente (commented):
There are few things left to do for this to be acceptable, in a first step. 

@raikarsagar, would you like to help? That would really be amazing !! :smiley:

egs/librimix/ConvTasNet/eval.py
```diff
@@ -18,6 +19,17 @@
 from asteroid.utils import tensors_to_device
 
 
+def import_wer():
+    try:
+        from jiwer import wer
```
mpariente, line 24: Since you're here @sw005320, what do you think about `jiwer`?
BTW, feel free to add any comment, advice etc..

egs/librimix/ConvTasNet/eval.py
popcornell: it should be specified by the user according to the recipe

egs/librimix/ConvTasNet/eval.py
popcornell: https://github.com/kaldi-asr/kaldi/blob/85a3dd5f0b71e419abf1169a26b759bfc423a543/src/bin/compute-wer.cc#L94
This is the way to go: compute it over all the dataset (e.g. test set) by summing the errors of all utterances then normalize by the sum of n_words. If you compute it by utterance and then take average of WERs say you have on word in one sequence and you miss that you have there 100% WER there. In another you have 20 words and you get them all correct. If you take your average you get 50% which is not really reperesentative of the performance at this point.

egs/librimix/ConvTasNet/eval.py
popcornell: we should accumulate n_words, sub del and ins for each utterance. 
Remember we need to handle the computation for those for both the speakers ( or N speakers) in case we do not perform speaker adaptation.

egs/librimix/ConvTasNet/eval.py
popcornell: I would use max version all along. Otherwise you would not be able to say plot SI-SDR vs WER and compare the two directly.

egs/librimix/ConvTasNet/eval.py
mpariente: This is what I thought, thanks !
How expensive is the computation of WER? At first, I had planned to accumulate all the hypothesis/transcripts and compute the overall WER at the end. I still log all of it so we can compute the average WER in the end (supported by jiwer, which accounts for total number of words, as we want). 

BTW, even if less critical, we do this "mistake" on SI-SDR, 3s examples and 12s examples have the same weight on the averaged metric.

egs/librimix/ConvTasNet/eval.py
mpariente: I see the point of doing that but it seems out of scope of this PR maybe. 
For ins, the word order doesn't matter but for del and sub, it does so that's tricky.

egs/librimix/ConvTasNet/eval.py
mpariente: Yes, but SISDR results are usually reported on min, we cannot change that without the user knowing it. Otherwise recipes inspired by this might report wrong results afterward..

egs/librimix/ConvTasNet/eval.py
mpariente: Yes, and we'll have to make sure it was trained on LibriSpeech.

egs/librimix/ConvTasNet/eval.py
sw005320: I checked the library and it looks good to me.
I was concerning punctuation handling, upper/lower case, and non-linguistic special annotations (e.g., <NOISE>), but the tool seems to handle everything.

egs/librimix/ConvTasNet/eval.py
sw005320: > This is what I thought, thanks !
> How expensive is the computation of WER? At first, I had planned to accumulate all the hypothesis/transcripts and compute the overall WER at the end. I still log all of it so we can compute the average WER in the end (supported by jiwer, which accounts for total number of words, as we want).

It depends on the internal algorithm (o(N) or (o(N^2)) and I don't know the details of this library.
Some toolkit takes a very long time for the WER computation of the long sequence.
I think your idea would be fine.

Some people may want to know the breakdown numbers per utterance or per speaker, but this would be too much for now.
This is one reason that we still use NIST sclite, which provides such analysis information, but it is not a pythonic solution.
(Another reason for using NIST toolkit is that I want to stick to use a standardized toolkit for the evaluation metric. It is a disaster if the scoring toolkit includes some errors or some dialects.).

Also, I recommend you generate the number of total word, insertion, deletion, and substitution counts, as well as WER.
This is good debugging information.
I often encountered the issue that people don't use the correct test set.
If we display the number of sentences and words, we can easily detect such mistakes.
Also, if we display insertion, deletion, and substitution counts, we can easily detect what issue causes the error.
Speech enhancement or separation tends to increase deletion
(if the speech is over suppressed) and the insertion (if there are interference speakers).

egs/librimix/ConvTasNet/eval.py
JorisCos: I assumed that ASR models couldn't be used to perform cross datasets evaluation but is this true ?

egs/librimix/ConvTasNet/eval.py
mpariente: Thank you very much for your feedback!
We'll log the numbers (ins, del, sub, tot, WER) per utterance then, as is done for the other metrics.

egs/librimix/ConvTasNet/eval.py
mpariente: Again, thank a lot ! It's going to make it much easier to use jiwer directly rather than a non-python alternative for us.

egs/librimix/ConvTasNet/eval.py
mpariente: There is the problem of vocab at least. If the vocab of the second dataset is included in the first one, then that's probably possible. Samu can probably answer better than me.

egs/librimix/ConvTasNet/eval.py
mpariente: For reference: 
Opened a PR in jiwer to access those numbers: https://github.com/jitsi/jiwer/pull/35

mpariente:
The PR was merged in `jiwer` and hits, sub, ins, del are exposed in v2.2.0

This shouldn't be too much work from now. @raikarsagar, could you give it a try please?

sw005320:
FYI, this is a WER report format we often use in espnet.
If asteroid produces such information, people in espnet can easily catch up the result.
Our script automatically generates the following markdown table after the ASR evaluation stage.

|dataset|Snt|Wrd|Corr|Sub|Del|Ins|Err|S.Err|
|---|---|---|---|---|---|---|---|---|
|decode_asr_lm_lm_train_lm_adam_bpe_valid.loss.ave_asr_model_valid.acc.ave/dev_clean|2703|54402|98.0|1.8|0.2|0.3|2.3|27.6|
|decode_asr_lm_lm_train_lm_adam_bpe_valid.loss.ave_asr_model_valid.acc.ave/dev_other|2864|50948|95.0|4.3|0.6|0.5|5.5|45.1|
|decode_asr_lm_lm_train_lm_adam_bpe_valid.loss.ave_asr_model_valid.acc.ave/test_clean|2620|52576|97.9|1.9|0.2|0.3|2.4|28.2|
|decode_asr_lm_lm_train_lm_adam_bpe_valid.loss.ave_asr_model_valid.acc.ave/test_other|2939|52343|94.9|4.5|0.7|0.6|5.8|48.6|

mpariente:
We'll adopt this convention as well, thanks. 

Does "S.Err" mean the percent of sentences without perfect score?

sw005320:
> Does "S.Err" mean the percent of sentences without perfect score?

S.Err means the sentence error rate.
If there is any word/char/token error in the sentence, it would become a sentence error.
This is very strict but some applications require such metrics.

egs/librimix/ConvTasNet/local/get_text.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--libridir", required=True, type=str)
+parser.add_argument("--outfile", required=True, type=str)
+parser.add_argument("--split", type=str, default="train-360")
+
+args = parser.parse_args()
+
+libridir = f"LibriSpeech/{args.split}"
```
JorisCos, line 14: `libridir = os.path.join(args.libridir, args.split)`

mpariente: Feel free to push to this branch. I won't touch it for a small while

egs/librimix/ConvTasNet/eval.py
```diff
@@ -41,11 +68,124 @@
 parser.add_argument(
     "--n_save_ex", type=int, default=10, help="Number of audio examples to save, -1 means all"
 )
+parser.add_argument(
+    "--compute-wer", type=int, default=0, help="Compute WER using ESPNet's pretrained model"
```
JorisCos, line 72: `"--compute-wer"` --> `"--compute_wer"`

egs/librimix/ConvTasNet/run.sh
```diff
@@ -95,10 +110,30 @@ if [[ $stage -le 1 ]]; then
...
+	    echo "Cannot compute WER without max mode. Start again with --stage 2 --compute_wer 1 --eval_mode max"
+	    exit 1
+	  fi
+    # Install espnet if not instaled
+    if not python -c "import espnet" &> /dev/null; then
+        echo 'This recipe requires espnet. Installing requirements.'
+        $python_path -m pip install espnet_model_zoo
+        $python_path -m pip install jiwer
+    fi
+  fi
```
JorisCos, line 128: I never go through this loop even when i don't have espnet installed . Not sure why.

mhu-coder: have you tried `if ! python -c "import espnet" &> /dev/null`? I don't think `not` is a valid bash keyword

egs/librimix/ConvTasNet/eval.py
```diff
@@ -154,9 +153,12 @@ def wer_from_hsdi(hits, substitutions, deletions, insertions):
 
     @staticmethod
     def hsdi(truth, hypothesis):
-        keep = ["hits", "substitutions", "deletions", "insertions"]
+        keep = {"hits": 0, "substitutions": 0, "deletions": 0, "insertions": 0}
         out = compute_measures(truth=truth, hypothesis=hypothesis).items()
-        return {k: v for k, v in out if k in keep}
```
mpariente: Didn't this work? What was the error message?

JorisCos: Yes it did ! My bad. The problems comes from `Counter()`.  That removes keys if they are non positive. I am gonna find a workaround ;) 
```python
>>> a = Counter({'menu': 20, 'good': 15, 'happy': 10, 'bar': 0})
>>> b = Counter()
>>> a+b
Counter({'menu': 20, 'good': 15, 'happy': 10})
```

JorisCos: Done

popcornell:
It seems that right now `espnet2.bin.asr_inference.Speech2Text`  does not supported batched audio: 
see https://github.com/espnet/espnet/blob/554e65ee638ebdc8256647ce273cf6303ea72e39/espnet2/bin/asr_inference.py#L180

Because of this WER computation is quite slow even on GPU (but so is SDR also right now with mir_eval).
Can you confirm @JorisCos I have in mind couple of workarounds in this case.

JorisCos:
> It seems that right now `espnet2.bin.asr_inference.Speech2Text` does not supported batched audio:
> see https://github.com/espnet/espnet/blob/554e65ee638ebdc8256647ce273cf6303ea72e39/espnet2/bin/asr_inference.py#L180
> 
> Because of this WER computation is quite slow even on GPU (but so is SDR also right now with mir_eval).
> Can you confirm @JorisCos I have in mind couple of workarounds in this case.

I haven't tried yet, but I wonder if we really need a workaround since we are only using `espnet2.bin.asr_inference.Speech2Text` in eval and evaluation isn't done on batched audio or am I missing something ?

mpariente:
We could have sped things up by using batch processing, because it seems slow otherwise.

egs/librimix/ConvTasNet/eval.py
```diff
@@ -18,6 +21,30 @@
...
+        #  {
+        #     "wer": wer,
+        #     "mer": mer,
+        #     "wil": wil,
+        #     "wip": wip,
+        #     "hits": H,
+        #     "substitutions": S,
+        #     "deletions": D,
+        #     "insertions": I,
+        # }
```
mpariente, line 39:
```suggestion
```

PR #331, opened 2020-11-17
Replace DPRNNTasNet with DPTNet

### LINKED ISSUE #330: Why egs/wham/DPTNet/eval.py and egs/wham/DPRNN/eval.py is the same ?
#### My Experiments:
I trained the DPTNet model with my own data and tried to evaluate on my testing data.
The result (SI-SDR) is : {"si_sdr": 7.398276868387594, "si_sdr_imp": 7.404194952754807}
Then I modified the code and wanted to restore the DPTNet model to continue training.
And I found that egs/wham/DPTNet/eval.py restored the model with DPRNNTasNet function.

#### My Questions:

1. Why to use DPRNNTasNet function to restore the DPTNet model, and it still working ?
2. I modified the code but got the same result: {"si_sdr": 7.398276868387594, "si_sdr_imp": 7.404194952754807}. Is it correct ?
> ```
> # model = DPRNNTasNet.from_pretrained(model_path)
> model = DPTNet.from_pretrained(model_path)
> ```

Thanks for your reply. :)

guiruli08650129:
Fix #330
:)

### DIFF
--- a/egs/wham/DPTNet/eval.py
+++ b/egs/wham/DPTNet/eval.py
@@ -9,7 +9,7 @@ import pandas as pd
 from tqdm import tqdm
 from pprint import pprint
 
-from asteroid import DPRNNTasNet
+from asteroid import DPTNet
 from asteroid.metrics import get_metrics
 from asteroid.losses import PITLossWrapper, pairwise_neg_sisdr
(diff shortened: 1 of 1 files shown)


mpariente:
Thanks !

PR #333, opened 2020-11-18
[src] New phase features and transforms

mpariente:
This PR adds
- Delta and double Delta computation of torch tensors
- A common phase correction that can be useful. 

### Questions:
Should we add `instantaneous_frequency` and `group_delay` functions? How about `nn.Module`? All the transforms are functional but having them as module can be useful.

### DIFF
--- a/asteroid/filterbanks/transforms.py
+++ b/asteroid/filterbanks/transforms.py
@@ -425,3 +425,54 @@ def concat_deltas(feats: torch.Tensor, order: int = 1, dim: int = -1) -> torch.T
     for _ in range(order):
         all_feats.append(compute_delta(feats[-1], dim=dim))
     return torch.cat(all_feats, dim=dim)
+
+
+def centerfreq_correction(
+    spec: torch.Tensor,
+    kernel_size: int,
+    stride: int = None,
+    dim: int = -2,
+) -> torch.Tensor:
+    """Corrects phase from the input spectrogram so that a sinusoid in the
+    middle of a bin keeps the same phase from one frame to the next.
+
+    Args:
+        spec: Spectrogram tensor of shape (batch, n_freq + 2, frames).
+        kernel_size (int): Kernel size of the STFT.
+        stride (int): Stride of the STFT.
+        dim (int): Only works of dim=-2.
+
+    Returns:
+        Tensor: the input spec with corrected phase.
+    """
+    if dim != -2:
+        raise NotImplementedError
+    if stride is None:
+        stride = kernel_size // 2
+    # Phase will be (batch, n_freq // 2 + 1, frames)
+    mag, phase = to_polar(spec, dim=dim)
+    new_phase = phase_centerfreq_correction(phase, kernel_size=kernel_size, stride=stride)
+    new_spec = from_polar(mag, new_phase, dim=dim)
+    return new_spec
+
+
+def phase_centerfreq_correction(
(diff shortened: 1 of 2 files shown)


asteroid/filterbanks/transforms.py
```diff
@@ -371,17 +385,104 @@ def ebased_vad(mag_spec, th_db: int = 40):
...
+        >>> # Compute instantaneous frequency
+        >>> inst_freq = compute_delta(phase, dim=-1)
+        >>> # Or group delay
+        >>> group_delay = compute_delta(phase, dim=-2)
+    """
+    if dim != -1:
+        return compute_delta(feats.transpose(-1, dim), dim=-1).transpose(-1, dim)
+    # First frame has nothing. Then each frame is the diff with the previous one.
+    delta = feats.new_zeros(feats.shape)
+    delta[..., 1:] = feats[..., 1:] - feats[..., :-1]
```
jonashaag, line 410: Wild speculation; could this be more efficient using `F.pad` instead?

asteroid/filterbanks/transforms.py
```diff
@@ -342,6 +345,17 @@ def from_mag_and_phase(mag, phase, dim: int = -2):
     return torch.cat([mag * torch.cos(phase), mag * torch.sin(phase)], dim=dim)
 
 
+# Alias
+from_polar = from_mag_and_phase
```
jonashaag, line 349: We should also add these aliases in `complex_nn` IMO or stick to any of the two names everywhere.

asteroid/filterbanks/transforms.py
mpariente: Maybe. Doesn't seem important to me at this point but you can have a go :wink:

asteroid/filterbanks/transforms.py
```diff
@@ -342,6 +345,17 @@ def from_mag_and_phase(mag, phase, dim: int = -2):
...
+
+
+def to_polar(spec: torch.Tensor, dim: int = -2) -> Tuple[torch.Tensor, torch.Tensor]:
```
popcornell, line 352: much needed IMO thanks Manu

asteroid/filterbanks/transforms.py
```diff
@@ -371,17 +385,104 @@ def ebased_vad(mag_spec, th_db: int = 40):
     return log_mag > (max_log_mag - th_db)
```
popcornell, line 385: maybe for VAD we should add some simple hangover and hangbefore (easy way is by using median filtering)

mpariente: Would you like to do it? 
Maybe we could simply rely on torchaudio's VAD? I didn't check it but it's probably better than ours, and scriptable.

asteroid/filterbanks/transforms.py
jonashaag: Hm ok this is harder than I thought 😬 so want to have 4 things:

- “good” (meaningful) names
- consistent with torchaudio 
- consistent within Asteroid
- compatible with current Asteroid names

We’d probably have to put a non trivial amount of effort into getting all of these right...

I don’t know what to do really.

mpariente: See #335

mpariente:
I think I'm going to make the changes I proposed in #335 after this PR is merged.

PR #334, opened 2020-11-18
[src] New Filterbank's hooks + MelGram_FB

mpariente:
This PR adds two things: 
- Post encoding and pre decoding hooks of the Filtebank to the Encoder and Decoder. 
- A filterbank instance for mel magnitude spectrogram, using the hooks. 

I'd be happy to have feedback on the hooks @jonashaag @popcornell . Until now, I don't see the downside of this. The encoder and decoder are already completely tied to the Filterbank class (it's not it's not like they could be used anywhere else without Filterbank) so we don't remove any feature but add new, without additional constraint. 

I could also add hooks on the waveform, but I don't see applications for it right now, we often switch to TF representation because we can manipulate them easier. 

In particular, I'm thinking about @Baldwin-disso and @faroit because those hooks could make it much easier to match torch's STFT with Asteroid

Last upside of this, we can easily benefit from other filterbank matrices from `librosa` (e.g chroma, constant_q, semitone_filterbank) for free.


asteroid/filterbanks/melgram_fb.py
```diff
@@ -0,0 +1,61 @@
...
+        self.register_buffer("fb_mat", torch.from_numpy(fb_mat).unsqueeze(0))
+        self.n_feats_out = n_mels
+
+    def post_encode(self, spec: torch.Tensor):
+        mag_spec = transforms.take_mag(spec, dim=-2)
+        mel_spec = torch.matmul(self.fb_mat, mag_spec)
+        return mel_spec
+
+    def get_config(self):
+        conf = dict(fmin=self.fmin, fmax=self.fmax, n_mels=self.n_mels)
```
mhu-coder, line 60: would it be clearer if the info about the Mel Filterbank were gathered under the same keyword?
```python
return {**super().get_config(), 'mel_filterbank': conf}
```
Not sure if that's compatible with the way models are loaded though.

asteroid/filterbanks/melgram_fb.py
```diff
@@ -0,0 +1,61 @@
...
+            ``np.sqrt(np.hanning())``.
+        sample_rate (float): Sample rate of the expected audio.
+            Defaults to 8000.
+        n_mels (int): Number of mel bands.
+        fmin (float): Minimum frequency of the mel filters.
+        fmax (float): Maximum frequency of the mel filters.
+        **kwargs:
+    """
+
+    def __init__(
```
mhu-coder, line 25: Are you sure about the default of 128 for `n_mels`?
With a sample rate of 8kHz, we could imagine a use case of having `kernel_size` of 20 * 8 = 160, which would result in frequency bins that are 25Hz apart. With 128 mel bands, we would have the first few centres at ~0Hz, 10.5Hz and 21Hz (not sure about the values used by librosa as my current internet connection won't allow me to download the package). In other words, there may be some mel filterbanks that span the same frequency bins.

jonashaag:
LGTM but I wonder if we should call the hooks differently. Isn’t the Mel transform part of the enocding, and thus not “post” encoding? Maybe a more accurate (albeit very technical) name would be `post_conv`?

asteroid/filterbanks/melgram_fb.py
mpariente: It isn't compatible indeed.

asteroid/filterbanks/melgram_fb.py
```diff
@@ -0,0 +1,61 @@
...
+        self.n_mels = n_mels
+        super().__init__(
+            n_filters=n_filters,
+            kernel_size=kernel_size,
+            stride=stride,
+            window=window,
+            sample_rate=sample_rate,
+            **kwargs,
+        )
+        fb_mat = mel(sr=sample_rate, n_fft=n_filters, fmin=fmin, fmax=fmax, n_mels=n_mels)
```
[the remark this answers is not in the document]
mpariente: Yes, we can expose all the argument of `mel` in `MelGramFB`. If the normalization is important, I'll add it.

mpariente:
Thanks for the feedback!

True that it is part of the encoder. But `post_conv` doesn't specify if it is in the encoder or decoder (which also has a conv). I'd be ok with `post_encoder_conv` and `pre_decoder_conv`, this is accurate and precise, WDYT?

asteroid/filterbanks/melgram_fb.py
mhu-coder: Good question. It's a bit tricky as it depends on the frequency resolution, which depends on `kernel_size`. In this context, the mel magnitude spectrum will be used for separation, which, I guess, may be easier it there are more frequency bands? So I guess 64 should do (and be ok for short frames as well).

popcornell: 40 mels is pretty much the standard value for fbanks for ASR and diarization (x-vectors)

popcornell:
I guess it is fine, another option would be `post_analysis` and `pre_synthesis` in this way it is clear you are applying a (possible non-linear) transformation after the analysis filterbank and before synthesis filterbank. Almost all feats extraction can fall into this framework.

popcornell:
I see only a possible, minor downside regarding current interface: 
what if i need both STFT and Mels for example ? With current interface I am forced to compute the STFT two times. 
A simple option could be getting even more modular and expose the Mel transformation in its own module (and same for other transforms) so that one can do this:

```python 
stft = stft_enc(signal)
mels = mel_tf(stft)
# compute idk loss on both mag stft and mel domain
```

mpariente:
+1 for `post_analysis` and `pre_synthesis`. 

Yes, I didn't think about this case, that's a good point. 
For example, torchaudio exposes `MelScale` which contains the filterbank matrix, and `MelSpectrogram` which uses `Spectrogram` and `MelScale`. We could do something similar and build modules that only apply the matrix in the forward, and use them in other modules as `post_analysis` filters.  Not sure if it's clear, but I can show an example if not.

jonashaag:
> Not sure if it's clear, but I can show an example if not.

Not clear 😅

mpariente:
I guess this is what Samu had in mind, and it makes complete sense to me
```python
class MelScale(torch.nn.Module):
    def __init__(
        self, n_filters, sample_rate=8000.0, n_mels=128, fmin=0.0, fmax=None,
    ):
        from librosa.filters import mel

        super().__init__()
        fb_mat = mel(sr=sample_rate, n_fft=n_filters, fmin=fmin, fmax=fmax, n_mels=n_mels)
        self.register_buffer("fb_mat", torch.from_numpy(fb_mat).unsqueeze(0))

    def forward(self, spec: torch.Tensor):
        mag_spec = transforms.take_mag(spec, dim=-2)
        mel_spec = torch.matmul(self.fb_mat, mag_spec)
        return mel_spec


class MelGramFB(STFTFB):
    def __init__(
        self,
        n_filters,
        kernel_size,
        stride=None,
        window=None,
        sample_rate=8000.0,
        n_mels=128,
        fmin=0.0,
        fmax=None,
        **kwargs,
    ):

        self.fmin = fmin
        self.fmax = fmax
        self.n_mels = n_mels
        super().__init__(
            n_filters=n_filters,
            kernel_size=kernel_size,
            stride=stride,
            window=window,
            sample_rate=sample_rate,
            **kwargs,
        )
        self.mel_scale = MelScale(
            n_filters, sample_rate=sample_rate, n_mels=n_mels, fmin=fmin, fmax=fmax
        )
        self.n_feats_out = n_mels

    def post_encode(self, spec: torch.Tensor):
        return self.mel_scale(spec)

    def get_config(self):
        conf = dict(fmin=self.fmin, fmax=self.fmax, n_mels=self.n_mels)
        return {**super().get_config(), **conf}
```

popcornell:
I think it looks much better !

mpariente:
Going for it then ! :wink:

Issue #318, opened 2020-11-07
Need feedback: BaseModel.sample_rate re-design


mpariente:
Currently, the `BaseModel` looks something like this. 

```python
class BaseModel(nn.Module):
    def __init__(self):
        super().__init__()

    @property
    def sample_rate(self):
        """Operating sample rate of the model (float)."""
        raise NotImplementedError
```

The interface to child model is good, we have `model.sample_rate` work as expected, and it's nice because it's a property. 

If a user wants to write his own child class of `BaseModel`, it would look like this

```python
class Model(BaseModel):
    def __init__(self, sample_rate):
        super().__init__()
        self._sample_rate = sample_rate
        
    @property
    def sample_rate(self):
        return self._sample_rate
```

which is fine, but not obvious to everybody. 

We could change `BaseModel` to look like this: 
```python
class BaseModel(nn.Module):
    def __init__(self, sample_rate):
        super().__init__()
        self._sample_rate = sample_rate

    @property
    def sample_rate(self):
        return self._sample_rate
```

and inheritance would now look like this: 
```python
class Model(BaseModel):
    def __init__(self, sample_rate):
        super().__init__(sample_rate=sample_rate)
```

which is much more user-friendly. 

What would be the downsides of this? 
@jonashaag @mhu-coder @popcornell @JorisCos feedback appreciated.

mhu-coder:
As every model needs to have `sample_rate` to be easy to use at inference, it makes sense to have it in BaseModel. In this case, I guess, `BaseEncoderMaskerDecoder` needs to change
```diff
- super.__init__()
+ super.__init__(sample_rate=encoder.sample_rate)
```
and none of the models inheriting from `BaseEncoderMaskerDecoder` would need to change IIRC.

However, at the moment, the `sample_rate` attribute points back to the value stored in `Filterbank` which ensures consistency throughout the model. At the moment, `sample_rate` is a public attribute, implying that changing it is fine. If we go forward with adding `sample_rate` to `BaseModel`, would it make sense to make `Filterbank.sample_rate` "private"so that people changing its value would know that they are not supposed to?

popcornell:
IMO i don't see when a user will overwrite without knowing the `sample_rate `attribute. 
He would have to do some monkey patching and the chance is that a user which will try to do some monkey patching will actually know what he is going to do. 
I would keep it as simple as possible. I actually had troubles when trying to implement my own model due to the property tag. 

Python does not directly implement private and public attributes and methods. 
[...]

PR #336, opened 2020-11-19
BaseModel takes sample_rate argument

mpariente:
Addresses #318

From what I understood, the main concern in passing `sample_rate` to `BaseModel` is that we duplicate reference to it, and users might change it without changing other part of the model that reference it. See [comment](https://github.com/mpariente/asteroid/issues/318#issuecomment-723481927). 

This is what I propose: 
Having `__sample_rate` with dunder means that `model.__sample_rate` cannot be accessed/changed directly. But it can be accessed from `model._BaseModel__sample_rate` which most users won't even know IMO. 
Instead, we provide a setter for the `sample_rate` property and warn the user that there might be some inconsistencies. 
WDYT? 

### Proof of concept
```python
In [1]: class Trial:
   ...:     def __init__(self, sample_rate):
   ...:         self.__sample_rate = sample_rate
   ...: 
   ...:     @property
   ...:     def sample_rate(self):
   ...:         return self.__sample_rate
   ...: 
   ...:     @sample_rate.setter
   ...:     def sample_rate(self, new_sample_rate):
   ...:         self.__sample_rate = new_sample_rate
   ...: 

In [2]: model = Trial(8000.0)

In [3]: model.__sample_rate
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-c29e6b91ff0f> in <module>
----> 1 model.__sample_rate

AttributeError: 'Trial' object has no attribute '__sample_rate'

In [4]: model._Trial__sample_rate
Out[4]: 8000.0

In [5]: model._Trial__sample_rate = 10

In [6]: model.sample_rate
Out[6]: 10

In [7]: model.sample_rate = 12

In [8]: model.sample_rate
Out[8]: 12

In [9]: model._Trial__sample_rate
Out[9]: 12

```

Obviously, the rest of the models need to account for the change, I just propose this design and will fix the rest when we agreed.

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -32,8 +32,9 @@ class BaseModel(nn.Module):
 
     """
 
-    def __init__(self):
+    def __init__(self, sample_rate: float = 8000.0):
         super().__init__()
+        self.__sample_rate = sample_rate
 
     def forward(self, *args, **kwargs):
         raise NotImplementedError
@@ -41,7 +42,16 @@ class BaseModel(nn.Module):
     @property
     def sample_rate(self):
         """Operating sample rate of the model (float)."""
-        raise NotImplementedError
+        return self.__sample_rate
+
+    @sample_rate.setter
+    def sample_rate(self, new_sample_rate: float):
+        warnings.warn(
+            "Other sub-components of the model might have a `sample_rate` "
+            "attribute, be sure to modify them for consistency.",
+            UserWarning,
+        )
+        self.__sample_rate = new_sample_rate
 
     @torch.no_grad()
     def separate(self, wav, output_dir=None, force_overwrite=False, resample=False, **kwargs):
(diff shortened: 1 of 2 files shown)


jonashaag:
I think this is a solid solution. But maybe the expected cost this scenario:

> users might change it without changing other part of the model that reference it

(with cost = probability of someone hitting that use × time it takes them to fix the issue) isn't worth doing such a "complicated" solution and we can simply go with a normal attribute instead?

popcornell (approved):
Seems fine to me, i like the use of the setter. 
It is advanced python but users should not be concerned with this now as they will simply provide sample_rate as argument and that is it.

mpariente:
I think it's nice to be able to raise the warning, which justifies this approach.
Going to fix the rest, add tests, and merge since you seem ok. 
Thanks for the review

PR #337, opened 2020-11-19
Rename _separate to forward_wav

mpariente:
Proposition: rename `_separate` to `forward_wav` in `BaseModel`. 

This is something that was a bit discussed in #305  : asking users to overwrite a private method doesn't feel right. This fixes it. 

Possibly in conflict with #305


asteroid/models/base_models.py
```diff
@@ -150,16 +167,22 @@ def file_separate(
...
         Args:
             wav (Union[torch.Tensor, numpy.ndarray, str]): waveform array/tensor.
                 Shape: 1D, 2D or 3D tensor, time last.
-
-        Returns:
-            The output of self(wav, *args, **kwargs).
         """
+        if is_overridden("_separate", self, parent=BaseModel):
+            return self._separate(wav, *args, **kwargs)
+        return self(wav, *args, **kwargs)
```
mpariente, line 181: Ok, maybe the `is_overriden` is completely overkill.. 
Removing `_separate` from the base class and `hasattr(self,"_separate")` + raise warning if it does would have been enough. 
Hahah, I happy to write that function anyway :sweat_smile:

mpariente:
This is ready IMO.
Waiting for review.

asteroid/models/base_models.py
```diff
@@ -27,8 +28,9 @@ class BaseModel(torch.nn.Module):
     the `get_model_args` method.
 
     Models inheriting from `BaseModel` can be used by :mod:`asteroid.separate`
-    and by the `asteroid-infer` CLI. For models whose `forward` doesn't return
-    waveform tensors, overwrite `_separate` to return waveform tensors.
+    and by the `asteroid-infer` CLI. For models whose `forward` doesn't go from
+    waveform to waveform waveform tensors, overwrite `forward_wav` to return
```
jonashaag, line 32:
```suggestion
    waveform to waveform tensors, overwrite `forward_wav` to return
```

asteroid/models/base_models.py
```diff
@@ -68,16 +70,29 @@ def file_separate(self, *args, **kwargs):
         """Convenience for ``asteroid.separate.file_separate(self, ...)``."""
         return separate.file_separate(self, *args, **kwargs)
 
-    def _separate(self, wav, *args, **kwargs):
-        """Hidden separation method
+    def forward_wav(self, wav, *args, **kwargs):
+        """Separation method for waveforms.
+        In case the network's `forward` doesn't have waveforms as input/output,
```
jonashaag, line 75:
```suggestion
        """Separation method for waveforms.

        In case the network's `forward` doesn't have waveforms as input/output,
```

jonashaag: We should document the expected return shape

asteroid/separate.py
```diff
@@ -45,23 +45,25 @@ def separate(
...
         Union[torch.Tensor, numpy.ndarray, None], the estimated sources.
             (batch, n_src, time) or (n_src, time) w/o batch dim.
 
     .. note::
-        By default, `separate` calls `model._separate` which calls `forward`.
-        For models whose `forward` doesn't return waveform tensors,
-        overwrite their `_separate` method to return waveform tensors.
+        `separate` calls `model.forward_wav` which calls `forward` by default.
+        For models whose `forward` doesn't have waveform tensors as input/ouput,
+        overwrite their `forward_wav` method to to separate from waveform to waveform.
```
jonashaag, line 66:
```suggestion
        overwrite their `forward_wav` method to separate from waveform to waveform.
```

asteroid/utils/deprecation_utils.py
```diff
@@ -48,3 +49,49 @@ def wrapped(*args, **kwargs):
...
+        for cls in mro_list[::-1]:
+            if hasattr(cls, fn):
+                return cls
+        return None
+
+    if not hasattr(obj, method_name):
+        return False
+
+    instance_attr = getattr(obj, method_name)
+    if not instance_attr:
```
jonashaag, line 84:
```suggestion
    try:
        instance_attr = getattr(obj, method_name)
    except AttributeError:
        return False
```

jonashaag: Otherwise, 1) will raise `AttributeError` if the attr doesn't exist, and 2) the check will consider falsey values as missing.

asteroid/models/base_models.py
jonashaag: Added whitespace so that Sphinx makes a line break, assuming that was your intention?

jonashaag:
See comments, otherwise LGTM

asteroid/models/base_models.py
mpariente: Yes, thanks.

PR #338, opened 2020-11-19
DNS dataset: 'segment' and sr support, clean-only loading

jonashaag:
- Add support for `segment`
- Add support for loading mixture + clean only, no noise
- Had to add sample rate for `segment` implementation

The way the code deals with files that are too short is different from the other datasets (here: zero pad to the right). This needs to be discussed.

### DIFF
--- a/asteroid/data/dns_dataset.py
+++ b/asteroid/data/dns_dataset.py
@@ -4,6 +4,8 @@ import json
 import os
 import soundfile as sf
 
+from asteroid.utils import get_wav_random_start_stop
+
 
 class DNSDataset(data.Dataset):
     """Deep Noise Suppression (DNS) Challenge's dataset.
@@ -18,9 +20,11 @@ class DNSDataset(data.Dataset):
 
     dataset_name = "DNS"
 
-    def __init__(self, json_dir):
-
+    def __init__(self, json_dir, sample_rate=16000, segment=None, load_noise=True):
         super(DNSDataset, self).__init__()
+        self.sample_rate = sample_rate
+        self.segment = segment
+        self.load_noise = load_noise
         self.json_dir = json_dir
         with open(os.path.join(json_dir, "file_infos.json"), "r") as f:
             self.mix_infos = json.load(f)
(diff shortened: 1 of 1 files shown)


asteroid/data/dns_dataset.py
```diff
@@ -33,28 +37,54 @@ def __len__(self):
...
-        infos["task"] = "enhancement"
-        infos["licenses"] = [dns_license]
-        return infos
+        return {
+            "dataset": self.dataset_name,
+            "sample_rate": self.sample_rate,
+            "segment": self.segment,
+            "task": "enhancement",
+            "licenses": [dns_license],
+            "n_src": 2 if self.load_noise else 1,
```
mpariente, line 86: I'm unsure about that. This will depend on the model. You might need to noise to compute a mask of something but still output only the speech in your model. 
I'd have to check all the places this `n_src` is used to be sure it's ok.

asteroid/data/dns_dataset.py
```diff
@@ -33,28 +37,54 @@ def __len__(self):
...
+            noise = torch.from_numpy(
+                sf.read(utt_info["noise"], dtype="float32", start=start, stop=stop)[0]
+            )
+            if self.segment is not None:
+                noise = torch.nn.functional.pad(
+                    noise, (0, max(0, int(self.segment * self.sample_rate) - len(noise)))
+                )
+            return x, speech, noise
+        else:
+            return x, speech
```
mpariente, line 72: Use `if not self.load_noise: return x, speech` and the rest, without the `else` clause.

asteroid/data/dns_dataset.py
```diff
@@ -33,28 +37,54 @@ def __len__(self):
...
         """
         utt_info = self.mix_infos[self.wav_ids[idx]]
         # Load mixture
-        x = torch.from_numpy(sf.read(utt_info["mix"], dtype="float32")[0])
+        x_np, sr = sf.read(utt_info["mix"], dtype="float32")
+        assert sr == self.sample_rate
+        start, stop = get_wav_random_start_stop(
+            len(x_np), int(self.segment * self.sample_rate) if self.segment is not None else None
+        )
+        x = torch.from_numpy(x_np[start:stop])
```
mpariente, line 49: Use `start` and `stop` in soundfile.
You can use this to get the length without loading the full file (just reads the header)
```python
wav = sf.SoundFile(filename)
len(wav)
```

asteroid/data/dns_dataset.py
```diff
@@ -33,28 +37,54 @@ def __len__(self):
...
         # Load clean
-        speech = torch.from_numpy(sf.read(utt_info["clean"], dtype="float32")[0])
-        # Load noise
-        noise = torch.from_numpy(sf.read(utt_info["noise"], dtype="float32")[0])
-        return x, speech, noise
+        speech = torch.from_numpy(
+            sf.read(utt_info["clean"], dtype="float32", start=start, stop=stop)[0]
+        )
+        if self.segment is not None:
+            x = torch.nn.functional.pad(
```
mpariente, line 55: Padding seems ok to me.

mpariente (approved):
Generally, this looks fine.
Let's write a `load_and_pad` function to factor out it's three usage, it'll be much easier to read IMO.

asteroid/data/dns_dataset.py
mpariente: Actually, if we write `load_and_pad`, the above change is not needed, things will still be very readable.

asteroid/data/dns_dataset.py
jonashaag: Oh I though `n_src` in the dataset info means the number of sources returned by `__getitem__`

asteroid/data/dns_dataset.py
jonashaag: Ah, using `utt_info["file_len"]`?

mpariente: IDK, I don't remember if there is the `file_len` field in utt_info.. 
If there is, then let's use it. Otherwise, use `sf.SoundFile` because it doesn't load the file in memory, just read the header and is faster.

asteroid/data/dns_dataset.py
mpariente: Actually, none of the datasets return `n_src` in info, so it's a choice we have to make. 
`get_infos` is only used for publishing models for now, and `n_src` is not used, is there specific reason you added it?

jonashaag: No, we can leave it out for now

jonashaag:
I wonder if we should close this or wait for our dataset refactoring research to be done...

mpariente:
It depends if we need those changes fast, do we?

jonashaag:
Personally I don't :)

mpariente:
Ok, I close it then. 
Let's keep the `load_and_pad` in mind though

PR #339, opened 2020-11-19
get_wav_rand_start_stop: Add random generator support

jonashaag:
I needed reproducible randomness in this particular function. I think it's valuable to add this to any code that uses randomness. Let's discuss :)

### DIFF
--- a/asteroid/utils/generic_utils.py
+++ b/asteroid/utils/generic_utils.py
@@ -65,19 +65,24 @@ def average_arrays_in_dic(dic):
     return dic
 
 
-def get_wav_random_start_stop(signal_len, desired_len=4 * 8000):
+def get_wav_random_start_stop(signal_len, desired_len=4 * 8000, np_rng=None):
     """Get indexes for a chunk of signal of a given length.
 
     Args:
         signal_len (int): length of the signal to trim.
         desired_len (int): the length of [start:stop]
+        np_rng (numpy.random.Generator, optional): The NumPy random generator to use.
+            If `None` (the default), uses the generatored returned by
+            :func:`numpy.random.default_rng()`.
 
     Returns:
         tuple: random start integer, stop integer.
     """
     if desired_len is None:
         return 0, signal_len
-    rand_start = np.random.randint(0, max(1, signal_len - desired_len))
+    if np_rng is None:
+        np_rng = np.random.default_rng()
+    rand_start = np_rng.integers(0, max(1, signal_len - desired_len))
     stop = min(signal_len, rand_start + desired_len)
     return rand_start, stop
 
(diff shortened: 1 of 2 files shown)


mpariente:
Can you show an example please?

jonashaag:
Sure! Why I wanted this is a) for debugging and b) for determinism.

For a), in my augmentation code (which uses lots of randomness), if there was an error I wanted to be able to reproduce the exact same randomness to debug the error. So I'd get the random seed from the logs.

For b), you could eg. use seed = sample ID, like this:


```py
def __getitem__(self, idx):
    rng = np.random.default_rng(seed=idx)
    ... = get_wav_rand_start_stop(..., random=rng)
```

mpariente:
Thanks. 
This sounds fine, but needs to be clearly documented. 
Maybe starting with a different name, I wouldn't know that it expects a `numpy.random._generator.Generator` from reading the signature.
Maybe `np_rng`?

jonashaag:
We could also make it use Torch RNG, maybe that's better?

mpariente:
I really know nothing about RNG. 
Numpy seems fine but up to you.

jonashaag:
Decided to go with NumPy RNG because that's what's being used everywhere else.

The change I made is backwards incompatible in the sense that it produces different random numbers. So if someone used `np.seed()` to make their randomness deterministic, they will get different numbers now.

On a related note, there is a general problem with using NumPy randomness because if you use multiple data loaders, all of the data loaders will have the same seed (because PyTorch only changes its own RNG seed but not NumPy's). Not sure if PyTorch-Lightning fixes that for you though.

mpariente:
Cannot we get the current RNG, to make it backward compatible? 
Maybe, it's not worth the effort, I'm not sure. 

When we use `pl.seed_everything`, results won't be deterministic and this is a problem..

asteroid/utils/generic_utils.py
```diff
@@ -65,19 +65,24 @@ def average_arrays_in_dic(dic):
...
 
     Returns:
         tuple: random start integer, stop integer.
     """
     if desired_len is None:
         return 0, signal_len
-    rand_start = np.random.randint(0, max(1, signal_len - desired_len))
+    if np_rng is None:
+        np_rng = np.random.default_rng()
+    rand_start = np_rng.integers(0, max(1, signal_len - desired_len))
```
mpariente, line 85: Maybe, instead. We can draw from `np_rng` if it's fed, otherwise, use `np.random.randint(0, max(1, signal_len - desired_len))`, WDYT?

mpariente:
Should we go on with this?

mpariente:
@jonashaag, any updates on that?

jonashaag:
No bandwidth to work on it currently, we can close if you prefer

mpariente:
We should switch all the randomness to use torch, because of [this](https://tanelp.github.io/posts/a-bug-that-plagues-thousands-of-open-source-ml-projects/). 
Closing this PR then.

Issue #335, opened 2020-11-19
Better names in filterbanks.transforms 


mpariente:
The conversation started [here](https://github.com/mpariente/asteroid/pull/333#discussion_r526647058), let's move it to an issue for more visibility. 

So, the `transforms.py` file has been here almost untouched since the first draft of Asteroid, and the names in there are not so great. 
Since then `complex_nn.py` has introduced "conflicting" names (`torch_complex_from_magphase` compared to `from_mag_and_phase` in `transform`). 

- `take_reim` is useless, and even confusing. You might expect to receive a tuple of (re, im) when calling it, but you get your input. This was introduced to be used in `Container`, a base class that disappeared since then. 
- `take_cat` is also not as clear as it could be.

### Propositions
1. Remove `take_reim`.
2. Rename `take_mag` to `mag`
3. Rename `take_cat` to `magreim` ???
3. Rename `to_polar` to `magphase`
4. Rename `from_mag_and_phase` to `from_magphase`
5. Keep `apply_*_mask`
6. Keep all in `complex_nn` untouched
7. Introduce `reim` that returns `(real, imag)` tuple? 

Of course, this means deprecating things properly.

popcornell:
Makes sense for me

jonashaag:
For reference, in torchaudo there are

- `magphase(x) -> (mag(x), angle(x))`
- `angle(x)` and `complex_norm(x)` (= `mag`)

and in Torch there are

- `.angle()`
- `.abs()`

🙄

I don't really have an opinion on the names other than we should not introduce multiple names in Asteroid in addition to the Torch names :-D

mpariente:
We could have `angle`, but then `magphase` is weird, why not `magangle`?
OTOH, `abs` is not appropriate IMO (but it is for native tensors).

mpariente:
Closed in #342

PR #342, opened 2020-11-20
Better names in asteroid.filterbanks.transforms

mpariente:
As announced in #335 

I didn't add a version, but I guess it'll be the release after the next.

### DIFF
--- a/asteroid/filterbanks/transforms.py
+++ b/asteroid/filterbanks/transforms.py
@@ -1,9 +1,9 @@
+import math
+import torch
 from typing import Tuple
 
-import torch
-import numpy as np
-
 from ..utils.torch_utils import script_if_tracing
+from ..utils.deprecation_utils import mark_deprecated
 
 
 def mul_c(inp, other, dim: int = -2):
@@ -47,11 +47,18 @@ def mul_c(inp, other, dim: int = -2):
     return torch.cat([real1 * real2 - imag1 * imag2, real1 * imag2 + imag1 * real2], dim=dim)
 
 
-def take_reim(x, dim: int = -2):
-    return x
+def reim(x, dim: int = -2) -> Tuple[torch.Tensor]:
+    """Returns a tuple (re, im).
+
+    Args:
+        x (:class:`torch.Tensor`): Complex valued tensor.
+        dim (int): frequency (or equivalent) dimension along which real and
+            imaginary values are concatenated.
+    """
+    return torch.chunk(x, 2, dim=dim)
 
 
-def take_mag(x, dim: int = -2, EPS: float = 1e-8):
+def mag(x, dim: int = -2, EPS: float = 1e-8):
     """Takes the magnitude of a complex tensor.
 
     The operands is assumed to have the real parts of each entry followed by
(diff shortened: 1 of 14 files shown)


asteroid/filterbanks/transforms.py
```diff
@@ -47,11 +47,18 @@ def mul_c(inp, other, dim: int = -2):
     return torch.cat([real1 * real2 - imag1 * imag2, real1 * imag2 + imag1 * real2], dim=dim)
 
 
-def take_reim(x, dim: int = -2):
-    return x
+def reim(x, dim: int = -2) -> Tuple[torch.Tensor]:
```
jonashaag, line 50: Shouldn't this be `Tuple[torch.Tensor, torch.Tensor]`

jonashaag:
Sorry for making this late suggestion, but since we're changing names anyway... how about moving the "Asteroid complex" functions into a submodule, like `asteroid.asteroid_complex`? If I understand correctly most/all of these functions only work on Asteroid-style complex numbers by default.

asteroid/filterbanks/transforms.py
mpariente: I'm not very familiar with typing yet, you're right, I'll fix it.

mpariente:
No problem at all. 


> most/all of these functions only work on Asteroid-style complex numbers 

This is correct. Three exceptions: `compute_delta`, `concat_deltas` and `ebased_vad` 
What do you suggest exactly? What would we do with those three exceptions? 
Would `complex_nn` stay untouched?

jonashaag:
I don't know where these would fit. Are those "transforms"? I guess yes, but many thing are "transforms", so I really don't know how to classify those, sorry!

We could move the rest to `asteroid.asteroid_complex` and add a note to the top of the module that this is only one of the three complex representations (as in `complex_nn`). And `complex_nn` stays the same.

mpariente:
Agree that "transforms" is not informative. 
Several person already asked if we can outsource the code for asteroid filterbanks outside of asteroid and I wonder today if this wouldn't be a good opportunity. What your take on this?

mpariente:
The only functions on which the filterbank submodule depend are `script_if_tracing` and `mixture_consistency` which are unlikely to change so I'd be ok to copy them there as well.. Not the best design but I guess it's ok. 

The advantages of this: 
- More people would use it (`torch-audiomentations` and `pyannote-audio` for example).
- More people might contribute to it? Not sure. 

And, the API is stable IMO so we don't expect to struggle to keep things compatible. 

Mabe @hbredin would be happy about that?

hbredin:
Happy about having a standalone `asteroid.filterbanks`  package? I certainly would 💯!

jonashaag:
Well this PR escalated quickly 😄

@mpariente I think what you suggest is reasonable (including the tiny amount of copy pasta).

mpariente:
> Well this PR escalated quickly :smile: 

I think it's because if we'd move things to `asteroid.asteroid_complex`, making a new package out of those things would be a pain..

asteroid/filterbanks/transforms.py
mpariente: And now the linter is complaining.. 
```def chunk(self: Tensor, chunks: _int, dim: _int=0) -> Union[Tuple[Tensor, ...], List[Tensor]]: ...```

PR #347, opened 2020-11-23
BaseModel shape reconstruction: Use shape instead of waveform

jonashaag:
Does this work? I was wondering this just now when working on the trimming/padding stuff in #276. It could be a way to save some GPU memory.

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -6,7 +6,7 @@ import torch
 from torch import nn
 
 from ..masknn import activations
-from ..utils.torch_utils import pad_x_to_y, script_if_tracing, jitable_shape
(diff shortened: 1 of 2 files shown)


mpariente:
In our code, it'll work because we pass a `jitable_shape`. But if I pass a "normal" shape (e.g `x.shape`), this won't be scriptable anymore.

I'm undecided about this. 
And it doesn't really save GPU memory because you pass a pointer. If you'd keep reassigning the same variable name to garbage collect and save memory, but you can't because you need to keep the reference, then it makes sense, but I'm not convinced that's a bottleneck.

I'd love to have feedback from torchscript experts, because I'm quite sure there are some things that we're overcomplicating :sweat_smile:

jonashaag:
Well at least the tests are ok so we know this works. I'll close this; we can keep it for further reference if we want to work on it in the future

PR #349, opened 2020-11-24
Add asteroid-versions script to print installed versions

jonashaag:
We could use this as for the issue templates.

```
$ asteroid-versions
Asteroid            0.4.0rc0, Git checkout f7938f3413bb (asteroid-env)
PyTorch             1.7.0+cpu
PyTorch-Lightning   0.9.0
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -28,7 +28,21 @@ Steps to reproduce the behavior (code sample and stack trace):
 
 ### Environment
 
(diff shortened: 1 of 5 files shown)


mpariente:
Looks great, thanks ! 
Let's add a test for it and merge it

jonashaag:
Well it's going to be a mess to test this properly (checking all of the values etc) but how about adding a simple test that checks that the function does anything at all?

mpariente:
Yes, just execute it in a test file. Not rigorous unit test :sweat_smile:

jonashaag:
Added suggestion for update to issue template, feel free to ignore or change :)

PR #350, opened 2020-11-24
[CI] Build with torch-nightly 

mpariente:
Ugly, but simple (if it works).

I'd be ok with that for now.

### DIFF
--- a/.github/workflows/test_asteroid_package.yml
+++ b/.github/workflows/test_asteroid_package.yml
@@ -11,10 +11,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6]  #, 3.7, 3.8]
-        pytorch-version: [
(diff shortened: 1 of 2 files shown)


jonashaag:
Yes looks ok. Shall we also test with 1.6?

mpariente:
Yes, let's do that as well

jonashaag:
Very nice!

mpariente:
Don't know.. 
It's not the first time it happens.

mpariente:
Approved solution: https://github.community/t/manually-restart-actions-and-entire-workflows/16262/2

:joy: :joy: :joy:

jonashaag:
I've been told it only works with that exact commit message

mpariente:
I'm wondering if it's necessary to run the tests 6 times

PR #353, opened 2020-11-24
Fix rebase action

mpariente:
Try fixing it by adding a dummy commit afterwards.


mpariente:
It seems that the command fetches the version in master.. No way to try it easily without ruining master..

Issue #332, opened 2020-11-18
Asteroid conda environment


mcernak:
## 🚀 Feature
I suggest managing the project's dependencies using the conda environment.

### Motivation

Conda is a cross-platform management system. Besides, it allows clear separation of different projects on a single computer.

### What you'd like

Create PR. `requirements.txt` would be replaced by `environment.yml.`

### Alternatives

Keep the status quo.

mpariente:
Thanks for raising the issue. 

Do you mean being able to install asteroid with `conda install asteroid`, or just having a `environment.yml` from which a conda env can be created? 

I'm unsure about forcing people to use conda instead of pip..

mcernak:
I meant the latter one; looks to be a better way of managing dependencies. I suggest having `environment.yml` (attached), creating the env by `conda env create -f environment.yml `, and then `conda activate asteroid`. That's it.

I know many developers who love conda as it makes life easier. But you are also right, it is just one way how to manage dependencies.

[environment.yml.txt](https://github.com/mpariente/asteroid/files/5562456/environment.yml.txt)

mpariente:
This is cool indeed and I've had this discussion few times but I still didn't get the answer I'm looking for. 

In the .yml file, it's possible to add `-r requirements.txt` in the `pip` section so if we'd create a `environment.yml` file with only this as content, that would probably work for both pip and conda users. 
If we do that, the all deps are installed with pip, is it a problem? If yes, how? 

I only use conda to install torch and torchaudio. For the rest, I use pip. So I wouldn't see the problem in my case, but not sure about others. 

I know @jonashaag has a tooth against conda, curious to have your opinion.

jonashaag:
> If we do that, the all deps are installed with pip, is it a problem? If yes, how?

It can become a problem because Conda resolves pip installed packages and their dependencies differently from native ones. I think they are working on interoperability but it’s not done yet. So if you’re using Conda you’re probably best off installing as little as possible using pip. 

> I only use conda to install torch and torchaudio. For the rest, I use pip. So I wouldn't see the problem in my case, but not sure about others.

That’s exactly how I do it now because I’ve had a ton of issues with Conda but it’s still better at installing the correct CUDA, Torch, MKL, etc. versions.

jonashaag:
[...]

PR #354, opened 2020-11-25
Add conda environment

mcernak:


mpariente:
Ok, I didn't realize that this is what you meant. 
This seems to complicate the instructions without any benefit. Any user can create an empty env, we don't need a empty env file for that right?

mcernak:
My original idea was to use the complete environment as I uploaded in #332. 

But then you asked to use `requirements.txt` for pip install, which is not possible within `environment.yml.`

A solution would be to create an empty conda env, activate it, and install whatever is needed, which updates the environment.

Alternatively, I can prepare the full conda env as I thought.

mpariente:
You can have `-r requirements.txt` in the pip subsection of conda.  

Have a look [here](https://github.com/conda/conda/blob/54e4a91d0da4d659a67e3097040764d3a2f6aa16/tests/conda_env/support/advanced-pip/environment.yml)

We can have 
```yaml
name: asteroid
channels:
  - anaconda
  - conda-forge
dependencies:
  - python=3.6
  - Cython
  -pip
  -pip:
    - -r file:requirements.txt
    - -e .
```
and at least everything is installed in one go.

mcernak:
Correct, I updated the environment.

PR #357, opened 2020-11-25
Test error

jonashaag:


jonashaag:
I think this works now.

Is there a way to look at CI-built docs?

mpariente:
Yes, I upload it as an artifact. 

image

image

PR #358, opened 2020-11-25
[tests] Fix complex and add tests

mpariente:
This PR aims at improving the coverage. 

On the way, I found other small import issues.


asteroid/complex_nn.py
```diff
@@ -61,16 +61,19 @@ def as_torch_complex(x, asteroid_dim: int = -2):
...
         is_asteroid_complex = transforms.is_asteroid_complex(x, asteroid_dim)
         if is_torchaudio_complex and is_asteroid_complex:
-            raise ValueError(
+            warnings.warn(
                 f"Tensor of shape {x.shape} is both a valid Torchaudio-style and "
                 "Asteroid-style complex. PyTorch complex conversion is ambiguous."
+                "Assuming Torchaudio-style.",
+                RuntimeWarning,
             )
+            return torch.view_as_complex(x)
```
mpariente, line 70: I thought you'd like to debate that @jonashaag.
It's very unlikely that an asteroid tensor will have length 2 in it's last dimension, however very possible that a torchaudio tensor has even shape or dim -2. That's why we should default to torchaudio style IMO.

asteroid/complex_nn.py
```diff
@@ -61,16 +61,19 @@ def as_torch_complex(x, asteroid_dim: int = -2):
...
         elif is_torchaudio_complex:
             return torch.view_as_complex(x)
         elif is_asteroid_complex:
             return torch.view_as_complex(transforms.to_torchaudio(x, asteroid_dim))
         else:
-            raise ValueError(
+            raise RuntimeError(
```
mpariente, line 76: Actually, not a ValueError.

mpariente:
```python
In [9]: conv1d(torch.randn(1), torch.randn(1))
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-77ce17263db7> in <module>
----> 1 conv1d(torch.randn(1), torch.randn(1))

RuntimeError: weight should have at least three dimensions

In [10]: conv1d(torch.randn(1), torch.randn(1, 1, 1))
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-10-b74bee98ca1e> in <module>
----> 1 conv1d(torch.randn(1), torch.randn(1, 1, 1))

RuntimeError: Expected 3-dimensional input for 3-dimensional weight [1, 1, 1], but got 1-dimensional input of size [1] instead

In [11]: conv1d(torch.randn(11, 2, 3), torch.randn(1, 1, 1))
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-11-3f5f81e79961> in <module>
----> 1 conv1d(torch.randn(11, 2, 3), torch.randn(1, 1, 1))

RuntimeError: Given groups=1, weight of size [1, 1, 1], expected input[11, 2, 3] to have 1 channels, but got 2 channels instead
```

asteroid/complex_nn.py
```diff
@@ -54,23 +58,27 @@ def as_torch_complex(x, asteroid_dim: int = -2):
         ValueError: If type of `x` is not understood.
     """
     if isinstance(x, (list, tuple)) and len(x) == 2:
-        return torch_complex_from_magphase(*x)
+        return torch_complex_from_reim(*x)
```
mpariente, line 61: I'd be in favour of removing it completely as it will be confusing if it works from mag and phase or from real and imag. 
And using either of `torch_complex_from*` will be more explicit.

jonashaag: Or rather, rename it to `torch_complex_from_reim`?

asteroid/complex_nn.py
jonashaag: Hmm... I'm not sure what's the advantage of handling this case. The idea of this generic conversion function was that if you write some code (say a library) that uses complex numbers, you can accept any type of complex representation. So you don't have to know the representation upfront.

If you know upfront in the library what type of tensor you have then you can do the conversion directly without this function.

If you don't know what you have in the library, then IMO the library should complain to its user that it got input that it didn't understand, which is what the current code does.

mpariente: I knew you would disagree hahah 
As we discussed. Let's remove this function entirely and add `to_torch_complex` is `asteroid-filterbanks`.

mpariente:
\rebase

asteroid/models/dccrnet.py
```diff
@@ -33,11 +32,9 @@ def __init__(self, *args, stft_kernel_size=512, **masknet_kwargs):
     def forward_encoder(self, wav):
         tf_rep = self.encoder(wav)
         # Remove Nyquist frequency bin
-        return complex_nn.as_torch_complex(tf_rep)[..., :-1, :]
+        return to_torch_complex(tf_rep)[..., :-1, :]
```
mpariente, line 35: Here.

asteroid/models/dccrnet.py
```diff
@@ -33,11 +32,9 @@ def __init__(self, *args, stft_kernel_size=512, **masknet_kwargs):
...
 
     def apply_masks(self, tf_rep, est_masks):
         masked_tf_rep = est_masks * tf_rep.unsqueeze(1)
         # Pad Nyquist frequency bin
-        return from_torchaudio(
-            torch.view_as_real(torch.nn.functional.pad(masked_tf_rep, (0, 0, 0, 1)))
-        )
+        return from_torch_complex(torch.nn.functional.pad(masked_tf_rep, [0, 0, 0, 1]))
```
mpariente, line 40: Here

tests/cli_test.sh
```diff
@@ -0,0 +1,18 @@
+# Save a model (tmp.th) and two wavfiles (tmp.wav, tmp2.wav)
+python -m pip install -e . --quiet
```
mpariente, line 2: Needed to install. Otherwise the CLI didn't work. Don't know why.

mpariente (commented):
Ready for review.

PR #362, opened 2020-11-27
[hub] TorchHub integration

mpariente:
We need a test for it, because module level imports of other deps will break everyting. 
It's a simple test to make so let's set it up.


mpariente:
Q: how to handle additional deps? 
Should I install torchaudio, torch_stoi etc.. in the CI which supposes that the user will have to as well, or I try to find a way that hub works with the least deps? 
Not clear IMO

mpariente:
So, this works but I had to install numpy torchaudio and torch_stoi. 
We could bypass `torch_stoi` and `torch_audio` quite simply.

asteroid/losses/__init__.py
```diff
@@ -1,22 +1,26 @@
...
 from .pmsqe import SingleSrcPMSQE
-from .stoi import NegSTOILoss as SingleSrcNegSTOI
 from .multi_scale_spectral import SingleSrcMultiScaleSpectral
 
+try:
+    from .stoi import NegSTOILoss as SingleSrcNegSTOI
+except ModuleNotFoundError:
+    # Is installed with asteroid, but remove the deps for TorchHub.
+    SingleSrcNegSTOI = lambda *a, **kw: print("Install `torch_stoi` to use `SingleSrcNegSTOI`")
+
```
mpariente, line 18: Point of doubt

asteroid/losses/pit_wrapper.py
```diff
@@ -1,7 +1,11 @@
 from itertools import permutations
 import torch
 from torch import nn
-from scipy.optimize import linear_sum_assignment
+
+try:
+    from scipy.optimize import linear_sum_assignment
+except ModuleNotFoundError:
+    linear_sum_assignment = None  # Hub
```
mpariente, line 8: Point of doubt

asteroid/losses/pit_wrapper.py
```diff
@@ -216,7 +220,7 @@ def find_best_perm(pair_wise_losses, perm_reduce=None, **kwargs):
                 :class:`torch.Tensor`: The indices of the best permutations.
         """
         n_src = pair_wise_losses.shape[-1]
-        if perm_reduce is not None or n_src <= 3:
+        if perm_reduce is not None or n_src <= 3 or linear_sum_assignment is None:
```
mpariente, line 223: And last one.

mpariente (commented):
I point out my three points of doubt in the comments.

jonashaag:
What's the rationale of making the dependencies "soft" dependencies? So that people don't have to bother with installing the dependencies, so merely convenience for users?

mpariente:
If you use torch.hub, you don't need to install asteroid, that's the cool part.

jonashaag:
Hmm... I'm not sure if it is worth keeping Asteroid "dependency free" "just" for Torch Hub users who don't want to install dependencies... but I'll try to give helpful feedback here anyways :D

asteroid/losses/__init__.py
jonashaag: Better raise an error rather than printing only

asteroid/losses/pit_wrapper.py
jonashaag: Hm generally speaking I don't think silently changing behaviour depending on the users environment is a good idea. It may be ok in this case, I don't know.

asteroid/masknn/recurrent.py
```diff
@@ -571,6 +568,8 @@ class DCCRMaskNet(BaseDCUMaskNet):
     _architectures = DCCRN_ARCHITECTURES
 
     def __init__(self, encoders, decoders, n_freqs, **kwargs):
+        import numpy as np
+
         self.encoders_stride_product = np.prod(
             [enc_stride for _, _, _, enc_stride, _ in encoders], axis=0
         )
```
jonashaag, line 575:
```suggestion
        self.encoders_stride_product = torch.prod(
            torch.tensor([enc_stride for _, _, _, enc_stride, _ in encoders], axis=0)
        )
```

asteroid/losses/__init__.py
mpariente: Yes, you're right, I'll do that.

asteroid/losses/pit_wrapper.py
mpariente: Maybe for scipy we can keep the deps.

mpariente:
Thanks for the constructive feedback even if you disagree ^^
I'm still undecided about it

mpariente:
I added back the numpy and scipy dependency. 
And I think it's ok to cut the torchstoi's dependency.

PR #371, opened 2020-11-30
[src] Upgrade Lightning to 1.x

mpariente:
### Fixed 
- `configure_optimizers` needed to return a dictionary when the scheduler is `ReduceLROnPlateau`.

### Changed
- Use `self.log` in training and validation step (very practical)

### Didn't find out yet
- Logging hyperparams, tried several way.

### Before merging
Need to test that the recipes still work.


mpariente:
@JorisCos could you have a look if things are working how you expect them to on your end please?

egs/librimix/ConvTasNet/train.py
```diff
@@ -84,24 +84,24 @@ def main(conf):
...
     # Don't ask GPU if they are not available.
     gpus = -1 if torch.cuda.is_available() else None
     trainer = pl.Trainer(
         max_epochs=conf["training"]["epochs"],
-        checkpoint_callback=checkpoint,
-        early_stop_callback=early_stopping,
+        callbacks=callbacks,
         default_root_dir=exp_dir,
         gpus=gpus,
         distributed_backend="dp",
```
JorisCos, line 103: I started to test on LibriMix. I'm letting you know what I've found.

This breaks if you don't have any gpu. A workaround is :

```suggestion
        distributed_backend="distributed_backend",
```
And declare `distributed_backend = "ddp" if torch.cuda.is_available() else None`
I think we should use by default `distributed_backend = "ddp"` instead of `distributed_backend = "dp"` in all recipes

JorisCos:
Not very important but is it me or there is something wrong here ? 
Capture d’écran du 2020-12-01 14-50-52 Why is it written 'loEpoch' before val_losss ?

JorisCos:
Otherwise, everything looks fine to me !

mpariente:
> Not very important but is it me or there is something wrong here ?
> Capture d’écran du 2020-12-01 14-50-52 Why is it written 'loEpoch' before val_losss ?

Yes, it does look weird. 


Actually, the `distributed_backend` has been renamed `accelerator` ([see here](https://pytorch-lightning.readthedocs.io/en/1.0.8/trainer.html#trainer-flags)), maybe it's more automatized when we use that. I'll have a look.

PR #375, opened 2020-12-03
Add device= support to asteroid-infer

jonashaag:


jonashaag:
Should be good for review!

asteroid/scripts/asteroid_cli.py
```diff
@@ -140,7 +140,20 @@ def infer():
...
+        "--device",
+        default=None,
+        type=str,
+        help="Device to run the model on, eg. 'cuda:0'."
+        "Defaults to 'cuda' if CUDA is available, else 'cpu'.",
+    )
+    args = parser.parse_args(argv)
+
+    if args.device is None:
+        device = "cuda" if torch.cuda.is_available() else "cpu"
```
mpariente, line 154: Should have a `ifcuda(with_cuda=..., without_cuda=...)` function? 
This could be rewritten as 
```python
device = args.device if args.device is not None else ifcuda("cuda", "cpu") 
```

mpariente (commented):
Looks good. 
Just one minor question

asteroid/scripts/asteroid_cli.py
jonashaag: Hmmmm... not convinced :-D because that function would be exactly

```
def ifcuda(a, b):
    return a if torch.cuda.is_available() else b
```

which isn't much different to just inlining that exact code

mpariente: Yes, it would be exactly this. 
Inlining that code would be 
```python
device = args.device if args.device is not None else "cuda" if torch.cuda.is_available() else "cpu"
```
which is much less clear. 
I'll give it more thought :wink:

jonashaag: I’m ok with both :)

mpariente:
Let's fix the linting and merge that.

PR #377, opened 2020-12-04
[hub_utils] Support for huggingface model hub

julien-c:
### Hi Asteroid team! I hereby propose an integration with the HuggingFace model hub 🤗🤗

In this Pull request:
- The few important code changes (not too many lines) are in `cached_download()`, and in `hf_bucket_url()` to support falling back to the huggingface.co model hub if we don't know what to do with the method param.
-  The rest of the additions are copy/pasted from transformers and are not strictly speaking required, but should work well and cover a few additional edge cases compared to your current file downloading/caching code.

If it's too much code, feel free to remove/update some of it. We could also spin this code off into a utility library (`pip install huggingface_hub`?) at some point.

The unit tests mostly check that we can resolve model ids to local paths, and adds an example test that loads a model from a hf model id (I had to update the model to include a sample rate inside the .bin – see [this commit](https://huggingface.co/julien-c/DPRNNTasNet-ks16_WHAM_sepclean/commit/d01f179c5687d1942a99394408eb426c18dfd03d) – this is actually a nice example of the power of model versioning 🔥)

For reference, model is at https://huggingface.co/julien-c/DPRNNTasNet-ks16_WHAM_sepclean

Also cc'ing @thomwolf for info!

### DIFF
--- a/asteroid/utils/hub_utils.py
+++ b/asteroid/utils/hub_utils.py
@@ -190,7 +190,7 @@ def hf_get_from_cache(
     resume_download=False,
     user_agent: Union[Dict, str, None] = None,
     local_files_only=False,
-) -> Optional[str]:
+) -> Optional[str]:  # pragma: no cover
     """
     Given a URL, look for the corresponding file in the local cache. If it's not there, download it. Then return the
     path to the cached file.
--- a/asteroid/utils/hub_utils.py
+++ b/asteroid/utils/hub_utils.py
@@ -72,6 +72,13 @@ def cached_download(filename_or_url):
             revision = None
         url = hf_bucket_url(model_id=model_id, filename=HF_WEIGHTS_NAME, revision=revision)
         return hf_get_from_cache(url, cache_dir=get_cache_dir())
+        # Note to maintainers:
+        # You can remove the `return hf_get_from_cache()` line above
+        # if you want to keep the exact same file downloading/caching behavior
+        # as the current one. In which case, you can remove all functions below
+        # except for `hf_bucket_url`.
+        # However the implementation adds some nice features
+        # (notably versioning-aware caching) so I'd suggest keeping it.
     cached_filename = url_to_filename(url)
     cached_dir = os.path.join(get_cache_dir(), cached_filename)
     cached_path = os.path.join(cached_dir, "model.pth")
--- a/asteroid/utils/hub_utils.py
+++ b/asteroid/utils/hub_utils.py
@@ -1,6 +1,19 @@
+import fnmatch
+import io
+import json
 import os
-from torch import hub
+import sys
+import tempfile
+from contextlib import contextmanager
+from functools import partial
 from hashlib import sha256
+from typing import BinaryIO, Dict, Optional, Union
+from urllib.parse import urlparse
+
+import requests
+import torch
+from filelock import FileLock
+from torch import hub
 
 
 CACHE_DIR = os.getenv(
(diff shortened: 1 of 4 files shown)


asteroid/utils/hub_utils.py
```diff
@@ -190,7 +190,7 @@ def hf_get_from_cache(
     resume_download=False,
     user_agent: Union[Dict, str, None] = None,
     local_files_only=False,
-) -> Optional[str]:
+) -> Optional[str]:  # pragma: no cover
```
mpariente, line 193: :joy: :joy: :joy:

asteroid/utils/hub_utils.py
```diff
@@ -39,11 +56,29 @@ def cached_download(filename_or_url):
...
+            revision = None
+        url = hf_bucket_url(model_id=model_id, filename=HF_WEIGHTS_NAME, revision=revision)
+        return hf_get_from_cache(url, cache_dir=get_cache_dir())
+        # Note to maintainers:
+        # You can remove the `return hf_get_from_cache(…)` line above
+        # if you want to keep the exact same file downloading/caching behavior
+        # as the current one. In which case, you can remove all functions below
+        # except for `hf_bucket_url`.
+        # However the implementation adds some nice features
+        # (notably versioning-aware caching) so I'd suggest keeping it.
```
mpariente, line 81: We'll keep `hf_get_from_cache` because we'll drop Zenodo support as soon as our models are fully migrated to HuggingFace's hub. 
And we'll add a note about this behaviour in the docstring above. 
Thanks for the note!

mpariente:
Thank you so much for this @julien-c, this is really exciting!

> If it's too much code, feel free to remove/update some of it. We could also spin this code off into a utility library (`pip install huggingface_hub`?) at some point.

This would make complete sense at some point, but the amount of copy/paste is fine, we'll keep everything. 

We need to somehow be able to maintain that code so I'll have to read it carefully, I don't grasp everything yet :upside_down_face:

asteroid/utils/hub_utils.py
```diff
@@ -68,3 +103,216 @@ def url_to_filename(url):
...
+    # etag is None == we don't have a connection or we passed local_files_only.
+    # try to get the last downloaded one
+    if etag is None:
+        if os.path.exists(cache_path):
+            return cache_path
+        else:
+            matching_files = [
+                file
+                for file in fnmatch.filter(os.listdir(cache_dir), filename.split(".")[0] + ".*")
+                if not file.endswith(".json") and not file.endswith(".lock")
```
jonashaag, line 248: Not a change request but just a FYI for the curious reader :-) – this would be much nicer using `pathlib`:

```py
matching_files =  [f.name for f in cache_dir.glob(cache_path.with_suffix(".*").name)
                   if f.suffix not in {".json", ".lock"}]
```

jonashaag (approved):

jonashaag:
Looks very good to me

jonashaag:
What about uploads? 🤔

mpariente:
You create a repo on the hub's page, clone it and push to it what you want, no fancy CLI required. 

image

julien-c:
re. uploads: yes. In `transformers` we have programmatic ability to create (and even delete) model repos, which we could also spin off into the same small utility-focused library at some point. 

But the design goal is to be able to upload models using just `git + git-lfs` so that it's quite independent from library implementations.

The one thing that's not supported out of the box on the upload side is upload of files larger than 5GB: you need a custom lfs transfer agent that's currently bundled to `transformers` (and in the future in a separate library). That's quite a big model size though so should be quite rare.

asteroid/utils/hub_utils.py
julien-c: True!

asteroid/utils/hub_utils.py
```diff
@@ -68,3 +103,216 @@ def url_to_filename(url):
...
+    url_to_download = url
+    etag = None
+    if not local_files_only:
+        try:
+            headers = {"user-agent": http_user_agent(user_agent)}
+            r = requests.head(
+                url, headers=headers, allow_redirects=False, proxies=proxies, timeout=etag_timeout
+            )
+            r.raise_for_status()
+            etag = r.headers.get("X-Linked-Etag") or r.headers.get("ETag")
```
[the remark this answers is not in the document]
julien-c: In case of an LFS file (i.e. where the HTTP request does not return the file's content directly, but a redirect to a Cloudfront URL), our server uses `"X-Linked-Etag"` to include the `sha256` of the actual linked file (the large file itself). 

It made more sense to us to use this hash (which we already have and don't have to compute again, as it can be very costly for super large files), but then it's not really the Etag of the redirect response itself.

We could probably document this better at some point.

asteroid/utils/hub_utils.py
mpariente, line 81: Let's remove this note.  
```suggestion
```

asteroid/utils/hub_utils.py
mpariente: Let's assume this is tested in HF, we'll test it later if we have issues with it.

PR #382, opened 2020-12-08
List asteroid models in HF's hub

mpariente:
Just so that we can do that. 

```python
In [6]: from asteroid.utils.hub_utils import model_list

In [7]: model_list(name_only=True)
Out[7]: 
['julien-c/DPRNNTasNet-ks16_WHAM_sepclean',
 'mpariente/ConvTasNet_WHAM_sepclean']

```

We'll exchange it with the current `show_available_models` etc.. when we get rid of Zenodo.

### DIFF
--- a/asteroid/utils/hub_utils.py
+++ b/asteroid/utils/hub_utils.py
@@ -5,9 +5,9 @@ import os
 import sys
 import tempfile
 from contextlib import contextmanager
-from functools import partial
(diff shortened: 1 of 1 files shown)


jonashaag:
Looks good!

jonashaag:
Of course the cache will not survive an interpreter restart. We could implement a file based cache if we find it too slow.

PR #383, opened 2020-12-08
hopefully fixes DPTNet this time 

popcornell:
see https://github.com/asteroid-team/asteroid/issues/380

### DIFF
--- a/asteroid/masknn/attention.py
+++ b/asteroid/masknn/attention.py
@@ -44,8 +44,14 @@ class ImprovedTransformedLayer(nn.Module):
         super(ImprovedTransformedLayer, self).__init__()
 
         self.mha = MultiheadAttention(embed_dim, n_heads, dropout=dropout)
-        self.recurrent = nn.LSTM(embed_dim, dim_ff, bidirectional=bidirectional)
         self.dropout = nn.Dropout(dropout)
+        self.pos_wise_ff = nn.Sequential(
+            nn.Linear(embed_dim, dim_ff),
+            activations.get(activation)(),
+            self.dropout,
+            nn.Linear(dim_ff, embed_dim),
+        )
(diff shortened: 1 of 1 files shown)


asteroid/masknn/attention.py
```diff
@@ -45,12 +45,6 @@ def __init__(
 
         self.mha = MultiheadAttention(embed_dim, n_heads, dropout=dropout)
         self.dropout = nn.Dropout(dropout)
-        self.pos_wise_ff = nn.Sequential(
```
mpariente: Why did you decide to remove it?

popcornell: Because I got confused it is not in DPTNet sorry I confused also you. 
Also we can implement the whole with two permutations. This is not efficient

SungFeng-Huang (approved):
Seems to be correct right now.

asteroid/masknn/attention.py
```diff
@@ -64,8 +58,7 @@ def forward(self, x):
         # mha is seq_len, batch, channels
         # self-attention is applied
         out = self.mha(tomha, tomha, tomha)[0]
-        out = self.pos_wise_ff(out.permute(1, 0, 2)).transpose(1, -1)
-        x = self.dropout(out) + x
+        x = self.dropout(out.permute(1, 2, 0)) + x
```
SungFeng-Huang: I'm curious whether the dropouts at line 61, 65, 66 shared the same self.dropout would cause any problem?

mpariente: It's a wrapper above a function and has no state, so it doesn't case any problem, I asked myself the same question in the past :wink: 
```python
    def forward(self, input: Tensor) -> Tensor:
        return F.dropout(input, self.p, self.training, self.inplace)
```

popcornell:
Thank you very much guys, I thought that maybe one can use only two permute operations. But our normalizations actually expects batch, channel, seq so the current version is fine. Merging.

PR #385, opened 2020-12-09
[ci] Layout for model consistency tests

mpariente:
Just checking if both timed tests and manual trigger work.


mpariente:
The action doesn't appear anywhere. Might be because it's not on master.
I'll merge and remove if it's useless.

mpariente:
For reference: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events

Issue #378, opened 2020-12-07
Switching to julius for resampling


jonashaag:
## 🚀 Feature

I'd like to start a discussion about switching to [julius](https://github.com/adefossez/julius) for resampling.

### Motivation

- Much faster resampling
- Resampling on GPU
- Can drop librosa dependency (which has a large number of other dependencies)

mpariente:
I'd be happy to drop librosa as well, its install is quite long.

PR #386, opened 2020-12-10
Drop librosa dependency and refactor requirements files #378

jonashaag:
- Use julius for resampling
- Add optional librosa support for `file_separate`
- Refactor `file_separate`

TODO: Fix librosa usage in one notebook


jonashaag:
Shall we add soundfile to the Torch test dependencies?

asteroid/separate.py
```diff
@@ -120,50 +121,84 @@ def file_separate(
...
-        if fs != model.sample_rate:
-            from librosa import resample
-
-            est_src = resample(est_src, orig_sr=model.sample_rate, target_sr=fs)
-        sf.write(save_name, est_src, fs)
+    for src_idx, est_src in enumerate(est_srcs, 1):
+        sf.write(save_name_template.format(src_idx), est_src, fs)
+
+
+def _resample(wav, orig_sr, target_sr, _resamplers={}):
```
mpariente, line 174: This is cool, and I've never seen it. 
I wonder if it's not too intricate though.

asteroid/separate.py
```diff
@@ -120,50 +121,84 @@ def file_separate(
...
+
+
+def _load_audio(filename):
+    try:
+        import librosa
+    except ModuleNotFoundError:
+        librosa = None
+
+    if librosa:
+        wav, sr = librosa.load(filename, dtype="float32", sr=None, duration=10)
```
mpariente, line 197: Why is duration equal to 10s?

asteroid/separate.py
```diff
@@ -120,50 +121,84 @@ def file_separate(
...
-                f"File {save_name} already exists, pass `force_overwrite=True` to overwrite it",
-                UserWarning,
-            )
-            return
-        if fs != model.sample_rate:
-            from librosa import resample
-
-            est_src = resample(est_src, orig_sr=model.sample_rate, target_sr=fs)
-        sf.write(save_name, est_src, fs)
+    for src_idx, est_src in enumerate(est_srcs, 1):
```
mpariente, line 170: This is much clearer.

asteroid/separate.py
```diff
@@ -120,50 +121,84 @@ def file_separate(
...
+
+    if librosa:
+        wav, sr = librosa.load(filename, dtype="float32", sr=None, duration=10)
+        # Always return wav of shape [time, n_chan]
+        if wav.ndim == 1:
+            return wav[:, None], sr
+        else:
+            return wav.T, sr
+    else:
+        return sf.read(filename, dtype="float32", always_2d=True)
```
mpariente, line 204: What is the rationale behind this? 

The only I see, supporting all the audio formats supported by librosa. 

So instead, let's try to load with soundfile directly, if it doesn't work, try to import librosa, if it doesn't work raise an error. 
Librosa has a large import time, if we can avoid it by using soundfile by default, that would be better IMO.

mpariente (commented):
This is prettier this way, cool!

mpariente:
> Shall we add soundfile to the Torch test dependencies?

Yes. For the torch.hub tests as well then.

asteroid/separate.py
jonashaag: Oopsie, debugging oversight

asteroid/separate.py
jonashaag: You mean the caching pattern? I can make it more explicit if you want. Maybe adding "cache" in the name would be helpful already.

asteroid/separate.py
jonashaag: Agreed, will change order of things!

asteroid/separate.py
mpariente: > You mean the caching pattern?

Yes. 

Including it in the name is good. `_autocached_resamplers={}` as a kwarg? 
My first idea was that I'd have to pass it to benefit from the caching and then I understood the trick because it's mutable. 

By the way, do you know where the `_resamplers` object lives? A link to the logic would be cool.

jonashaag: You mean the dictionary? It "lives in" (to be exact, is referenced by) the `_resample` function object. Funny how I take this caching pattern for granted. I don't actually know if its a wide spread pattern. I think before adding more documentation on how it works we should rather move it from the function object to the module, which is more obvious to most people.

mpariente: It's just my Python level which is not too high. 
If it's in the module, it's much simpler to get indeed.
 
For module-level `_resamplers`, it's in `globals` and it's modified from there, that seems simple. But, when it's in the kwargs default, I don't yet understand what's happening under the hood. It's not in `globals` for sure, where is the reference?

mpariente: I'm gonna read about it :smiley:

mpariente: https://docs.python.org/3/faq/programming.html#why-are-default-values-shared-between-objects

Sounds like it's a common pattern. 
Happy that I learned it!

jonashaag: Oh indeed! Nice find!

> It's not in `globals` for sure, where is the reference?

`globals()` actually returns the dict (= attributes) of the current module object (note everything is an object in Python, even ints and modules and functions). In the module attributes you find the function object. When you do `sf.read(...)` the Python interpreter looks for the `read` attribute of the `sf` (module) object and then calls it. `read` itself is an object (a function object) that contains attributes, like the actual byte code that defines the function, the list of type annotations in the signature, ... and the default values of all arguments:


```pysh
>>> def foo(a: dict = {}):
...  x = a["somekey"]
...  return x
...
>>> foo.__code__
<code object foo at 0x105435710, file "<stdin>", line 1>
>>> foo.__code__.co_argcount
1
>>> foo.__code__.co_nlocals
2
>>> foo.__code__.co_varnames
('a', 'x')
>>> foo.__annotations__
{'a': <class 'dict'>}

>>> foo.__defaults__  # argument default values
({},)
>>> foo()  # This will obviously fail because the default dict is empty
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in foo
KeyError: 'somekey'
>>> foo.__defaults__[0]["somekey"] = 42   # We can modify the default values (never do that kind of f*ckery)
>>> foo()
42
```

mpariente: Ok, so it's defined in `foo.__default__` when the function is defined and never updated, that's perfect ! 
Thanks for the detailed answer.

mpariente:
There is a mismatch between the requirements filenames (test vs dev).

jonashaag:
OK, will fix that notebook. Are you ok with the requirements changes? (No actual changes to requirements except librosa, only moved them to files)

.github/workflows/test_asteroid_package.yml
```diff
@@ -42,7 +41,7 @@ jobs:
           INSTALL="--pre torch torchvision torchaudio -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html"
         fi
         python -m pip install $INSTALL
-        python -m pip install -r requirements.txt --quiet
+        python -m pip install -r requirements_dev.txt Cython --upgrade-strategy only-if-needed --quiet
```
mpariente, line 44: Remove Cython from here

mpariente:
I like the split, but I wonder if we don't make a `requirements` folder to put all the files in, what's your take on that?

jonashaag:
yeah why not, we could also move the docs requirements there:

```
requirements/
  requirements.txt
  requirements_dev.txt
  requirements_test.txt
  requirements_docs.txt
```

Or?

```
requirements/
  default.txt
  dev.txt
  test.txt
  docs.txt
```

mpariente:
Yes. 
Option 2 for me

mpariente:
LGTM
Only librosa to install at the beginning of the notebook right?

requirements/install.txt
```diff
@@ -0,0 +1,10 @@
+# Requirements for using Asteroid
+-r ./torchhub.txt
+PyYAML>=5.0
+pandas>=0.23.4
+pytorch-lightning>=0.8,<0.10.0
```
mpariente, line 5:
```suggestion
pytorch-lightning>=1.0.1
```

jonashaag:
> LGTM
> Only librosa to install at the beginning of the notebook right?

I think so

jonashaag:
When I change the notebook the diff is huge (almost the entire notebook JSON), is this expected? I'll upload the version I wanted to commit

jonashaag:
Stupid GitHub does not allow ipynb

[00_GettingStarted.ipynb.zip](https://github.com/asteroid-team/asteroid/files/5695963/00_GettingStarted.ipynb.zip)

mpariente:
Yes, this is expected.. 
That's why nbdev uses something like nbdiff. 
Did you rerun the notebook?

mpariente:
`librosa` should be installed at the begining of the notebook right?

jonashaag:
I didn't rerun it.

I deliberately moved the `librosa` import to the visualisation part and used `soundfile` for loading the example waveform to show that you don't need `librosa`.

mpariente:
But then it's not installed right for the visualization right? 
The notebook should be ran again IMO

jonashaag:
Ran it again and did some changes to the text regarding Zenodo. https://colab.research.google.com/drive/1_2Dbd0lXcSbaTjzUKtdmFtfV_Y_2QPYX?usp=sharing

mpariente:
Cannot open the colab 

image

mpariente:
Let's fix the tests and merge this.

jonashaag:
Oh no I deleted that Google account recently 🤦‍♂️ Gotta redo the changes

setup.py
```diff
@@ -36,24 +36,24 @@ def find_version(*file_paths):
...
         "requests",
+        "filelock",
+        "SoundFile>=0.10.2",
+        # From requirements/install.txt
+        "PyYAML>=5.0",
+        "pandas>=0.23.4",
+        "pytorch-lightning>=1.0.1",
+        "torchaudio>=0.5.0",
+        "pb_bss_eval>=0.0.2",
+        "torch_stoi>=0.0.1",
```
mpariente, line 53:
```suggestion
        "torch_stoi>=0.1.2",
```

jonashaag:
I think this should be good now

mpariente:
The diffs from notebooks are really weird. We might want to use nbdev for reviewing in the future

jonashaag:
Yeah it’s a mess. There are also some stripping tools to remove stuff like “Out [42]” that may frequently change

PR #394, opened 2020-12-15
Add MetricTracker

mpariente:
Just because it's less code duplicate and it seems useful. 
Inspired by the `WerTracker` in #329.


mpariente:
@JorisCos @jonashaag some feedback?

jonashaag:
VERY quick code review, looks ok, maybe remove the dumping (CSV/JSON) code entirely and let the user to it? It's only an additional `.to_csv()` call to be done by the user

jonashaag:
At some point I will try to think of an integration of this into `System` for live metrics. Right now I have something similar in my recipes

mpariente:
> At some point I will try to think of an integration of this into `System` for live metrics. Right now I have something similar in my recipes

Yes, using Lightning for evaluation would really make things simple, and help us get rid on bash..

mpariente:
Thanks for the review.

mpariente:
Resolved conflicts and clean things up.

We should converge on a format for the output (dataframe, aggregate final results in dictionnary, dump to JSON, autogenerate markdown table).

PR #395, opened 2020-12-19
Remove call to super().training_step()

jonashaag:
The base implementation now throws a warning, see https://github.com/PyTorchLightning/pytorch-lightning/blob/81fd33b431578129de6ea19b99af7da87b1081bf/pytorch_lightning/core/lightning.py#L502


mpariente:
Cool, I've seen the warning few times but didn't look into it !

Issue #404, opened 2021-01-04
Can we had n_filters to BaseDCUNet ?


JorisCos:
## ❓ BaseDCUNet and DCCRN:
In the BaseDCUNet class which is also used for DCCRNet we assume that n_filters=stft_kernel_size. 

```Python
 def __init__(
        self,
        architecture,
        stft_kernel_size=512,
        stft_stride=None,
        sample_rate=16000.0,
        **masknet_kwargs,
    ):
        self.architecture = architecture
        self.stft_kernel_size = stft_kernel_size
        self.stft_stride = stft_stride
        self.masknet_kwargs = masknet_kwargs

        encoder, decoder = make_enc_dec(
            "stft",
            kernel_size=stft_kernel_size,
            n_filters=stft_kernel_size,
            stride=stft_stride,
            sample_rate=sample_rate,
        )
```
In the DCCRN paper it's said
> 3.2. Training setup and baselines
For all of the models, the window length and hop size are 25 msand 6.25 ms, and the FFT length is 512.

To match the paper design, can we add n_filters as an argument of BaseDCUNet ?

jonashaag:
Sure, why not, I'd also `fb_name` (see https://github.com/asteroid-team/asteroid-filterbanks/pull/6)

JorisCos:
Then we should remove the stft from  `self.stft_kernel_size` and `self.stft_stride` and put stft as default for fb_name.

mpariente:
Let's start with adding `stft_n_filters` only because it's backward compatible. We'll see later about adding `fb_name`, because it has other implications (as said by @JorisCos)

jonashaag:
I was just working on fixing the #406 failures when I realised that we're going to break backwards compat there anyways (with the #400 fix). So, two suggestions:

- Only add `stft_n_filters` without breaking anything, and not fixing #400
- Fix #400, also fix DCCRNet to the paper values, and also change filterbank options so that you have free choice of filterbank. The latter we could make backwards compatible because of the `stft_` prefix that we currently have

mpariente:
If you feel like going for option 2, that would be nice !

jonashaag:
Hmm... the models already accept `**masknet_kwargs`, so how are we going to add the filterbank params?

1. Using an extra hierarchy for encoder params, eg. `encoder={"kernel_size": ...}` dictionary
2. Explicitly listing any supported filterbank params (eg. a common subset of params used in most filterbanks)

Other ideas?

[...]

Issue #400, opened 2020-12-29
Proper STFT configuration for DCUNet


jonashaag:
## 🐛 Bug

For DCUNet, we currently use a 512/256 STFT. In the paper they use 1024/256. My own experiments show a large drop of performance for 512 FFT, and a large drop for hop >= 1/2 FFT. I don't see a performance increase with 2048/512.

I suggest to change the default to 1024/256.

Note that my tests were done with `torch_stft_fb`.

mpariente:
Yes, let's use the defaults as in the paper, sure.

mpariente:
Fixed by #406

PR #406, opened 2021-01-05
Add stft_n_filters to BaseDCUNet

JorisCos:
This PR adds `stft_n_filters` as mentionned in #404.
It also fixes the defaults parameter of DCUNet to match its paper as mentionned in #400

### DIFF
--- a/tests/models/models_test.py
+++ b/tests/models/models_test.py
@@ -155,8 +155,8 @@ def test_dptnet(fb):
 
 
 def test_dcunet():
-    _, istft = make_enc_dec("stft", 512, 512)
-    input_samples = istft(torch.zeros((514, 17))).shape[0]
+    _, istft = make_enc_dec("stft", 1024, 1024, 256, 16000)
+    input_samples = istft(torch.zeros((1026, 17))).shape[0]
     _default_test_model(DCUNet("DCUNet-10"), input_samples=input_samples)
     _default_test_model(DCUNet("DCUNet-10", n_src=2), input_samples=input_samples)
 
@@ -173,7 +173,7 @@ def test_dcunet():
 
 
 def test_dccrnet():
-    _, istft = make_enc_dec("stft", 512, 512)
+    _, istft = make_enc_dec("stft", 512, 400, 100,16000)
     input_samples = istft(torch.zeros((514, 16))).shape[0]
     _default_test_model(DCCRNet("DCCRN-CL"), input_samples=input_samples)
     _default_test_model(DCCRNet("DCCRN-CL", n_src=2), input_samples=input_samples)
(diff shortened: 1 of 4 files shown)


mpariente:
There is a problem with CI (1.7.0 build) that might be related to the stride (changed input shape), could someone have a look? 

The other STOI problem on the nightly build _should_ be fixed now.

JorisCos:
We can have a closer look to what's wrong with DCCRNet @jonashaag

mpariente:
Could you explain the fix please? What was wrong?

JorisCos:
Sure, I added `stft_n_filters` in DCUNet and DCCRNet. I also changed the default values of `stft_n_filters` `stft_n_kernel_size` and `stft_stride`. In the tests the models are initialized with the default values and the tests use hard-coded values tied to those default parameters, so they needed to be changed.

JorisCos:
At first, I was confused by the jit tests on DCCRNet that kept failing but it was because I didn't changed this ligne 
` masknet_kwargs.setdefault("n_freqs", stft_n_filters // 2)` .

JorisCos:
Before, because ` stft_n_filters` was equal to ` stft_kernel_size` ` masknet_kwargs.setdefault("n_freqs", stft_kernel_size // 2)` was ok but it had to be changed

mpariente:
Makes complete sense, thanks for the explanation!

tests/models/models_test.py
```diff
@@ -155,8 +155,8 @@ def test_dptnet(fb):
 
 
 def test_dcunet():
-    _, istft = make_enc_dec("stft", 512, 512)
-    input_samples = istft(torch.zeros((514, 17))).shape[0]
+    _, istft = make_enc_dec("stft", 1024, 1024, 256, 16000)
+    input_samples = istft(torch.zeros((1026, 17))).shape[0]
```
mpariente, line 159: It took time to find out what the bug was because those tests were not clear enough. 
Let's try to improve it: 
- Pass everything to `make_enc_dec` as kwargs, this is unclear currently.
- Add a `n_fft` variable, and reference `n_fft + 2` in the `istft` to be clear.

tests/models/models_test.py
```diff
@@ -173,7 +173,7 @@ def test_dcunet():
 
 
 def test_dccrnet():
-    _, istft = make_enc_dec("stft", 512, 512)
+    _, istft = make_enc_dec("stft", 512, 400, 100, 16000)
```
mpariente, line 176: Same here, name kwargs + introduce `n_fft`.

mpariente (commented):
After the required edits on the tests, we'll merge that, thanks again

mpariente:
Thanks again !

Issue #401, opened 2021-01-01
Allow https:// URL for Hugging Face download


jonashaag:
## 🚀 Feature

Currently when you just copy and paste a HF hosted model URL from the browser (eg. https://huggingface.co/mhu-coder/ConvTasNet_Libri1Mix_enhsingle) to `asteroid-infer`, you this error:

```
Traceback (most recent call last):
  File "main.py", line 27, in <module>
    model = BaseModel.from_pretrained(model_id)
  File "/usr/local/lib/python3.8/dist-packages/asteroid/models/base_models.py", line 122, in from_pretrained
    conf = torch.load(cached_model, map_location="cpu")
  File "/usr/local/lib/python3.8/dist-packages/torch/serialization.py", line 595, in load
    return _legacy_load(opened_file, map_location, pickle_module, **pickle_load_args)
  File "/usr/local/lib/python3.8/dist-packages/torch/serialization.py", line 764, in _legacy_load
    magic_number = pickle_module.load(f, **pickle_load_args)
_pickle.UnpicklingError: invalid load key, '<'.
```

Probably because it downloads the page as HTML and then tries to parse that as pickle file.

### What you'd like

Recognise HF urls like https://huggingface.co/mhu-coder/ConvTasNet_Libri1Mix_enhsingle and have them behave the same as without the https://huggingface.co/ prefix

cc @julien-c

mpariente:
That would be great indeed ! 
Do you feel like having a look?

jonashaag:
@julien-c what are your thoughts on adding support for `https://huggingface.co/user/model` to `hf_hub_url` in the Hub library?

mpariente:
Let's first migrate to the new library https://github.com/huggingface/huggingface_hub, and then we'll have a look at how we can do that.

jonashaag:
I can also submit a PR to the integration library if you agree with that feature

PR #409, opened 2021-01-07
Use huggingface_hub dependency, remove inlined HF code

jonashaag:
Refs #401

Switch to huggingface_hub and allow for https://huggingface.co/ prefixed URLs. 

We can also merge without the URL change if @julien-c doesn't like it.

CI currently fails due to missing dependency bump

### DIFF
--- a/asteroid/utils/hub_utils.py
+++ b/asteroid/utils/hub_utils.py
@@ -1,19 +1,12 @@
-import fnmatch
-import io
-import json
 import os
-import sys
-import tempfile
-from contextlib import contextmanager
-from functools import partial, lru_cache
+from functools import lru_cache
 from hashlib import sha256
-from typing import BinaryIO, Dict, Optional, Union, List
-from urllib.parse import urlparse
+from typing import Union, Dict, List
 
 import requests
-import torch
-from filelock import FileLock
 from torch import hub
+import huggingface_hub
+import huggingface_hub.file_download
 
 
 CACHE_DIR = os.getenv(
@@ -36,10 +29,6 @@ MODELS_URLS_HASHTABLE = {
 
 SR_HASHTABLE = {k: 8000.0 if not "DeMask" in k else 16000.0 for k in MODELS_URLS_HASHTABLE}
 
-HF_WEIGHTS_NAME = "pytorch_model.bin"
-HUGGINGFACE_CO_PREFIX = "https://huggingface.co/{model_id}/resolve/{revision}/{filename}"
-ENDPOINT = "https://huggingface.co"
-
 
 def cached_download(filename_or_url):
     """Download from URL and cache the result in ASTEROID_CACHE.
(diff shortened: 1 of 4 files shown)


mpariente:
Thanks ! 
I like the URL change, it makes sense to keep it for me. 
Let's add the deps so that CI passes. 

We might be able to remove some of the hub dependencies thanks to this. 
We might want to merge #386 before this PR, right?

asteroid/utils/hub_utils.py
```diff
@@ -71,8 +62,10 @@ def cached_download(filename_or_url):
         else:
             model_id = filename_or_url
             revision = None
-        url = hf_bucket_url(model_id=model_id, filename=HF_WEIGHTS_NAME, revision=revision)
-        return hf_get_from_cache(url, cache_dir=get_cache_dir())
+        url = huggingface_hub.hf_hub_url(
+            model_id, filename=huggingface_hub.file_download.PYTORCH_WEIGHTS_NAME, revision=revision
```
julien-c, line 66: I can expose `PYTORCH_WEIGHTS_NAME` in the library's init.py so that you don't have to import from a nested file

asteroid/utils/hub_utils.py
```diff
@@ -57,7 +46,9 @@ def cached_download(filename_or_url):
     if os.path.isfile(filename_or_url):
         return filename_or_url
 
-    if urlparse(filename_or_url).scheme in ("http", "https"):
+    if filename_or_url.startswith(("http://", "https://")) and not filename_or_url.startswith(
+        huggingface_hub.file_download.HUGGINGFACE_CO_REPO_URL_BASE
```
julien-c, line 50: I would just add a `https://huggingface.co/` constant in this local file as this is unlikely to ever change:)

julien-c: (and I would add the prefix-stripping here, unless you feel strongly about it!)

asteroid/utils/hub_utils.py
jonashaag: That would be great.

asteroid/utils/hub_utils.py
jonashaag: No strong feelings, we can do it here.

mpariente:
Let's address the comments from Julien and merge this? 

Your PR in the official repo has been merged right? So we can have the good pip deps?

jonashaag:
No, we need to move the URL stripping here, see https://github.com/huggingface/huggingface_hub/pull/8. And create a PR for `PYTORCH_WEIGHTS_NAME`

julien-c:
Sorry about the looong delay!

I've opened https://github.com/asteroid-team/asteroid/pull/440 on top of this PR, to make code simpler and re-add the hf.co url prefix trimming.

Issue #402, opened 2021-01-03
fix_length_mode="trim" can lead to confusing loss/metrics


jonashaag:
When using `fix_length_mode="trim"` in DCUNet, the signal is right-trimmed to the next possible size, and then zero-padded on the right to reconstruct the original size. When using a loss or metric that does not know about the actual input size to the model, the loss/metric values can be way off, since essentially a chunk of the expected signal is zeroed in the estimate. (I don't think it hurts training though.)

What to do? Maybe just add note to the docs?

mpariente:
A note in the docs seems good. 

Also maybe exposing a function to compute the length would make sense so that the metric can be computed on the non-zero part of the signal?

jonashaag:
I did a patch that adds the doc hint and the function. I realised the function may be generally useful, even to non-STFT models, to return the reconstructed size (without the `pad_x_to_y` reconstruction done in `BaseEncoderMaskerDecoder.forward()`). Do you think we should move it to `BaseEncoderMaskerDecoder`, and also move the docs hint there?

mpariente:
Can you open a PR with this patch please? And we'll move the discussion there?

PR #410, opened 2021-01-08
Warning about masker internal working size

jonashaag:
Refs #402

Open questions:

- Is getting the masker "working size" useful for all filterbanks? (I think so but not sure)
- The code requires the `fix_input_dims` method that's only defined for DCUNet right now, shall we add it to the base EMD model? (Are we even able to define this for the majority of models?)

The implementation is very naive (stupid/inefficient) but should work well for all maskers.

### DIFF
--- a/asteroid/masknn/convolutional.py
+++ b/asteroid/masknn/convolutional.py
@@ -486,7 +486,20 @@ class DCUMaskNet(BaseDCUMaskNet):
     Input shape is expected to be $(batch, nfreqs, time)$, with $nfreqs - 1$ divisible
     by $f_0 * f_1 * ... * f_N$ where $f_k$ are the frequency strides of the encoders,
     and $time - 1$ is divisible by $t_0 * t_1 * ... * t_N$ where $t_N$ are the time
-    strides of the encoders.
+    strides of the encoders. If `fix_length_mode` is not `None`, the time dimension
+    may is automatically padded or trimmed to a valid size before running it through
+    the network.
+
+    .. note::
+       If using `fix_length_mode="trim"`, the network's output will be all-zero at the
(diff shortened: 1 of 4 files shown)


mpariente:
> Is getting the masker "working size" useful for all filterbanks? 

Seems so, yes.

> The code requires the fix_input_dims method that's only defined for DCUNet right now, shall we add it to the base EMD model? (Are we even able to define this for the majority of models?)

I'm unsure about adding in the EMD model. Would this mean adding it to the forward as well? 

For most models, `fix_input_dims` would just be identity.

jonashaag:
> Would this mean adding it to the forward as well?

Yes. I don't see another way to do it if we want that feature for all models. We have to have a way of finding the internal size, so without any shape fixing. Maybe we can add an optional argument to the masker's forward that controls whether fixing should be applied, and then we could use introspection to check if the masker supports that argument, but that doesn't seem to have any advantages. If you have other ideas, please tell :)

jonashaag:
Another way could be to just add an optional-to-define method `.get_working_size()` on the masker, and if it's defined, `BaseEncoderMaskerDecoder.get_masker_working_size()` could use it.

Another crazy idea is to have a mode ("training mode") that just doesn't re-pad anything?

mpariente:
> Another crazy idea is to have a mode ("training mode") that just doesn't re-pad anything? (Both in the masker and models)

So that the loss is only computed on non-zero segment?

jonashaag:
> So that the loss is only computed on non-zero segment?

Well to be exact, so that the loss is computed on "everything the model did produce", which is different from non-zero, generally speaking

mpariente:
Yes, you're right

mpariente:
I don't feel that adding it to the `forward` is the right way to do.

The `get_output_size` option on the masker seems better to me.

PR #414, opened 2021-01-21
Wer tracker

JorisCos:
# About this PR

- This PR  makes it possible to keep track of the transcriptions made by the ASR models.

The file containing the transcriptions is a `.json` that looks like this :
all_trans_ex

- This PR also adds jiwer transformation to the measure computation. It removes the punctuation and puts everything to lowercase. This leads to a more accurate WER.

### DIFF
--- a/asteroid/metrics.py
+++ b/asteroid/metrics.py
@@ -176,11 +176,11 @@ class WERTracker:
         txt = self.predict_hypothesis(mix)
 
         # Dict to gather transcriptions and IDs
-        trans_dict = dict(mixture_txt={},clean={},estimates={},truth={})
+        trans_dict = dict(mixture_txt={}, clean={}, estimates={}, truth={})
         # Get mixture transcription
         trans_dict["mixture_txt"] = txt
         #  Get ground truth transcription and IDs
-        for i,tmp_id in enumerate(wav_id):
+        for i, tmp_id in enumerate(wav_id):
             trans_dict["truth"][f"ID_{i}"] = tmp_id
             trans_dict["truth"][f"Txt_{i}"] = self.trans_dic[tmp_id]
             self.true_txt_list.append(dict(utt_id=tmp_id, text=self.trans_dic[tmp_id]))
@@ -224,14 +224,15 @@ class WERTracker:
     def hsdi(truth, hypothesis):
         import jiwer
         from jiwer import compute_measures
-        transformation = jiwer.Compose([
-            jiwer.ToLowerCase(),
-            jiwer.RemovePunctuation()
-        ])
+
+        transformation = jiwer.Compose([jiwer.ToLowerCase(), jiwer.RemovePunctuation()])
         keep = ["hits", "substitutions", "deletions", "insertions"]
-        out = compute_measures(truth=truth, hypothesis=hypothesis,
-                               truth_transform=transformation,
-                               hypothesis_transform=transformation).items()
+        out = compute_measures(
+            truth=truth,
+            hypothesis=hypothesis,
+            truth_transform=transformation,
+            hypothesis_transform=transformation,
+        ).items()
         return {k: v for k, v in out if k in keep}
 
     def predict_hypothesis(self, wav):
--- a/asteroid/metrics.py
+++ b/asteroid/metrics.py
@@ -155,7 +155,14 @@ class WERTracker:
         self.mix_counter = Counter()
         self.clean_counter = Counter()
         self.est_counter = Counter()
-        self.transformation = jiwer.Compose([jiwer.ToLowerCase(), jiwer.RemovePunctuation()])
+        self.transformation = jiwer.Compose(
+            [jiwer.ToLowerCase(),
+             jiwer.RemovePunctuation(),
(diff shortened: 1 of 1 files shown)


asteroid/metrics.py
```diff
@@ -146,6 +146,8 @@ def __init__(self, model_name, trans_df):
         self.input_txt_list = []
         self.clean_txt_list = []
         self.output_txt_list = []
+        self.transcription = []
```
mpariente, line 149: Should be in plural right?

asteroid/metrics.py
```diff
@@ -172,25 +174,41 @@ def __call__(
...
+            self.true_txt_list.append(dict(utt_id=tmp_id, text=self.trans_dic[tmp_id]))
+        # Mixture
         for tmp_id in wav_id:
             out_count = Counter(self.hsdi(truth=self.trans_dic[tmp_id], hypothesis=txt))
             self.mix_counter += out_count
             local_mix_counter += out_count
             self.input_txt_list.append(dict(utt_id=tmp_id, text=txt))
         # Average WER for the clean pair
-        for wav, tmp_id in zip(clean, wav_id):
+        for i, (wav, tmp_id) in enumerate(zip(clean, wav_id)):
```
mpariente, line 194: That's a bit ugly and probably unclear for researchers. Not sure.

asteroid/metrics.py
```diff
@@ -204,10 +222,17 @@ def wer_from_hsdi(hits=0, substitutions=0, deletions=0, insertions=0):
 
     @staticmethod
     def hsdi(truth, hypothesis):
+        import jiwer
         from jiwer import compute_measures
 
+        transformation = jiwer.Compose([jiwer.ToLowerCase(), jiwer.RemovePunctuation()])
```
mpariente, line 228: Should this be defined in the constructor (init) instead?

mpariente (commented):
The estimates seem a bit off :joy: but this was clearly missing, thanks!

asteroid/metrics.py
JorisCos: It should :)

JorisCos:
Haha the estimates are wrong because the model was trained on noisy inputs and the enhancement is making things worse...

asteroid/metrics.py
JorisCos: Well, if we do that we have to make `self.transformation` an argument of `hsdi()` or `hsdi()` can stay a static method

mpariente: Let's keep hsdi a static function and pass the transformation.

mpariente:
Let's include the wav normalization in the WERTracker class?

popcornell:
What is the ID field in .json annotation ?

JorisCos:
> Let's include the wav normalization in the WERTracker class?

We can but we have to it anyway before saving the files.


> What is the ID field in .json annotation ?

In a general way the ID is something we introduced for librimix to match transcriptions and wav files . For this specific screen shot this annotation and ID's are taken from CHIME 4. ( I will open a PR soon)

popcornell:
`In a general way the ID is something we introduced for librimix to match transcriptions and wav files . For this specific screen shot this annotation and ID's are taken from CHIME 4. ( I will open a PR soon)`

Maybe call em UtteranceID or ExampleID?  Because we might need also speaker IDs

mpariente:
Also, please make the fields in JSON all lower case: "text_0", "utt_id_0" etc...

asteroid/metrics.py
```diff
@@ -139,19 +139,23 @@ def __init__(self, model_name, trans_df):
...
         self.output_txt_list = []
+        self.transcriptions = []
+        self.true_txt_list = []
         self.sample_rate = int(d.data_frame[d.data_frame["name"] == model_name]["fs"])
         self.trans_df = trans_df
         self.trans_dic = self._df_to_dict(trans_df)
         self.mix_counter = Counter()
         self.clean_counter = Counter()
         self.est_counter = Counter()
+        self.transformation = jiwer.Compose([jiwer.ToLowerCase(), jiwer.RemovePunctuation()])
```
mpariente, line 158: Is this transformation enough? 
The default is 
```
[<jiwer.transforms.RemoveMultipleSpaces at 0x7fbc79a75df0>,
 <jiwer.transforms.Strip at 0x7fbc79a75e20>,
 <jiwer.transforms.SentencesToListOfWords at 0x7fbc79a75f10>,
 <jiwer.transforms.RemoveEmptyStrings at 0x7fbc7aa17bb0>]
```

asteroid/metrics.py
```diff
@@ -258,5 +295,8 @@ def final_df(self):
         )
         return df
 
+    def all_transcriptions(self):
+        return dict(transcriptions=self.transcriptions)
```
mpariente, line 299: I don't really see the point of the dict with one field, returning the list. 
I'd remove this method entirely

asteroid/metrics.py
JorisCos: When I tested on CHIME4 these were the two that made a difference but you are right let's add the others. It doesn't cost that much anyway.

PR #415, opened 2021-01-21
Mulcat option, ran black

JunzheJosephZhu:


JunzheJosephZhu:
It looks like I accidentally also included unfinished code from recipe in my last commit.

mpariente:
Yes, you shouldn't include everything. 
We can start a new PR if you want.

PR #416, opened 2021-01-21
Mulcat option

JunzheJosephZhu:


mpariente:
Great this time, thanks !

mpariente:
@JorisCos do you mind reviewing please?

asteroid/masknn/recurrent.py
```diff
@@ -196,22 +258,31 @@ def __init__(
...
+            )
         # IntraRNN and linear projection layer (always bi-directional)
-        self.intra_RNN = SingleRNN(
-            rnn_type, in_chan, hid_size, num_layers, dropout=dropout, bidirectional=True
-        )
-        self.intra_linear = nn.Linear(hid_size * 2, in_chan)
+        self.intra_linear = nn.Linear(self.intra_RNN.output_size, in_chan)
         self.intra_norm = norms.get(norm_type)(in_chan)
+
         # InterRNN block and linear projection layer (uni or bi-directional)
```
JorisCos, line 293: Same here

asteroid/masknn/recurrent.py
```diff
@@ -196,22 +258,31 @@ def __init__(
...
+            rnn_type, in_chan, hid_size, num_layers, dropout=dropout, bidirectional=bidirectional
+            )
+        else:
+            self.intra_RNN = SingleRNN(
+                rnn_type, in_chan, hid_size, num_layers, dropout=dropout, bidirectional=True
+            )
+            self.inter_RNN = SingleRNN(
+            rnn_type, in_chan, hid_size, num_layers, dropout=dropout, bidirectional=bidirectional
+            )
         # IntraRNN and linear projection layer (always bi-directional)
```
JorisCos, line 289: This comment should be before `  self.intra_RNN = DoubleRNN(...)`

JorisCos (commented):
Thanks for the PR, this looks really nice. A minor fix on the docs is needed otherwise LGTM

asteroid/masknn/recurrent.py
```diff
@@ -62,6 +62,67 @@ def forward(self, inp):
         return rnn_output
 
 
+class DoubleRNN(nn.Module):
+    """Module for a RNN block.
+
+    Inspired from https://github.com/yluo42/TAC/blob/master/utility/models.py
+    Licensed under CC BY-NC-SA 3.0 US.
```
mpariente, line 69:
```suggestion
```

mpariente: I don't think it's inspired by this repo right?

asteroid/masknn/recurrent.py
```diff
@@ -62,6 +62,67 @@ def forward(self, inp):
         return rnn_output
 
 
+class DoubleRNN(nn.Module):
+    """Module for a RNN block.
```
mpariente, line 66: Need to be more explicit.

mpariente (commented):
Let's think about a way which doesn't copy paste the `SingleRNN` code in `DoubleRNN`.

mpariente:
@JunzheJosephZhu ?

JunzheJosephZhu:
I guess the best way to do it would be combining them into s single object, adding an option for single/doubleRNN and adding a switch statement into the code?

For example,
class singleRNN(...):
   def init(...,...,mulcat=False)
   If mulcat:
       self.inter_RNN = ...
   else:
       self.inter_RNN1 =
       self.inter_RNN2 =

The code would be somewhat more confusing to readers though

asteroid/masknn/recurrent.py
```diff
@@ -62,6 +62,64 @@ def forward(self, inp):
...
+        return self.hidden_size * (2 if self.bidirectional else 1) + self.input_size
+
+    def forward(self, inp):
+        """ Input shape [batch, seq, feats] """
+        self.rnn1.flatten_parameters()  # Enables faster multi-GPU training.
+        self.rnn2.flatten_parameters()  # Enables faster multi-GPU training.
+        output = inp
+        rnn_output1, _ = self.rnn1(output)
+        rnn_output2, _ = self.rnn2(output)
+        return torch.cat((rnn_output1 * rnn_output2, output), -1)
```
mpariente, line 120:
```suggestion
        rnn_output1, _ = self.rnn1(inp)
        rnn_output2, _ = self.rnn2(inp)
        return torch.cat((rnn_output1 * rnn_output2, inp), -1)
```

mpariente:
Let's keep the code like this for now. 
Please
- Rename `DoubleRNN` to `MulCatRNN`
- Add a description in the docstring. 
- Test the `mul_cat` option in jit_test.py

JunzheJosephZhu:
It looks like when running the jit test, sudormrf always fails the test. Is this something I need to worry about, or is this unrelated to DPRNN?

asteroid/masknn/recurrent.py
```diff
@@ -62,6 +62,63 @@ def forward(self, inp):
         return rnn_output
 
 
+class MulCatRNN(nn.Module):
+    """Module for a MulCat RNN block.
```
mpariente, line 66:
```suggestion
    """MulCat RNN block from [1].
    
    Composed of two RNNs, returns ``cat([RNN_1(x) * RNN_2(x), x])``.
```

mpariente: Need to add reference, as done in other models.

tests/jit/jit_models_test.py
```diff
@@ -106,7 +107,7 @@ def small_model_params():
     return params
 
 
-@pytest.mark.parametrize("model_def", [DCCRNet, DeMask])
+@pytest.mark.parametrize("model_def", [DPRNNTasNet])
```
mpariente, line 110: That doesn't look right. Why did you do that?

mpariente (commented):
We are getting much closer, thanks ! What's left

- [ ] Add `use_mulcat` is `get_config` so that serialization will work
- [ ] Add the `use_mulcat` parameterize in `tests/models/models_test.py` to ensure serialization will work. 
- [ ] Fix the jit tests which don't look right to me. 
- [ ] Add the reference to the original MulCat paper in `MulCatRNN`

tests/jit/jit_models_test.py
JunzheJosephZhu: I think we just add whatever model we just want to test with jit here, no?

JunzheJosephZhu:
Also, do you want to implement multi-stage loss? That was also in the facebook paper.

tests/jit/jit_models_test.py
mpariente: We need to keep all the tests, otherwise we don't know when we break things.

JunzheJosephZhu: I added them back

mpariente:
Eventually, yes, but in the recipe sounds fine. 

> since I'm gonna re-write the backbone anyway.

What do you mean?

JunzheJosephZhu:
I mean I will use DPRNNBlock but not DPRNN for my separator, since the part starting from Conv2D diverge.

mpariente:
Alright. 
Can you fix the other issues I mentioned so that we can merge?

JunzheJosephZhu:
Ok, I'll do it ASAP. Btw, how do you write test cases for dataset objects?

mpariente:
We don't have tests for datasets.. Which is a large limitation.

JunzheJosephZhu:
I just finished my work with variable speaker wsj0mix dataset in my recipe folder. Should I start a PR yet?

tests/models/models_test.py
```diff
@@ -109,7 +109,9 @@ def test_save_and_load_convtasnet(fb, sample_rate):
 
 
 def test_dprnntasnet_sep():
-    nnet = DPRNNTasNet(n_src=2, n_repeats=2, bn_chan=16, hid_size=4, chunk_size=20, n_filters=32)
+    nnet = DPRNNTasNet(
+        n_src=2, n_repeats=2, bn_chan=16, hid_size=4, chunk_size=20, n_filters=32, use_mulcat=True
```
mpariente, line 113: It's not what I meant. 
I mean to use the `@pytest.mark.parameterize` to test both. 
I'll take care of it.

JunzheJosephZhu:
I have no clue why it's not passing the test_save_and_load_dprnn. Do you have any idea?

mpariente:
The loading is failing. 

We need to test the mulcat version, but still keep the old tests. For that, we can use `@pytest.mark.parameterize`. 

Regarding the recipe, yes, you can make a PR :smiley:  Thanks

mpariente:
Because `use_mulcat` is not returned in `get_config` (as I mentioned before) so the new model doesn't use it.

JunzheJosephZhu:
Where do I find get_config?

JunzheJosephZhu:
Hi, I'm starting to write the model class. I noticed that in all your models that are in the source code just use make_optimizer, but those in the recipe define make_model_and_optimizer. Do I have to follow this?

mpariente:
No, you don't have to. 
The most important is that loading the model for eval is easy.

JunzheJosephZhu:
how do i write test cases for models in my recipe?
Also, github send me an email everyday telling me the version consistency check failed. I have no clue what it means

mpariente:
> how do i write test cases for models in my recipe?

They aren't tested if they are in the recipe folder. 

> Also, github send me an email everyday telling me the version consistency check failed. I have no clue what it means

Sorry about that, I don't receive those e-mails.. Anybody else receives those e-mails?

asteroid/masknn/recurrent.py
```diff
@@ -196,22 +257,51 @@ def __init__(
...
+                bidirectional=bidirectional,
+            )
+        else:
+            self.intra_RNN = SingleRNN(
+                rnn_type,
+                in_chan,
+                hid_size,
+                num_layers,
+                dropout=dropout,
+                bidirectional=bidirectional,
```
mpariente, line 291:
```suggestion
                 bidirectional=True,
```

mpariente:
I
- Fixed the jit tests
- Parameterize model test with `use_mulcat`
- Restore always bidirectional IntraRNN
- Removed the recipe files (which should be added to another branch). 

This is ready for merging on my side. 
Next time, don't make PRs from your master branch please :wink:

JunzheJosephZhu:
---------- Forwarded message ---------

mpariente:
Not sure..

PR #421, opened 2021-02-02
Add linting workflow

mpariente:
Because nobody installs the pre-commit hooks.


mpariente:
Ok, we won't have permission to push to forks, which is where the linting doesn't happen, so this is useless.

jonashaag:
Even if the forker enables "allow modifications from maintainer"?

mpariente:
I didn't work for other actions.. Let's try but I guess it won't work.

Issue #420, opened 2021-02-02
[Design] Multichannel model support


mpariente:
### What we want 
- Suport in `BaseModel` for save/load/share
- Support in `separate` for the inference API + CLI.

### Proposed design
- Add `channels` (or `n_chan`, `n_channels`?) to `BaseModel`.
- Add `channels` to `separate` and `Separable`. 

Is there anywhere else where we need to change it? 

### Questions: 
- Should `channels` be handle the same way as `sample_rate` (hidden attribute + property)? 
- How should we handle models that support several number of channels? Using `List` or `None`? 
- Should we raise errors in `separate` if the number of input channels is different than the expected number? Or remove the channels? Or apply the network to each channel? etc...

popcornell:
Maybe mic_channels is better ? 

> 
> Should channels be handle the same way as sample_rate (hidden attribute + property)?
> How should we handle models that support several number of channels? Using List or None?

FasNet TAC can handle an arbitrary amount of channels actually.

mpariente:
> What do you mean by list ? empty list ?

If a method works with 1, 2 or 3 mics but not more. But actually, in #427, I propose either `int` or `None`. If `None`, we don't check anything. 

> BTW now is there a way to apply easily a single-channel system to multiple channel audio ?

No, there is not. 
Do you think we should have it in the CLI? 
Because otherwise the user can loop over the channels, I guess it's pretty simple. 
```python
torch.stack([model.separate(w[None]) for w in wav]) 
```


> Maybe mic_channels is better ?

I opted for `n_channels` in the PR but we can discuss it. 

Thanks for the feedback

jonashaag:
> If a method works with 1, 2 or 3 mics but not more. But actually, in #427, I propose either `int` or `None`. If `None`, we don't check anything.

Yes, let's keep it simple for now, as you suggested.

> > BTW now is there a way to apply easily a single-channel system to multiple channel audio ?
> 
> No, there is not.
> Do you think we should have it in the CLI?

I think this could be useful and easy to implement. We can start by supporting only input channels >= 1, and model channels = 1, and add a CLI flag that applies to model to all the channels separately.

> > Maybe mic_channels is better ?
> 
> I opted for `n_channels` in the PR but we can discuss it.

I have seen people confused about "channels" and "sources" etc multiple times. I think anyone not familiar with source separation is quick to think of "source" = "my audio files". How about naming the both simply "input_channels" and "output_channels"?

[...]

PR #427, opened 2021-02-02
[src] Multichannel models 

mpariente:
Started implementing what I suggested in #420. 

I'll highlight the places where I doubt in the PR. 

Funny things I found along the way: 
- It's hard to make a model that doesn't take the sample rate as argument 
```python
class Model(BaseModel):
    def __init__(self):
        super().__init__(sample_rate=44100)
        
    def forward(self, x):
        return x

    def get_model_args(self):
        return {}
```
This can be serialized, but not loaded because `sample_rate` is needed in the model conf to load the model. 

OTOH:
```python
class Model(BaseModel):
    def __init__(self):
        super().__init__(sample_rate=44100)

    def forward(self, x):
        return x

    def get_model_args(self):
        return {"sample_rate": self.sample_rate}
```
This fails at loading, because it tries to pass `sample_rate` as kwargs, but cannot. 
Should we do something about this?

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -1,5 +1,6 @@
 import torch
 import warnings
+from typing import Optional
 
 from .. import separate
 from ..masknn import activations
@@ -22,18 +23,21 @@ def _unsqueeze_to_3d(x):
 class BaseModel(torch.nn.Module):
     """Base class for serializable models.
 
-    Defines saving/loading procedures as well as separation methods from
-    file, torch tensors and numpy arrays.
-    Need to overwrite the `forward` method, the `sample_rate` property and
-    the `get_model_args` method.
+    Defines saving/loading procedures, and separation interface to `separate`.
+    Need to overwrite the `forward` and `get_model_args` methods.
 
     Models inheriting from `BaseModel` can be used by :mod:`asteroid.separate`
     and by the `asteroid-infer` CLI. For models whose `forward` doesn't go from
     waveform to waveform tensors, overwrite `forward_wav` to return
     waveform tensors.
+
+    Args:
+        sample_rate (float): Operating sample rate of the model.
+        n_channels: Supported number of channels of the model.
+            If None, no checks will be performed.
     """
 
-    def __init__(self, sample_rate: float = 8000.0, n_channels: int = 1):
+    def __init__(self, sample_rate: float = 8000.0, n_channels: Optional[int] = 1):
         super().__init__()
         self.__sample_rate = sample_rate
         self.n_channels = n_channels
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -179,6 +179,11 @@ class BaseModel(torch.nn.Module):
             state_dict=self.get_state_dict(),
             model_args=self.get_model_args(),
         )
+        # FIXME
+        # if hasattr(self, "sample_rate") and "sample_rate" not in model_conf["model_args"]:
+        #     model_conf["model_args"]["sample_rate"] = self.sample_rate
+        # if hasattr(self, "n_channels") and "n_channels" not in model_conf["model_args"]:
+        #     model_conf["model_args"]["n_channels"] = self.n_channels
         # Additional infos
         infos = dict()
         infos["software_versions"] = dict(
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -179,11 +179,6 @@ class BaseModel(torch.nn.Module):
             state_dict=self.get_state_dict(),
             model_args=self.get_model_args(),
         )
-        # FIXME
-        # if hasattr(self, "sample_rate") and "sample_rate" not in model_conf["model_args"]:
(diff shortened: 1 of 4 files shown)


asteroid/models/base_models.py
```diff
@@ -22,20 +23,24 @@ def _unsqueeze_to_3d(x):
...
+        sample_rate (float): Operating sample rate of the model.
+        n_channels: Supported number of channels of the model.
+            If None, no checks will be performed.
     """
 
-    def __init__(self, sample_rate: float = 8000.0):
+    def __init__(self, sample_rate: float = 8000.0, n_channels: Optional[int] = 1):
         super().__init__()
         self.__sample_rate = sample_rate
+        self.n_channels = n_channels
```
mpariente, line 43: I don't think this needs to be a property as the `sample_rate` but we could do it.

asteroid/separate.py
```diff
@@ -88,6 +90,11 @@ def separate(
 @torch.no_grad()
 def torch_separate(model: Separatable, wav: torch.Tensor, **kwargs) -> torch.Tensor:
     """Core logic of `separate`."""
+    if model.n_channels is not None and wav.shape[-2] != model.n_channels:
+        raise RuntimeError(
+            f"Model supports {model.n_channels}-channel inputs but found audio with {wav.shape[-2]} channels."
+            f"Please match the number of channels."
+        )
```
mpariente, line 98: Should we add a flag to ignore that, if passed, make us ignore that and take the first channels or something? 
Something that would be passed from the CLI to here (`--ignore-channels-check`).
I'm not sure it's useful.

asteroid/separate.py
```diff
@@ -159,7 +166,7 @@ def file_separate(
             f"of {model.sample_rate}Hz. You can pass `resample=True` to resample automatically."
         )
     # Pass wav as [batch, n_chan, time]; here: [1, 1, time]
```
mpariente, line 169:
```suggestion
    # Pass wav as [batch, n_chan, time]; here: [1, chan, time]
```

tests/models/models_test.py
```diff
@@ -32,6 +32,24 @@ def test_set_sample_rate_raises_warning():
         model.sample_rate = 16000.0
 
 
+def test_multichannel_model_loading():
+    class MCModel(BaseModel):
+        def __init__(self, sample_rate=8000.0, n_channels=2):
```
mpariente, line 37: As said in the PR. We must have the `sample_rate` as argument. 

If we want fixed number of channels, we don't need it in the `__init__`, we just `super(n_channels=2)` and that works.

mpariente:
Another note: there will be several scenarios for `LambdaOverlapAdd`:
- Multichannel input, single output (should work fine)
- Multichannel input, multichannel output (not sure it works). 

We should add some tests and if the second case doesn't work, raise a useful error.

asteroid/models/base_models.py
popcornell: why not ? some models are tied to the number of channels and actually to the array topology

mpariente: For sample_rate, we made it like that because the model holds reference to the sample_rate, but the filterbank as well, so we wanted the raise the warning when setting it. 

But for the number of channels, for now nothing holds reference to it. 
If we see it's a limitation in the future we can always write a setter/getter. 

Asteroid, as Python, is for consenting adults :wink:

jonashaag:
> This can be serialized, but not loaded because `sample_rate` is needed in the model conf to load the model.

How about moving the check for missing sample rate to *after* the model object has been constructed, and then checking using `hasattr(model, "sample_rate")`?

asteroid/separate.py
jonashaag: Not sure...

mpariente:
> > This can be serialized, but not loaded because `sample_rate` is needed in the model conf to load the model.
> 
> How about moving the check for missing sample rate to _after_ the model object has been constructed, and then checking using `hasattr(model, "sample_rate")`? That way you are free to set the `sample_rate` property however you like as long as it's present.

I thought about that. 
That seems fine to me, let's do that.

asteroid/separate.py
mpariente: @popcornell what's your opinion on that? 
Maybe we can start by not having it. And if we find it useful later, or there is a user demand, we can change that.

mpariente:
>  How about moving the check for missing sample rate to after the model object has been constructed, and then checking using hasattr(model, "sample_rate")? That way you are free to set the sample_rate property however you like as long as it's present.

`sample_rate` has a default value, so actually this won't work because the `sample_rate` property will be there.

jonashaag:
How about we drop it? It's backwards incompatible, but loading models will still work, and it's easy to fix for people. Maybe it's time we don't default to 8 kHz anymore now that people are using Asteroid for other things except traditional 8 kHz speech separation

mpariente:
Well, I'd be ok with that!

mpariente:
This is enough for this PR. 
After merging this, I'll create a PR to remove the default on the sample rate.

Issue #411, opened 2021-01-09
Hooks: Document passing System.on_xxx to model.on_xxx


jonashaag:
## 📚 Documentation

`System` defines a bunch of placeholder hooks like `on_batch_start`, but doesn't pass them to the wrapped model or datasets. This was surprising to me because I expected `System` to be a transparent wrapped around `model`. I don't think we should change it to automatically forward those hooks to the model etc., but I think we should document it in `System` that if you want those hooks to reach your model instance, you'll have to do it yourself.

Plus, what do you think about removing the empty `on_xxx` methods from `System`? They aren't doing anything and the fact that only an arbitrary subset of possible hooks is defined there may be confusing to people (me).

mpariente:
Just to be sure I understand, you wrote the `hooks` in your model and expected them to be called from the `System`'s hooks? 
The forwarding is pretty simple, we could do it, IDK. 

> what do you think about removing the empty `on_xxx` methods from `System`?

Yes, it makes sense. We just link to the lightning docs about hooks.

jonashaag:
> Just to be sure I understand, you wrote the `hooks` in your model and expected them to be called from the `System`'s hooks?

Yes, or to be exact I forgot that there is a `System` "between" PL and my model.

For auto-forwarding, we could implement it using `__getattr__` or model introspection. `System` could forward to the model automatically, or raise a warning that it doesn't do automatic forwarding.

mpariente:
If we choose to forward it, what happens if both the model and a subclass of `System` implement hooks? Do we forward everything or just the known hooks from PL? 
What about data modules? 

It's easier just not to forward IMO. Hooks might involve something else than the model most of the time, so it makes sense to keep them in `System`. 

> For auto-forwarding, we could implement it using `__getattr__`.

That's what I had in mind for the implementation.

jonashaag:
Agreed, let's start with removing the empty `on_xxx` methods and adding a note to the docs.

PR #428, opened 2021-02-03
Add note about System Lightning hooks

jonashaag:
Refs #411

### DIFF
--- a/asteroid/engine/system.py
+++ b/asteroid/engine/system.py
@@ -10,6 +10,13 @@ class System(pl.LightningModule):
     """Base class for deep learning systems.
     Contains a model, an optimizer, a loss function, training and validation
     dataloaders and learning rate scheduler.
+    
+    Note that by default, any PyTorch-Lightning hooks are *not* passed to the model.
+    If you want to use Lightning hooks, add the hooks to a subclass::
+    
+        class MySystem(System):
+            def on_train_batch_start(batch, batch_idx, dataloader_idx):
+                return self.model.on_train_batch_start(batch, batch_idx, dataloader_idx)
 
     Args:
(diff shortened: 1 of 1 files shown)


asteroid/engine/system.py
```diff
@@ -10,6 +10,13 @@ class System(pl.LightningModule):
     """Base class for deep learning systems.
     Contains a model, an optimizer, a loss function, training and validation
     dataloaders and learning rate scheduler.
+    
+    Note that by default, any PyTorch-Lightning hooks are *not* passed to the model.
+    If you want to use Lightning hooks, add the hooks to a subclass::
+    
+        class MySystem(System):
+            def on_train_batch_start(self, batch, batch_idx, dataloader_idx):
+                return self.model.on_train_batch_start(batch, batch_idx, dataloader_idx)
```
mpariente, line 19: Should this be a code block?

mpariente:
Let's fix the linter and merge, thanks.

asteroid/engine/system.py
jonashaag: It is, `::` at the end of a line is special ReST syntax for "insert : and expect a code block"

jonashaag:
Did you install /lint now or not? :D

mpariente:
I couldn't make it work.. 
If you want to give it a try, feel free

mpariente:
The action did run, successfully, but it tells there is nothing to lint.. 

I'd have to dig into that a bit more. 
Maybe it's because we lost the info of the branch on the way, and it's linting master.
Really don't know. 

image

jonashaag:
It’s 2021 and we’ll soon all be driven around in our autonomous cars but  auto linting and formatting GitHub PRs is an unsolved problem

PR #429, opened 2021-02-03
Setup automatic linter

mpariente:
Try to set it up again. 
Inspired by how it works in this project: https://github.com/UnityTechnologies/open-project-1


mpariente:
Both linter and the push work, but the build is not triggered, again.

PR #430, opened 2021-02-05
[WIP, just for checking] A sample of re-factoring on bash and python scripts for further effective integration

underdogliu:
So first of all thanks for such great work on Asteroid! 

I am making this PR about some minor re-factoring of bash runners, as discussed with @mpariente earlier. I found the bash script is written in a bit-messy fashion with some commands included not working properly. So using WHAM's ConvTasnet runner as an example, I would like to propose such re-factoring on bash-level scripts. 

@mpariente I can confirm the model I uploaded earlier today (2021-02-05) uses this PR. Thus, this PR itself should work as before if setting everything properly.

I am still thinking about re-factoring python with minor regulations as well so **please do not merge it yet** even if it will look nicer. This is open just to receive criticism on time so I can make it better. Sharp comments such as "this is BS" are welcome.


egs/wham/ConvTasNet/eval.py
```diff
@@ -1,14 +1,16 @@
...
+import argparse
 import soundfile as sf
 import torch
 import yaml
 import json
-import argparse
 import numpy as np
 import pandas as pd
 from tqdm import tqdm
-from pprint import pprint
```
mpariente, line 11: Why changing imports?

egs/wham/ConvTasNet/eval.py
```diff
@@ -1,14 +1,16 @@
+#!/usr/bin/env python3
```
mpariente, line 1: Can you elaborate on the advantage of that?

egs/wham/ConvTasNet/local/prepare_data.sh
```diff
@@ -24,8 +25,9 @@ wait
 echo "Run python scripts to create the WHAM mixtures"
 # Requires : Numpy, Scipy, Pandas, and Pysoundfile
 cd $out_dir/wham_scripts
-$python_path create_wham_from_scratch.py \
+ln -sf ../../../corpus .
```
mpariente, line 28: Why this? 
This looks like a personal path, right?

egs/wham/ConvTasNet/run.sh
```diff
@@ -36,11 +20,14 @@ mode=min
 nondefault_src=  # If you want to train a network with 3 output streams for example.
 
 # Training
+# note: for batch_size=8 this needs more than 1 regular GPU (around 12-15G RAM)
+#   so id="0,1" should be specified. But of course when users try various
+#   network arch setting below this can change.
```
mpariente, line 25: We could reduce the default batch size if needed. 
But adding a comment that larger batch needs larger memory is not useful IMO.

egs/wham/ConvTasNet/run.sh
```diff
@@ -1,33 +1,17 @@
...
 stage=0  # Controls from which stage to start
-tag=""  # Controls the directory name associated to the experiment
+tag=1a_wham  # Controls the directory name associated to the experiment
+
+# GPU setting.
 # You can ask for several GPUs using id (passed to CUDA_VISIBLE_DEVICES)
-id=$CUDA_VISIBLE_DEVICES
+#   For single gpu, "0" or "1" would work
+#   For multi-gpu, something like "0,1"
+id=0
```
mpariente, line 14: Why? 
By default, this uses all available GPUs, why is it wrong?

egs/wham/ConvTasNet/run.sh
```diff
@@ -1,33 +1,17 @@
...
-python_path=python
-
-# Example usage
-# ./run.sh --stage 3 --tag my_tag --task sep_noisy --id 0,1
+set -eo pipefail
 
 # General
 stage=0  # Controls from which stage to start
-tag=""  # Controls the directory name associated to the experiment
+tag=1a_wham  # Controls the directory name associated to the experiment
```
mpariente, line 8: Why?

egs/wham/ConvTasNet/run.sh
```diff
@@ -36,11 +20,14 @@ mode=min
 nondefault_src=  # If you want to train a network with 3 output streams for example.
 
 # Training
+# note: for batch_size=8 this needs more than 1 regular GPU (around 12-15G RAM)
+#   so id="0,1" should be specified. But of course when users try various
+#   network arch setting below this can change.
 batch_size=8
 num_workers=8
-#optimizer=adam
+optimizer=adam
```
mpariente, line 28: This is unused in the script, so we can remove it.

egs/wham/ConvTasNet/run.sh
```diff
@@ -36,11 +20,14 @@ mode=min
...
 lr=0.001
-epochs=200
+epochs=20
```
mpariente, line 30: No need to change the default.

egs/wham/ConvTasNet/run.sh
```diff
@@ -50,9 +37,22 @@ mask_nonlinear=relu
...
+storage_dir=corpus
 
 . utils/parse_options.sh
 
+# If you start from the sphere files, specify the path to the directory and start from stage 0
+sphere_dir=$storage_dir/sph_files
+# If you already have wsj0 wav files, specify the path to the directory here and start from stage 1
+wsj0_wav_dir=$storage_dir/wsj0_wavs
+# If you already have the WHAM mixtures, specify the path to the directory here and start from stage 2
+wham_wav_dir=$storage_dir/wham_wavs
```
mpariente, line 51: Can you elaborate on what was the problem with the previous setting? 
IMO, having the storage options at the top makes more sense.

mpariente: And what if somebody starts from stage 1 with wsj0 wavs, but the folder name isn't called wsj0_wavs? Then the user has to scroll until here and change it by hand?

egs/wham/ConvTasNet/run.sh
```diff
@@ -61,17 +61,27 @@ train_dir=$dumpdir/tr
...
+
+if [ $stage -le 0 ]; then
+    echo "Stage 0: Converting sphere files to wav files"
+    local/convert_sphere2wav.sh --sphere_dir $sphere_dir --wav_dir $wsj0_wav_dir
 fi
 
-if [[ $stage -le  1 ]]; then
-	echo "Stage 1: Generating 8k and 16k WHAM dataset"
-  . local/prepare_data.sh --wav_dir $wsj0_wav_dir --out_dir $wham_wav_dir --python_path $python_path
+if [ $stage -le 1 ]; then
```
mpariente, line 79: Why not double brackets (`[[`) instead?

egs/wham/ConvTasNet/run.sh
```diff
@@ -61,17 +61,27 @@ train_dir=$dumpdir/tr
...
+    echo "Stage 1: Generating 8k and 16k WHAM dataset"
+    local/prepare_data.sh --wav-dir $wsj0_wav_dir --out-dir $wham_wav_dir
```
mpariente, line 81: Why remove the `.` ?

egs/wham/ConvTasNet/run.sh
```diff
@@ -80,49 +90,44 @@ if [[ $stage -le  2 ]]; then
...
 		--task $task \
 		--test_dir $test_dir \
 		--use_gpu $eval_use_gpu \
 		--exp_dir ${expdir} | tee logs/eval_${tag}.log
 	cp logs/eval_${tag}.log $expdir/eval.log
 fi
+
+wait
+echo done
+exit 0
```
mpariente, line 133: Why those things?

mpariente (commented):
Thanks for the suggestions ! 
I'm open to any improvement on the bash side, but they need to be explained and motivated. 
If I don't understand why, I won't accept it :wink:

egs/wham/ConvTasNet/eval.py
underdogliu: Oh I don't know what happened here. Will fix. 

But actually I will change them later if necessary. This is more like a style practice: for imports, usually we put built-in imports first, libraries then and functions from other scripts at last.

egs/wham/ConvTasNet/eval.py
underdogliu: Like we discussed, this is called `shebang` and it will allows the user to automatically use the python from his/her virtual environment.

egs/wham/ConvTasNet/local/prepare_data.sh
underdogliu: Sorry will fix this one.

egs/wham/ConvTasNet/run.sh
underdogliu: Tried on grid5k and seems not the case for me. But it works on the cluster from my school. Here's result on grid5k:

```
$ nvidia-smi
Fri Feb  5 19:13:32 2021
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 450.80.02    Driver Version: 450.80.02    CUDA Version: 11.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  GeForce GTX TIT...  On   | 00000000:82:00.0 Off |                  N/A |
| 26%   23C    P8    13W / 250W |      1MiB /  6083MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
$ echo $CUDA_VISIBLE_DEVICES

$ echo $CUDA_VISIBLE_DEVICES
```

egs/wham/ConvTasNet/run.sh
underdogliu: because we are not supposed to have directory jumping in `run.sh` so the dot is not necessary. And we do not have other `local` folders.

In some good bash practice from my point of view (like kaldi), we avoid directory jumping at very high level. I do not know if in sub scripts from folders such jumping via `cd` and `cd -` is good practice. Of course, if they are wrapped in functions, they are for sure.

egs/wham/ConvTasNet/run.sh
underdogliu: single brackets works and it looks more neat. But if you prefer double ones I can take them back.

egs/wham/ConvTasNet/eval.py
mpariente: The advantage of the other technique is that, without changing anything the user's env is used. But if the user wants, he can point to another env, so that it'll always be the same. 
This can also be done by activating the env by hand in run.sh. 

So, not a big advantage for the current way. I'd be willing to change that. 
@jonashaag, point of view? Mitigations?

egs/wham/ConvTasNet/run.sh
mpariente: The $CUDA_VISIBLE_DEVICES is set on most clusters I worked on. 
I don't think it's worth changing to a fixed id, which most of the time may be wrong.

egs/wham/ConvTasNet/run.sh
mpariente: Whatever actually, I'm not an opinionated bash user. 
But if we change double to single, we should do it everywhere!

egs/wham/ConvTasNet/run.sh
mpariente: > In some good bash practice from my point of view (like kaldi), we avoid directory jumping at very high level.

I agree with that.

egs/wham/ConvTasNet/eval.py
mpariente: IMO, either we use isort which does it automatically (we could), or we mainly what you say, which is mainly the case here. 
It's not worth too much energy. 
Maybe `isort` would make sense, so it cuts the debate short

egs/wham/ConvTasNet/run.sh
underdogliu: hmm....ok I will change it back.

mpariente:
- I don't know what motivated the choice to path `python_path` to `run.sh` anymore. Today, it seems to me like parasite code, so I'd be ok with removing it. On the other hand, the shebang doesn't add much right? Because we can still explicitly write `python3 script.py ...`, and the shebang is bypassed right? 
- Ok for the single brackets, because it indeed looks less cluttered. 
- About the GPUs, I think this is already well thought out, and we're not willing to change. 
- About the structure of `run.sh`, I think storage- and data-related stuff should be at the top, as it's already the case in most our recipes. 
- I'd be ok with adding isort to Asteroid, in the pre-commit hooks and in the CI, but we should create a GH action that does it automatically. Would you like to work on that @underdogliu ?

egs/wham/ConvTasNet/local/prepare_data.sh
```diff
@@ -25,7 +25,7 @@ wait
 echo "Run python scripts to create the WHAM mixtures"
 # Requires : Numpy, Scipy, Pandas, and Pysoundfile
 cd $out_dir/wham_scripts
-ln -sf ../../../corpus .
+ln -sf $(realpath $wav_dir/../) .
```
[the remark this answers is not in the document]
underdogliu: this is because when you getting into the folder, you lost the path at higher level so it needs some absoulte path soft linking to simply make the code work. But this is for my case only (on grid5k). If it does not looks right, I will delete them.

underdogliu:
@mpariente In respond to your comments:
- Yes the shebang is actually for executing python/bash script files without specifying things like 'python3' at the beginning. For example, if we have shebang at top of the file, we can simply call `train.py --blablabla`, just like a regular script with arguments. This is just my personal habit so yes I will remove it for asteroid.
- Single bracket thing does make things cleaner doesn't it?
- Yep  no problem. I changed it back already.
- Yes I can work on that if I have time later.

egs/wham/ConvTasNet/eval.py
jonashaag: I don’t understand what shebang has to do with virtualenv. If you use `python` as a relative path, no matter if it’s used as part of shebang or written explicitly in a shell scripts, it will always use the virtualenv Python. The only case where the system Python will be used is if you use absolute paths to the binary but I don’t think we are doing that anywhere (if so we definitely should change it to relative)

jonashaag: Had a quick look at how the eval script is invoked in the current version—the shebang isn’t used at all so this change is a no-op.

i actually think the old version with the global Python interpreter path is more flexible since you can tell it to use a different Python interpreter to run the scripts than is currently used by any virtualenv installations (if you have any). It also allows you to run the scripts without activating the virtualenv in the first place if you want that.

underdogliu: yes so I revised the shebang behavihor and found the same thing. I removed it already. Thanks a lot for such fruitful advice @jonashaag 

"The only case where the system Python will be used is if you use absolute paths to the binary but I don’t think we are doing that anywhere (if so we definitely should change it to relative)" -> yes that's what I will be a bit concerned for people who are hacking scripts to make for their own user cases. I met it before: It happened when I used python from brew (well, brew is a terrible tool to be played with in terms of linking and compatibility control though.....)

mpariente: >  actually think the old version with the global Python interpreter path is more flexible since you can tell it to use a different Python interpreter to run the scripts than is currently used by any virtualenv installations (if you have any). It also allows you to run the scripts without activating the virtualenv in the first place if you want that.

I agree with this, and this was the original intent. I wonder if it's really necessary though..

mpariente:
Thanks again for the PR @underdogliu 
I think I'll close it because we don't intend to spend energy and time on the bash side. The goal is to remove it altogether ! :wink:

PR #431, opened 2021-02-06
[src] Make sample rate positional

mpariente:
As discussed in #427, we'll remove default sample rate argument. 

We start with a deprecation warning when it is not passed, and will remove it in a later release.

### DIFF
--- a/asteroid/models/base_models.py
+++ b/asteroid/models/base_models.py
@@ -5,7 +5,7 @@ from typing import Optional
 from .. import separate
 from ..masknn import activations
 from ..utils.torch_utils import pad_x_to_y, script_if_tracing, jitable_shape
-from ..utils.hub_utils import cached_download
+from ..utils.hub_utils import cached_download, SR_HASHTABLE
 from ..utils.deprecation_utils import is_overridden, mark_deprecated, VisibleDeprecationWarning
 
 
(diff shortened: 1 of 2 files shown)


jonashaag:
Positional and having a default are orthogonal concepts. I think the wording should be: “will be required argument in the future.”

Also, is it safe to remove the backwards compat SR table lookups now? (I am not up to date on that front)

mpariente:
> Positional and having a default are orthogonal concepts. I think the wording should be: “will be required argument in the future.”

Thanks.
I was thinking about the wording and I was sure I messed it up :wink: 

> Also, is it safe to remove the backwards compat SR table lookups now? (I am not up to date on that front)

No, the only model that has 16k FS is DeMask, which is not on HF's hub yet (@popcornell ? ^^)

asteroid/models/base_models.py
```diff
@@ -37,8 +37,16 @@ class BaseModel(torch.nn.Module):
             If None, no checks will be performed.
     """
 
-    def __init__(self, sample_rate: float = 8000.0, n_channels: Optional[int] = 1):
+    def __init__(self, sample_rate: float = None, n_channels: Optional[int] = 1):
         super().__init__()
+        if sample_rate is None:
+            sample_rate = 8000.0
+            warnings.warn(
+                "The argument `sample_rate` of `BaseModel` will be made positional. "
```
mpariente, line 45:
```suggestion
                "The argument `sample_rate` of `BaseModel` will be required in the future. "
```

asteroid/models/base_models.py
```diff
@@ -37,8 +37,16 @@ class BaseModel(torch.nn.Module):
...
+    def __init__(self, sample_rate: float = None, n_channels: Optional[int] = 1):
         super().__init__()
+        if sample_rate is None:
+            sample_rate = 8000.0
+            warnings.warn(
+                "The argument `sample_rate` of `BaseModel` will be required in the future. "
+                "It is no longer a keyword argument. This will raise an error in future release. "
+                "Defaults to 8000.0",
+                VisibleDeprecationWarning,
+            )
```
mpariente, line 49: How does it look now @jonashaag ?

jonashaag: :+1:

PR #432, opened 2021-02-07
[src] Change n_channels to in_channels

mpariente:
This was pointed out by @jonashaag as a potential name. 

I was against using `out_channels` instead of `n_src` and ruled out `in_channels` in the same time. Got back to it, I think `in_channels` makes more sense and is more consistent with PyTorch's code, so it's better.


jonashaag:
👍

PR #433, opened 2021-02-07
[src] Fix unpassed SR in TasNet and SudoRMRF

mpariente:
Important fix ! 

Because the encoder was wrapped in `LSTMTasNet` and `SudoRmRF*`, `getattr` couldn't find it and it was not passed to `BaseModel`. 

Funny that we set barriers to load models without sampling rate, but managed to load models with the wrong sampling rate..  :sweat_smile:


jonashaag:
How did you find it? Test failure?

mpariente:
Unexpected warning in `BaseModel` with the new mechanism

PR #434, opened 2021-02-09
Benchmark

JorisCos:
### About this PR
This PR introduces a wrapper around different benchmark libraries.
It makes possible to perform inference time, memory usage and FLOPs measurement on CPU.


JorisCos:
Oups didn't see that branch was so messy I will reopen

PR #435, opened 2021-02-09
Benchmark

JorisCos:
About this PR

This PR introduces a wrapper around different benchmark libraries.
It makes possible to perform inference time, memory usage and FLOPs measurement on CPU.

### DIFF
--- a/asteroid/models/benchmarker.py
+++ b/asteroid/models/benchmarker.py
@@ -11,11 +11,29 @@ class Benchmarker:
     """Wrapper around benchmark libraries on CPU. For inference time, memory
     usage and flops measurements.
 
-        Args:
-            model (torch.nn.Module): Instance of model.
-            inputs_shape (list(int)): A sequence of integers defining the shape of the output tensor.
-            repeat (int,optional) : The number of measurement to make for inference time measurement.
+    Args:
+        model (torch.nn.Module): Instance of model.
+        inputs_shape (list(int)): A sequence of integers defining the shape of the output tensor.
+        repeat (int,optional) : The number of measurement to make for inference time measurement.
 
+    Examples:
+        Examples should be written in doctest format, and should illustrate how
+        to use the function.
+        >>> from asteroid.models.benchmarker import Benchmarker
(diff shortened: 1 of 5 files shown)


asteroid/models/benchmarker.py
```diff
@@ -0,0 +1,67 @@
...
+import numpy as np
+
+from pypapi.exceptions import PapiNoEventError
+
+
+class Benchmarker:
+    """Wrapper around benchmark libraries on CPU. For inference time, memory
+    usage and flops measurements.
+
+        Args:
```
mpariente, line 14: Shouldn't be indented :wink:

mpariente (commented):
This looks cool !
Can you include an example code in the class docstring, with outputs as well? That would be very useful.  

@popcornell @jonashaag would you have things to add to the benchmark?

JorisCos:
I added example but memory report is unreadable at least on github

mpariente (commented):
Let's get this merged :-)
Two minor comments

JorisCos:
Done !

mpariente:
Great, thanks ! 

Could you add a test with a dummy model like `nn.Identity` please?

JorisCos:
I added a test, it just runs the functions for coverage.

mpariente:
We probably need to add the dependencies in requirements/dev.txt.

tests/models/benchmark_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Dummy_model(torch.nn.Module):
+    def __init__(self):
+        super(Dummy_model, self).__init__()
+        self.layer = torch.nn.Identity()
+
+    def forward(self, x):
+        return self.layer(x)
+
```
mpariente, line 12:
```suggestion
```

mpariente:
Given that there is a new profiler in torch and lightning, I wonder if it's useful to add our own here. 
Any feedback on that appreciated.

mpariente:
Let's close this until we decide to integrate something from PyTorch. 
Thanks again for the PR.

PR #437, opened 2021-02-12
fixed bug about reshaping

popcornell:
permutation bug

### DIFF
--- a/egs/TAC/local/conf.yml
+++ b/egs/TAC/local/conf.yml
@@ -14,7 +14,6 @@ net:
   n_src: 2
   window_ms: 4
   context_ms: 16
-  sample_rate: 16000
 optim:
   lr: 0.001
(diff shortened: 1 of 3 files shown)


egs/TAC/local/conf.yml
```diff
@@ -14,7 +14,6 @@ net:
   n_src: 2
   window_ms: 4
   context_ms: 16
-  sample_rate: 16000
```
mpariente, line 17: If it's not passed to the model, the loading will be wrong, right?

popcornell: If it is specified here then there is a conflict with the command line one and the recipe fails.

popcornell: Okay i will change train.py to take sample_rate from the json

PR #441, opened 2021-02-13
Reshaping issue in AVSpeech

vitrioil:
As suggested in [438](https://github.com/asteroid-team/asteroid/issues/438).

Also there was an issue with ls path that was fixed.


mpariente:
Thanks !

PR #451, opened 2021-02-24
ConferencingSpeech2021

JorisCos:
# About this PR
This PR is meant as a baseline for ConferencingSpeech2021 It tackles data download and data preparation for the challenge and uses FasNet to process the data.

# Things to discuss
@popcornell 
For now, I have left the data generation almost untouched from the official scripts but we don't store any generated file.
The config files generation takes a while especially for the train_set and fixes the chunk_length maybe we don't want that ?
I tested the recipe and listen to some sounds it seems ok. I can't test the training process. 
I was wondering what is the shape expected by FastNet ? [n_channel,time] or [time,n_channel]?

### DIFF
--- a/asteroid/models/conv_tasnet.py
+++ b/asteroid/models/conv_tasnet.py
@@ -1,6 +1,7 @@
 from asteroid_filterbanks import make_enc_dec
 from ..masknn import TDConvNet
 from .base_models import BaseEncoderMaskerDecoder
+import warnings
 
 
 class ConvTasNet(BaseEncoderMaskerDecoder):
@@ -81,6 +82,13 @@ class ConvTasNet(BaseEncoderMaskerDecoder):
                 " and number of input channels should "
                 "be the same. Received "
                 f"{n_feats} and {in_chan}"
+            )
+        if causal and norm_type not in ["cgLN", "cLN"]:
+            norm_type = "cgLN"
+            warnings.warn(
+                "In causal configuration cumulative layer normalization (cgLN)"
+                "or channel-wise Layer Normalization (chanLN)  "
+                f"must be used. Changing f{norm_type} to cgLN"
             )
         # Update in_chan
         masker = TDConvNet(


popcornell:
> I can't test the training process.

cluster is down I know

mpariente:
@popcornell would you like to review?

popcornell:
Sure I am testing it right now. Started yesterday from data downloading so it will take a time, but at least we will be sure it will run e2e

asteroid/data/conferencing_speech_datastet.py
```diff
@@ -0,0 +1,319 @@
...
+        segment (int, optional) : The desired sources and mixtures length in s.
+
+    """
+
+    dataset_name = "CS2021"
+
+    def __init__(
+        self,
+        config_dir,
+        rir_type="linear",
```
popcornell, line 30: because FasNetTAC is array topology independent here we can use a list of RIRs also to train on everything. 
I will add this

asteroid/data/conferencing_speech_datastet.py
```diff
@@ -0,0 +1,319 @@
...
+        line = self.config[idx]
+        if self.on_the_fly:
+            mix, reverb_ref, noreverb_ref = mix_func(line, self.chunk, self.sample_rate)
+        else:
+            mix_path, ref_path, ref_noreverb_path = line.strip("\n").split(" ")
+            mix, _ = sf.read(mix_path, dtype="float32")
+            reverb_ref, _ = sf.read(ref_path, dtype="float32")
+            noreverb_ref, _ = sf.read(ref_noreverb_path, dtype="float32")
+        if self.reverb:
+            source = reverb_ref
```
popcornell, line 62: we should do dereverberation also no ?

asteroid/data/conferencing_speech_datastet.py
```diff
@@ -0,0 +1,319 @@
...
+
+    """
+
+    dataset_name = "CS2021"
+
+    def __init__(
+        self,
+        config_dir,
+        rir_type="linear",
+        reverb=True,
```
popcornell, line 31: we need an argument for the target: dry, reverb and early

egs/conferencing_speech_2021/local/download.sh
```diff
@@ -0,0 +1,119 @@
...
+fi
+
+if ! test -e $vctk; then
+    echo "Downloading VCTK into $vctk"
+          wget -c --tries=0 --read-timeout=20 https://datashare.ed.ac.uk/download/DS_10283_3443.zip -P $vctk --no-check-certificate
+          unzip -q $vctk/DS_10283_3443.zip -d $vctk
+          unzip -q $vctk/VCTK-Corpus-0.92.zip -d $vctk
+          rm -rf $vctk/DS_10283_3443.zip
+          rm -rf $vctk/VCTK-Corpus-0.92.zip
+fi
```
Emrys365, line 74: I think you may need to download VCTK-Corpus v0.80 instead. See the discussion here: https://github.com/ConferencingSpeech/ConferencingSpeech2021/issues/14.

popcornell: Thank you for noticing this !

JorisCos:
What should we do with this PR ? It would nice to have an example of multichannel source separation on Asteroid but I don't think using this challenge is the most relevant way to do it, plus we don't have any results or working recipe on it.

mpariente:
IDK @popcornell WDYT?

PR #454, opened 2021-02-24
Wavesplit 2021

popcornell:
Should work now with oracle embedding. I made a separate pull request because it is faster 
See previous pull request also: https://github.com/asteroid-team/asteroid/pull/70 from last year.


egs/wham/wavesplit/train.py
```diff
@@ -0,0 +1,218 @@
...
+        self.loss_func["spk_loss"] = self.loss_func["spk_loss"].to(self.device)
+
+    def on_validation_epoch_start(self) -> None:
+        self.loss_func["spk_loss"] = self.loss_func["spk_loss"].to(self.device)
+
+    def training_step(self, batch, batch_nb):
+        mixtures, oracle_s, oracle_ids = batch
+        b, n_spk, frames = oracle_s.size()
+
+        # spk_vectors = self.model.get_speaker_vectors(mixtures)
```
popcornell, line 38: Here and in validation steps I use oracle embeddings for now and no speaker stack

egs/wham/wavesplit/train.py
```diff
@@ -0,0 +1,218 @@
...
+        num_workers=conf["training"]["num_workers"],
+        drop_last=True,
+    )
+    # Update number of source values (It depends on the task)
+    conf["masknet"].update({"n_src": train_set.n_src})
+
+    model = Wavesplit(
+        conf["masknet"]["n_src"],
+        {"embed_dim": 512},
+        {"embed_dim": 512, "spk_vec_dim": 512, "n_repeats": 4, "return_all_layers": False},
```
popcornell, line 128: If anyone wants to experiment with this here you can change the hyperparams

egs/wham/wavesplit/local/conf.yml
```diff
@@ -0,0 +1,25 @@
...
+  lr: 0.001
+# Data config
+data:
+  train_dir: data/wav8k/min/tr/
+  valid_dir: data/wav8k/min/cv/
+  task: sep_clean
+  nondefault_nsrc:
+  sample_rate: 8000
+  mode: min
+  segment: 1.0
```
popcornell, line 25: 1.0 seconds or 0.75 as in the paper is enough

mpariente:
I'll review after @JorisCos

popcornell:
It would be cool if someone can try to run the training with the full system and not oracle embeddings. You can wait for review when the full system has been trained and performance is decent

egs/wham/wavesplit/dataloading.py
```diff
@@ -0,0 +1,153 @@
...
+        return torch.from_numpy(x), sources, torch.Tensor(c_ex["spk_id"]).long()
+
+
+if __name__ == "__main__":
+    a = WHAMID(
+        "/media/sam/bx500/wavesplit/asteroid/egs/wham/wavesplit/data/wav8k/min/tt", "sep_clean"
+    )
+
+    for i in a:
+        print(i[-1])
```
JorisCos, line 153: To be removed

egs/wham/wavesplit/local/preprocess_wham.py
```diff
@@ -0,0 +1,93 @@
...
+    parser = argparse.ArgumentParser("WHAM data preprocessing")
+    parser.add_argument(
+        "--in_dir", type=str, default=None, help="Directory path of wham including tr, cv and tt"
+    )
+    parser.add_argument(
+        "--out_dir", type=str, default=None, help="Directory path to put output files"
+    )
+    args = parser.parse_args()
+    print(args)
+    preprocess(args)
```
JorisCos, line 93: I think we should create a `def main(args)` at the beginning if the file,  put the args for the parser also at the beginning and call preprocess inside `main(args)` it's more user friendly we can see directly the arguments and the function that is called without scrolling

egs/wham/wavesplit/losses.py
```diff
@@ -0,0 +1,240 @@
...
+        distances = torch.exp(-distances).sum(1)
+
+        return (distance_utt + torch.log(distances)).sum(1)
+
+        # exp normalize trick
+        # with torch.no_grad():
+        #   b = torch.max(distances, dim=1, keepdim=True)[0]
+        # out = -distance_utt + b.squeeze(1) - torch.log(torch.exp(-distances + b).sum(1))
+        # return out.sum(1)
+
```
JorisCos, line 107: Remove ?

egs/wham/wavesplit/losses.py
```diff
@@ -0,0 +1,240 @@
...
+        .unsqueeze(-1)
+        .repeat(1, 1, 1, 200)
+    )
+
+    # testing exp normalize average
+    # distances = torch.ones((1, 101, 4000))
+    # with torch.no_grad():
+    #   b = torch.max(distances, dim=1, keepdim=True)[0]
+    # out = b.squeeze(1) - torch.log(torch.exp(-distances + b).sum(1))
+    # out2 = - torch.log(torch.exp(-distances).sum(1))
```
JorisCos, line 229: Remove ?

egs/wham/wavesplit/model.py
```diff
@@ -0,0 +1,281 @@
+from torch import nn
+import torch
+from asteroid.masknn import norms
+from kmeans_pytorch import kmeans, kmeans_predict
+
+
+class Conv1DBlock(nn.Module):
```
JorisCos, line 7: Don't you think we should make wavesplit part of asteroid not just the WHAM recipes ?

JorisCos (commented):
It was a very nice surprise to see a new Wavesplit PR for Asteroid thanks @popcornell.
I made my review with general comments and questions.  Aren't we missing the eval script and the tests ?

wangshuo182:
It seems to work well with oracle embedding (18.5dB score was improved in the WSJ-2mix validation after 50 epochs). But when two stacks are jointly trained, the separation stack yields almost the same signals as the mixture, and the SISDR metric tends to be zero. Could you please share the results if anyone has tried the complete pipeline? @popcornell @JorisCos

popcornell:
That's very interesting to know ! 
Unfortunately all I have is here on GitHub. Maybe Joris has more up to date code. 

Do you think the degradation is due to over fitting of the training speakers IDs ? 
It may be. In the paper they use some things like speaker dropout to mitigate that. 
WSJ2Mix is small regarding speakers diversity after all and for reasonable speakers ID extraction usually you need tons of diversity e.g. voxceleb

lminer:
I tried to run this implementation on a dataset with around 60000 speakers
and the speaker stack loss never changed. Could there be a bug somewhere?

TCord:
> It seems to work well with oracle embedding (18.5dB score was improved in the WSJ-2mix validation after 50 epochs). But when two stacks are jointly trained, the separation stack yields almost the same signals as the mixture, and the SISDR metric tends to be zero. Could you please share the results if anyone has tried the complete pipeline? @popcornell @JorisCos

Hi, I also tried to run some experiments with Wavesplit (albeit in our own framework) in the past. I think a stagnating training of the speaker stack might result from two things:
1. Missing shuffling of speaker IDs during training:
The target speaker IDs need to be shuffled every time so that the model really needs the embeddings of the speaker stack to solve the permutation problem at the output. I haven't checked all of the code, but at least at first glance, I have not seen this shuffling for the target IDs.
2. The latent dimension of the speaker stack is too high:
If you don't use dynamic mixing (DM), the WSJ2mix dataset has 101 speakers, while the latent dimension is set to 256 in the corresponding paper. So, without DM (resulting in 285 speakers of WSJ0+WSJ1) there is no information bottleneck that would enable the model to generalize to unseen speakers. Even then, the generalization tricks that @popcornell mentioned (centroid dropout / mixup) still should be necessary for the model to really generalize to unseen speakers.
Reducing the latent dimension to e.g. 64 and see if it's learning something then might work. 

But even then, all of this should not prevent the model from at least overfitting to the training set.

popcornell:
Do they use shuffling in the paper ? It sounds a very smart thing to do but they don't seem to use it. 
Also I remember in the first version of the paper there were not many augmentations on the speaker stack (no speaker dropout for example, maybe only gaussian noise ?). 
Maybe Neil Zeghidour still has some other hidden tricks to make the model generalize better. @TCord were you able to succesfully replicate it to some decent degree ?

popcornell:
@lminer did you use voxceleb ?

TCord:
If I remember it correctly, they also used the label shuffling in the paper. In my experiments, I did not use the architecture as proposed in the paper, but a Conv-TasNet as separation stack (i.e. I added an additional encoder/decoder layer) and reduced the total amount of layers. Here, I was able to train the model, but it did not improve upon the performance of a Conv-TasNet. 
My conjecture was that a sample-wise resolution as in the paper is necessary to obtain good results and provides the most significant improvements. I think, in the DPRNN papers it was also shown that choosing a very small window size and frame advance in the encoder further improved the separation for anechoic data. 
By employing an additional speaker stack jointly with a Conv-TasNet, the permutation problem could be solved through the speaker stack, but it did not improve the separation performance over a sole Conv-TasNet. As training the full-size Wavesplit model without any additional encoder/decoder layer takes a massive amount of GPU memory, I dropped these experiments afterwards.

popcornell:
I have observed the same actually. Also according to https://arxiv.org/abs/2202.00733 the use of speaker ID info does in fact nor really help.

PR #459, opened 2021-03-08
fix a bug in 00_GettingStarted.ipynb

quancs:
fix bug: mixture's dimension does not fit the requirement of model.separate (accept 2- or 3-dimensional array)


jonashaag:
Thanks for the patch! This probably comes from the multichannel refactor we've done recently.

I think both `.T` and `[None]` are hard to understand for people that are pretty new to NumPy, and in the tutorial we should try to keep the mental burden as low as possible. (Since people are likely learning a dozen other new things at the same time, which is hard already!)

So let's find a more obvious way to do the shape transform. I'll be making a suggestion but if you have different ideas, please suggest!

notebooks/00_GettingStarted.ipynb
```diff
@@ -1248,8 +1248,9 @@
         "model = BaseModel.from_pretrained(\"mpariente/DPRNNTasNet-ks2_WHAM_sepclean\")\n",
         "\n",
         "# You can pass a NumPy array:\n",
-        "mixture, _ = sf.read(\"female-female-mixture.wav\", dtype=\"float32\")\n",
-        "model.separate(mixture)\n",
+        "mixture, _ = sf.read(\"female-female-mixture.wav\", dtype=\"float32\", always_2d=True)\n",
+        "mixture = mixture.T[None]\n",
+        "out_wavs = model.separate(mixture)\n",
```
jonashaag, line 1253:
```suggestion
        "mixture, _ = sf.read(\"female-female-mixture.wav\", dtype=\"float32\", always_2d=True)\n",
        "# Soundfile returns the mixture as shape (time, channels), and Asteroid expects (batch, channels, time).\n",
        "mixture = mixture.reshape(1, mixture.shape[1], mixture.shape[0])\n",
        "out_wavs = model.separate(mixture)\n",
```

jonashaag: Note I just coded this up in the GitHub editor so maybe there is a syntax error.

Maybe we could also drop the batch dimension here, not sure if that's easier to understand for the reader.

jonashaag: On second thought, the code above only works if channels is 1, so if we decide to ignore batch here, maybe we should just use `mixture.transpose()` instead?

quancs: > On second thought, the code above only works if channels is 1, so if we decide to ignore batch here, maybe we should just use `mixture.transpose()` instead?

If 'separate' is not used in the training steps, I think using `mixture.transpose()` is easier. Otherwise, I think showing the users a consistent shape of the mixture is better, because if they want to train, they have to know the batch dimension. (I'm new to Asteroid, I don't know if 'separate' is called in the training steps, hahaha...)

mpariente: I think the `mixture.reshape` is very explicit, that's good, let's go with that.

mpariente:
Thanks for the PR, let's apply the suggestion of Jonas and we can merge.

mpariente:
@quancs do you want to go `forward` with this?

quancs:
> @quancs do you want to go `forward` with this?

Sorry, I'm not familiar with github. Should I commit the changes to my fork, and open a new pull request?

quancs:
I opened a new pull request. @mpariente

mpariente:
Thanks!

You can keep pushing to the same branch you opened the pull request with,
it will automatically add the commits to the PR.
But no problem for opening a new PR!

PR #464, opened 2021-03-18
Enable tanh masking in DCUNet-based models

adobles96:
For context, one of the central points of Hyeong-Seok Choi et al. (https://arxiv.org/abs/1903.03107) revolves around masking methods. Specifically they propose masking by first applying a `tanh` transform to the magnitude of the mask, thereby limiting the mask to the unit disc in the complex plane.  They claim it improves performance over the default "unbounded" masking method. This method is also used in Yanxin Hu et al. (https://arxiv.org/abs/2008.00264), in their DCCRN-E model, which they say had the best subjective performance.

This PR seeks to implement `tanh` masking for all DCUNet-based models.

### DIFF
--- a/asteroid/models/dcunet.py
+++ b/asteroid/models/dcunet.py
@@ -1,3 +1,5 @@
+import torch
+
 from asteroid_filterbanks import make_enc_dec
 from asteroid_filterbanks.transforms import from_torch_complex, to_torch_complex
 from ..masknn.convolutional import DCUMaskNet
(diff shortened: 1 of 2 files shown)


mpariente:
Thanks a lot for the PR! 
Have you had a look at `asteroid/complex_nn.py`?

adobles96:
Had not had a look, I see it's already implemented 😅. Closing the PR. Thanks!

mpariente:
Thanks for the PR anyway
We should document these things better. 
What part of the code did you read? How did you learn about specific functionalities? 

I'm trying to understand how to make these things more visible.

adobles96:
What I did was go to `models/dccrnet.py`. There, I looked at `DCCRNet` and saw that `apply_masks` was straight-forward complex multiplication. I didn't see any args/kwargs to change this other than potentially `masknet_kwargs`. So then I checked out `DCCRMaskNet`. Similarly, I saw no args/kwargs there related to masking. I stopped following the inheritance at this point. Of course, had I looked at `BaseDCUMaksnet` I would've found what I was looking for. 

In terms of making this more visible I would perhaps add `mask_bound` as an explicit kwarg for either `DCCRNet` or `DCCRMaskNet` so as to avoid having to traverse the inheritance hierarchy. Maybe it's too specific to this case but I do believe `mask_bound` to be particularly relevant to the DCCRNet model. On another note, I think the docs should be linked to in the main readme, they're quite good and some people might miss them in `docs/README.md`

Thanks again!

mpariente:
Thanks for the feedback! 

For the docs, if you click on the docs badge (0.4.4 and latest), you go there. 
image

But actually, if it's not obvious, we should add it in the main Readme somewhere. 

@jonashaag WDYT about the `mask_bound`?

PR #468, opened 2021-03-26
Add beamforming modules

mpariente:
This is inspired by the code in [`nn-gev`](https://github.com/fgnt/nn-gev/blob/f8287020b18a00dd34eb58ba439b8d7dfad74fac/fgnt/beamforming.py) from Paderborn and François Grondin's code in [speechbrain](https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/processing/decomposition.py).
Thanks for open-sourcing it. 
Also helped by @popcornell for MWF

### Left to code: 
- GCCPhat with quadratic interpolation (cf François Grondin's code)
- steering vector 
- Blind analytic normalization 
- GEV vector phase correction

### In parallel 
- More docs
- Examples
- Tests

Might be interesting for @boeddeker @FrancoisGrondin @faroit

### DIFF
--- a/asteroid/dsp/beamforming.py
+++ b/asteroid/dsp/beamforming.py
@@ -1,0 +1,182 @@
+import torch
+from torch import nn
+
+
+class SCM(nn.Module):
+    def forward(self, x: torch.Tensor, mask: torch.Tensor = None, normalize: bool = True):
+        """Compute the spatial covariance matrix from a STFT signal x.
+
+        Args:
+            x (torch.ComplexTensor): shape  [batch, mics, freqs, frames]
+            mask (torch.Tensor): [batch, 1, freqs, frames] or [batch, 1, freqs, frames]. Optional
+            normalize (bool): Whether to normalize with the mask mean per bin.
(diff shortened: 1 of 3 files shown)


popcornell:
GCCPhat with quadratic interpolation (cf François Grondin's code). 

We might want to do a Localization module for that.

mpariente:
Makes sense for the localisation, you're right. 

By the way, eternal question, should we make a functional version, and module wrappers? 
It might be easier to use for functional addicts

popcornell:
Can you be more specific ? 
Module wrappers you mean wrapping an STFT encoder module in the case of GCC-PHAT ?

mpariente:
Sorry. 

Currently, the beamforming code it's only accessible in nn.Modules. 
In PyTorch, all ops are first implemented as functions, and then it also provides some class wrappers. 

Should we outsource everything in functions, and only make the function calls in the class, so bothb usage are possible?

mpariente:
@popcornell could you review please?

PR #482, opened 2021-04-19
[src] Compute linalg beamforming ops in double

mpariente:
As @popcornell  pointed out, numpy automatically casts to double for solving, and torch doesn't. 

This PR makes sure that all the solve, cholesky and inverse ops are done in double (float or complex).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,4 +10,6 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
+        args:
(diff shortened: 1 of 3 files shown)


mpariente:
Another solution would have been to write a decorator that automatically casts before and after. It would have been simpler to read, once the decorator would be implemented. 
My first attempt looked like a lot of overhead. 
We might revert to this if we have more and more ops like that, but for now 3 ops by hand is fine.

PR #484, opened 2021-04-19
[src] Beamforming: optimal channel selection

mpariente:
Based on a posteriori SNR from [here](https://www.merl.com/publications/docs/TR2016-072.pdf) 

image

Still need to enable custom channel selection with integer and torch.Tensor

### DIFF
--- a/asteroid/dsp/beamforming.py
+++ b/asteroid/dsp/beamforming.py
@@ -135,8 +135,8 @@ class SoudenMVDRBeamformer(Beamformer):
         r"""Compute and apply MVDR beamformer from the speech and noise SCM matrices.
         This class uses Souden's formulation [1]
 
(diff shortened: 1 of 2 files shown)


mpariente:
I'm unsure about the generic `get_reference_mic_idx` function. 
I think it's useful but maybe it's too obscure?

popcornell:
what if we enforce in every beamforming function that the ref_channel is a vector of shape B, M  ? 
And `get_ref_mics` converts integer and vector of B integers to that format ? 
In that way one could use a DNN to output the ref_channel like in channel selection. I think it is the most flexible thing

mpariente:
And could this matrix have soft assignments (as opposed to the one-hot-like we have in this PR)?

mpariente:
A side note: 
I might be wrong but not all beamformers can choose the channel right? 
GEV for example doesn't, am I mistaken?

PR #503, opened 2021-05-17
Fix default norm_type for causal ConvTasNet

JorisCos:
This is to avoid using any non causal  `norm_type` when `causal=True`.

### DIFF
--- a/asteroid/models/conv_tasnet.py
+++ b/asteroid/models/conv_tasnet.py
@@ -83,7 +83,7 @@ class ConvTasNet(BaseEncoderMaskerDecoder):
                 "be the same. Received "
                 f"{n_feats} and {in_chan}"
             )
-        if causal and norm_type != "cgLN":
+        if causal and norm_type not in ['cgLN', 'cLN']:
             norm_type = "cgLN"
             warnings.warn(
                 "In causal configuration cumulative layer normalization (cgLN)"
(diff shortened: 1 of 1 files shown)


asteroid/models/conv_tasnet.py
```diff
@@ -82,6 +82,12 @@ def __init__(
                 "be the same. Received "
                 f"{n_feats} and {in_chan}"
             )
+        if causal and norm_type != "cgLN":
+            norm_type = "cgLN"
+            print(
+                "In causal configuration cumulative layer normalization"
+                f"must be used. Changing f{norm_type} to cgLN"
+            )
```
mpariente, line 90: Can you use `warnings.warn()` instead? 

How about channel layer norm, is it over time as well?

mpariente:
Thanks :blush:

asteroid/models/conv_tasnet.py
JorisCos: Channel Layer norm should be fine too, the mean and var is only computed on channel dim. In the paper they didn't use it though.

JorisCos: I think we can just warn about these two norms and use cumLN as default in causal mode, right ?

asteroid/models/conv_tasnet.py
```diff
@@ -84,8 +85,9 @@ def __init__(
             )
         if causal and norm_type != "cgLN":
```
mpariente, line 86: In this case: `if causal norm_type not in [...]` would be better, so that we can also allow the channel normalization.

mpariente:
Can you please lint it? Or is the failure unrelated? 
I'll merge after, thanks again!

JorisCos:
This should be ready now

PR #506, opened 2021-05-25
Replication of directional ConvTasNet

sufeidechabei:
The original paper:https://arxiv.org/pdf/2001.00391.pdf, main content are  in train.py and preprocess.py

### DIFF
--- a/egs/direction_tasnet/conv_stft.py
+++ b/egs/direction_tasnet/conv_stft.py
@@ -1,0 +1,186 @@
+import torch
+import numpy as np
+import torch.nn.functional as F
+from scipy.signal import get_window
+from librosa.util import pad_center, tiny
+from utils import window_sumsquare
+
+
+
+class STFT(torch.nn.Module):
+    def __init__(self, filter_length=1024, hop_length=512, win_length=None,
+                 window='hann'):
+        """
+        This module implements an STFT using 1D convolution and 1D transpose convolutions.
+        This is a bit tricky so there are some cases that probably won't work as working
+        out the same sizes before and after in all overlap add setups is tough. Right now,
+        this code should work with hop lengths that are half the filter length (50% overlap
+        between frames).
+
+        Keyword Arguments:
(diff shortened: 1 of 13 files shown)


egs/direction_tasnet/conv_stft.py
```diff
@@ -0,0 +1,186 @@
+import torch
```
mpariente, line 1: Use `asteroid_filterbanks` instead.

egs/direction_tasnet/utility/models.py
```diff
@@ -0,0 +1,249 @@
+import numpy as np
```
mpariente, line 1: Those should be in asteroid already. 
Use it instead.

egs/direction_tasnet/conv_tasnet.py
```diff
@@ -0,0 +1,95 @@
+import torch
```
mpariente, line 1: This should also be in asteroid, use it instead.

egs/direction_tasnet/preprocess.py
```diff
@@ -0,0 +1,126 @@
...
+V = 343
+z = np.zeros((n_mic, n_grid, m_total))
+a = np.arange(m_total).reshape(1, 1, m_total)
+m_data = torch.from_numpy(z + a)
+n_sp = 3
+
+
+
+
+def Prep(data):
```
mpariente, line 30: This is where the features are computed right? 
If you ask specific review points, I can read some code, with comments. But otherwise, the intent doesn't look clear enough that it will be fast to review.

mpariente (commented):
Thanks for opening the PR. 

This seems like there is a lot of copy-pasted code, without attribution, this should be fixed. 

I can review code that is intended to integrate Asteroid, otherwise, it's a loss of time for everybody. 
I'd suggest making a PR only with the spatial features, with clean and commented code, and there will be a much higher chance that you get useful feedback.

PR #511, opened 2021-06-03
Default norm in causal convTasnet 

JorisCos:
Using `cLN` gives significant better results than `cgLN` it should be used as default.


mpariente:
Thanks :pray:

PR #520, opened 2021-06-22
Add GEVD beamformer

ldelebec:
GEVD beamformer is a kind of SDW MWF but computed in eigen space. It gives a good SIR but with less distortion than GEV beamformer.
The code is based on the one written by @nfurnon.

### DIFF
--- a/asteroid/dsp/beamforming.py
+++ b/asteroid/dsp/beamforming.py
@@ -375,7 +375,7 @@ def generalized_eigenvalue_decomposition(a, b):
     if input_dtype not in [torch.float64, torch.complex128]:
         solve_dtype = _precision_mapping()[input_dtype]
     e_val, e_vec = _generalized_eigenvalue_decomposition(a.to(solve_dtype), b.to(solve_dtype))
-    return e_val.to(input_dtype), e_vec.to(input_dtype)
+    return e_val.to(input_dtype).real, e_vec.to(input_dtype)
 
 
 def _generalized_eigenvalue_decomposition(a, b):
--- a/asteroid/dsp/beamforming.py
+++ b/asteroid/dsp/beamforming.py
@@ -248,6 +248,91 @@ class GEVBeamformer(Beamformer):
         return bf_vect
 
 
+class GEVDBeamformer(Beamformer):
+    """Generalized eigenvalue decomposition speech distortion weighted multichannel Wiener filter.
+
+        Compare to SDW-MWF, spatial covariance matrix are computed from low rank approximation
+        based on eigen values decomposition,
+        see equation 62 in `[1] <https://hal.inria.fr/hal-01390918/file/14-1.pdf>`_.
+
+    Attributes:
+        mu (float): Speech distortion constant.
+        rank (int): Rank for the approximation of target covariance matrix,
+            no approximation is made if rank is None.
+
+    References:
+        [1] R. Serizel, M. Moonen, B. Van Dijk and J. Wouters,
+        "Low-rank Approximation Based Multichannel Wiener Filter Algorithms for
+        Noise Reduction with Application in Cochlear Implants,"
+        in IEEE/ACM Transactions on Audio, Speech, and Language Processing, April 2014.
+    """
+
+    def __init__(self, mu: float, rank: int):
+        self.mu = mu
+        self.rank = rank
+
+    def compute_beamforming_vector(self, target_scm: torch.Tensor, noise_scm: torch.Tensor):
+        """Compute beamforming vectors for GEVD beamFormer.
+
+        Args:
+            target_scm (torch.ComplexTensor): shape (batch, mics, mics, freqs)
+            noise_scm (torch.ComplexTensor): shape (batch, mics, mics, freqs)
+
+        Returns:
+            torch.ComplexTensor: shape (batch, mics, freqs)
+
+        """
+        #  GEV decomposition of noise_scm^(-1) * target_scm
+        e_values, e_vectors = _generalized_eigenvalue_decomposition(
+            target_scm.permute(0, 3, 1, 2),  # bmmf -> bfmm
+            noise_scm.permute(0, 3, 1, 2),  # bmmf -> bfmm
+        )
+
+        #  Prevent negative and infinite eigenvalues
+        eps = torch.finfo(e_values.dtype).eps
+
+        e_values = torch.maximum(e_values, eps * torch.ones(e_values.shape))
+        e_values = torch.minimum(e_values, 1e6 * torch.ones(e_values.shape))
+
+        #  Sort eigen values and vectors in descending-order
+        e_values = torch.diag_embed(torch.flip(e_values, [-1]))
+        e_vectors = torch.flip(e_vectors, [-1])
+
+        #  Force zero values for all GEV but the highest
+        if self.rank:
+            e_values[..., self.rank :, :] = 0.0
+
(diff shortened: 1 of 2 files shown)


mpariente:
Thank you for the PR !
I'll probably have a look next week, sorry for the delay.

asteroid/dsp/beamforming.py
```diff
@@ -248,6 +248,91 @@ def compute_beamforming_vector(target_scm: torch.Tensor, noise_scm: torch.Tensor
...
+    """Generalized eigenvalue decomposition speech distortion weighted multichannel Wiener filter.
+
+        Compare to SDW-MWF, spatial covariance matrix are computed from low rank approximation
+        based on eigen values decomposition,
+        see equation 62 in `[1] <https://hal.inria.fr/hal-01390918/file/14-1.pdf>`_.
+
+    Attributes:
+        mu (float): Speech distortion constant.
+        rank (int): Rank for the approximation of target covariance matrix,
+            no approximation is made if rank is None.
```
mpariente, line 261: Should None be the default, then?

asteroid/dsp/beamforming.py
```diff
@@ -248,6 +248,91 @@ def compute_beamforming_vector(target_scm: torch.Tensor, noise_scm: torch.Tensor
...
+
+    References:
+        [1] R. Serizel, M. Moonen, B. Van Dijk and J. Wouters,
+        "Low-rank Approximation Based Multichannel Wiener Filter Algorithms for
+        Noise Reduction with Application in Cochlear Implants,"
+        in IEEE/ACM Transactions on Audio, Speech, and Language Processing, April 2014.
+    """
+
+    def __init__(self, mu: float, rank: int):
```
mpariente, line 270: Make mu default to 1. 
How about rank? Default of 1, or None?

asteroid/dsp/beamforming.py
```diff
@@ -248,6 +248,91 @@ def compute_beamforming_vector(target_scm: torch.Tensor, noise_scm: torch.Tensor
...
+        e_values, e_vectors = _generalized_eigenvalue_decomposition(
+            target_scm.permute(0, 3, 1, 2),  # bmmf -> bfmm
+            noise_scm.permute(0, 3, 1, 2),  # bmmf -> bfmm
+        )
+
+        #  Prevent negative and infinite eigenvalues
+        eps = torch.finfo(e_values.dtype).eps
+
+        e_values = torch.maximum(e_values, eps * torch.ones(e_values.shape))
+        e_values = torch.minimum(e_values, 1e6 * torch.ones(e_values.shape))
```
mpariente, line 295: Does this work on GPU? It looks like the ones would be created on a different device. 
Anyway, I think this is equivalent to `torch.clamp(e_values, min=eps, max=1e6)` which doesn't have device problems, right?

asteroid/dsp/beamforming.py
```diff
@@ -248,6 +248,91 @@ def compute_beamforming_vector(target_scm: torch.Tensor, noise_scm: torch.Tensor
...
+            e_values[..., self.rank :, :] = 0.0
+
+        #  Compute bf vectors as SDW MWF filter  in eigen space
+        complex_type = e_vectors.dtype
+        inv_ev_plus_mu = torch.linalg.inv(
+            e_values + self.mu * torch.eye(e_values.shape[-1]).expand_as(e_values)
+        ).to(complex_type)
+        bf_vect = (
+            e_vectors @ e_values.to(complex_type) @ inv_ev_plus_mu @ torch.linalg.inv(e_vectors)
+        )
```
mpariente, line 312: I'm unsure, but could this be computed with a `solve` call rather than `inv`? It's usually much faster, and lighter in memory.

asteroid/dsp/beamforming.py
```diff
@@ -439,3 +524,4 @@ def _precision_mapping():
 BeamFormer = Beamformer
 SdwMwfBeamformer = SDWMWFBeamformer
 MvdrBeamformer = RTFMVDRBeamformer
+GevdBeamformer = GEVDBeamformer
```
mpariente, line 527: The aliases above were just not to break backward compat, no need to add it for new classes.

tests/dsp/beamforming_test.py
```diff
@@ -83,6 +84,15 @@ def test_mwf_indices(n_mics, batch_size):
...
 
+@pytest.mark.skipif(not torch_has_complex_support, reason="No complex support ")
+@pytest.mark.parametrize("n_mics", [2, 3, 4])
+@pytest.mark.parametrize("mu", [2.0, 1.0, 0.5])
+@pytest.mark.parametrize("rank", [1, 2, None])
+@pytest.mark.parametrize("batch_size", [1, 2])
+def test_gevd(n_mics, mu, rank, batch_size):
+    _default_beamformer_test(GEVDBeamformer(mu=mu, rank=rank), n_mics=n_mics, batch_size=batch_size)
+
+
```
mpariente, line 95: Looks great!

mpariente (commented):
Thanks for the PR ! 
Some minor comments to be addressed, but otherwise, it looks great :upside_down_face:

asteroid/dsp/beamforming.py
ldelebec: You're right, this don't work on the GPU, because tensor `eps * torch.ones(e_values.shape)` has to be moved on the GPU,
and `torch.clamp()` is more suitable.

asteroid/dsp/beamforming.py
ldelebec: For mu default should be 1.0, agree.
For the rank, 1 is the useful value but None make sense as well, I suggest to set to 1.

asteroid/dsp/beamforming.py
ldelebec: Ok

asteroid/dsp/beamforming.py
ldelebec: `solve` is advised to replace a multiplication on the left by the inverse: `torch.linalg.solve(A, B) == A.inv() @ B`
[see pytorch doc](https://pytorch.org/docs/stable/generated/torch.linalg.inv.html#torch.linalg.inv).
But in that case, I didn't find a way to replace `inv` by `solve`. But I suggest to use `inv` only once, this seems quicker.

asteroid/dsp/beamforming.py
mpariente: 1 it is then.

asteroid/dsp/beamforming.py
```diff
@@ -290,9 +290,7 @@ def compute_beamforming_vector(self, target_scm: torch.Tensor, noise_scm: torch.
 
         #  Prevent negative and infinite eigenvalues
         eps = torch.finfo(e_values.dtype).eps
-
-        e_values = torch.maximum(e_values, eps * torch.ones(e_values.shape))
-        e_values = torch.minimum(e_values, 1e6 * torch.ones(e_values.shape))
+        e_values = torch.clamp(e_values, min=eps, max=1.6)
```
mpariente, line 293:
```suggestion
        e_values = torch.clamp(e_values, min=eps, max=1e6)
```

mpariente: I guess that was the intent ;-)

asteroid/dsp/beamforming.py
mpariente: Perfect

PR #524, opened 2021-07-01
Add library version and name for Hugging Face download stats

osanseviero:
This will enable Asteroid-specific download stats.

FYI @julien-c

### DIFF
--- a/asteroid/utils/hub_utils.py
+++ b/asteroid/utils/hub_utils.py
@@ -43,6 +43,8 @@ def cached_download(filename_or_url):
     Returns:
         str, normalized path to the downloaded (or not) model
     """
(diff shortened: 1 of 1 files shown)


mpariente:
Thanks for the PR, that would be nice indeed! 
The minimum version in the requirements ([here](https://github.com/asteroid-team/asteroid/blob/master/requirements/torchhub.txt)) is 0.0.2, is that enough? 

Could you lint the PR please? I'll merge right after. 
Thanks !

osanseviero:
Done :)

Yes, `cached_download` had the library name/version params since the beginning, so it should work fine with early versions(@LysandreJik please correct me if I'm wrong :) )

LysandreJik:
You are correct!

mpariente:
Thanks to both of you

PR #557, opened 2021-09-24
Allow MixIT to return non-reduced loss over batch.

actuallyaswin:


mpariente:
Thanks for the PR. 

Could you lint the PR and add a test that the expected shape is right please?

PR #561, opened 2021-10-08
Add Readme and pretrained model to fasNet recipe

JorisCos:
I have trained a model with the FasNet recipe and thought it would be good to add results and pretrained model.
@popcornell I have run the experiment with the default recipe but I am a bit confused on how to compare it to the [paper](https://arxiv.org/pdf/1910.14104.pdf).

If I'm not mistaken we are reproducing the last row of table 1 in this configuration. However, I am confused about about the overlap part is it the mean of everything ? 

The result gives an improvement of `9.9 dB` SI-SNR


popcornell:
The average in the paper should be the result on all the test set. The other results are obtained by selecting a subset of the total mixtures with every subset composed of mixtures with < X overlap ratio.

PR #571, opened 2021-11-04
[egs] Fix the issue regarding X-UMX when receiving a mono audio as the input

r-sawata:
Addressed the following issue pointed out by Marius via Slack:
> Marius Miron  12:01 AM
> I think there is a bug in XUMX multidomain loss. whenever you do `2 * i : 2 * i + 2`, you should actually do `n_channel * i : n_channel * i + n_channel`
> otherwise it fails for monaural. what do you think?
> 
> Marius Miron  12:06 AM
> and also (in multidomain) this must be `_Spectrogram(spec_power=spec_power, mono=False)`, otherwise it mixes two sources in mono


mpariente:
Thanks @r-sawata !

@JorisCos would you like to have a look at unit tests and docs build? :-)

PR #574, opened 2021-11-29
Add models for kinect-wsj dataset

ccan1995:
Create a generic model system corresponding to kinect-wsj and create models


mpariente:
Thanks for the PR, I don't know if we will merge this.. It's not very maintainable, but it's Asteroid's fault, not yours ^^

PR #575, opened 2021-12-06
Fix unit-tests

JorisCos:
This fixes unit tests.
The problem was coming from PL version which is  1.5.4 now

### DIFF
--- a/tests/losses/sinkpit_wrapper_test.py
+++ b/tests/losses/sinkpit_wrapper_test.py
@@ -104,7 +104,7 @@ class _TestCallback(pl.callbacks.Callback):
         assert self.epoch * self.n_batch <= step
         assert step <= (self.epoch + 1) * self.n_batch
 
-    def on_train_epoch_end(self, trainer, pl_module, outputs):
+    def on_train_epoch_end(self, trainer, pl_module):
         epoch = trainer.current_epoch
         assert epoch == self.epoch
         assert pl_module.loss_func.beta == self.f(epoch)


jonashaag:
Does this break older PL versions though? We have `>=1.01` in the requirements.

mpariente:
Thanks for fixing the tests !

mpariente:
It probably does break previous versions. 
So let's fix the PL version under the backward incompatible change, make a release, and then we'll upgrade PL and drop previous versions in a new release, does this sound ok?

jonashaag:
Or we could add a manual check to that particular module only. Probably using this code https://github.com/PyTorchLightning/pytorch-lightning/blob/f63222d9663e6af06738c64744522acf7cb20221/pytorch_lightning/utilities/imports.py#L46

mpariente:
Just for reference, the deprecation cycle started [here](https://github.com/PyTorchLightning/pytorch-lightning/pull/7339) in 1.3 and was merged [here](https://github.com/PyTorchLightning/pytorch-lightning/pull/8587) for 1.5. 

I haven't been following PL too much, my bad

mpariente:
> Or we could add a manual check to that particular module only. Probably using this code https://github.com/PyTorchLightning/pytorch-lightning/blob/f63222d9663e6af06738c64744522acf7cb20221/pytorch_lightning/utilities/imports.py#L46

Yes, that's also ok. For now, I'll just fix PL's version.

PR #586, opened 2022-01-11
Fix loading FasNetTAC model

jc5201:
I found a mistake while trying to use FasNetTAC with pretrained model (https://huggingface.co/popcornell/FasNetTAC-paper).
This simple fix works for me.


mpariente:
Thanks a lot for your first contribution ! :pray: 

We'll merge it when CI agrees

Issue #594, opened 2022-02-07
MixIT recipe, pairwise_neg_sisdr


zmolikova:
Hi,
in the MixIT recipe on WHAM, the loss uses `pairwise_neg_sisdr`. It seems to me it should be `multisrc_neg_sisdr` instead. 
https://github.com/asteroid-team/asteroid/blob/fae2f7d1d4eb83da741818a5c375267fe8d98847/egs/wham/MixIT/train.py#L116

Right now, for each mixit partition, each estimate is compared with each reference. This leads to some redundant computation as the permutations are already covered by the mixit partitions. Also, I think it can sometimes even lead to incorrect loss value, because the minimum is taken over each element of the pairwise matrix. That means that for each source, an estimate of different mixit partition can be used (I'm not 100% sure about this).

I can prepare PR but I would be glad if someone confirms this first.

mpariente:
I think there is no better person to confirm this than @etzinis, can you have a look please?

etzinis:
I agree, after the combinations are extracted, the loss should be computed pairwise.

mpariente:
> I agree, after the combinations are extracted, the loss should be computed pairwise.

Do you mean, assuming the sources and estimates are aligned? Or seeking the source/estimate alignment?

etzinis:
I mean, assuming that the combinations are fixed (e.g. the mixture of a subset of sources are assigned to mixture 1 and the rest of the to mixture 2), then the loss for each combination should be computed pairwise given the two target mixtures.

zmolikova:
I think we're just having different terminology. In Asteroid, "pairwise" refers to computing the loss each-to-each, i.e. for 2 sources and 2 estimates, it returns 2x2 matrix. Here, given the fixed combination, we want to compare the estimates only to the assigned sources (mixtures), that is the `multisrc_neg_sisdr` variant.

popcornell:
Yes.  `multisrc_neg_sisdr` is the correct one. Current is incorrect but somehow also works because in the end you care about the min of the losses. Probably we need to add in the mixit loss some dimension checks. 

https://github.com/asteroid-team/asteroid/blob/fae2f7d1d4eb83da741818a5c375267fe8d98847/asteroid/losses/mixit_wrapper.py#L189

mpariente:
@zmolikova, it's confirmed :green_circle: Can you make a PR please?

PR #595, opened 2022-02-08
MixIT loss function

zmolikova:
This solves the issue #594 and adds a simple dimension check to prevent using wrong loss function with MixIT wrapper.


zmolikova:
Yes, I added the test.

mpariente:
Thank you very much, that's perfect !

PR #603, opened 2022-02-18
Fix device in padding for sudormrf  #598  

JorisCos:
Fix #598 . `padding` device wasn't set which caused errors when training on gpus


mpariente:
Looks good, thanks ! But what happened to the tests?

JorisCos:
I think zenodo wasn't responding. It's fine now

PR #612, opened 2022-05-01
Raise PL requirement to 1.3.0

paulfd:
- Remove the deprecated `output` parameter from `on_train_epoch_end`
- Add a simple custom `lr_scheduler_step` to the base model
- Remove the PL upper version constraint

This allows also torchmetrics to follow PL, which caused problems when PL was pinned and not torchmetrics.

### DIFF
--- a/asteroid/engine/schedulers.py
+++ b/asteroid/engine/schedulers.py
@@ -28,7 +28,7 @@ class BaseScheduler(object):
         for param_group in self.optimizer.param_groups:
             param_group["lr"] = lr
 
-    def step(self):
+    def step(self, metrics=None, epoch=None):
         """Update step-wise learning rate before optimizer.step."""
         self.step_num += 1
         lr = self._get_lr()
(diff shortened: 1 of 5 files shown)


paulfd:
Sorry for the linter, this one is simple to fix though. I don't really understand why the docs and coverage break; maybe some supporting package got upgraded through pytorch-lightning newer version or something?

mpariente:
Thanks for the PR. 
The coverage fail is normal because you added some code without tests, but for the docs, it's probably a dependency problem, as you mention, do you mind having a look please ?

mpariente:
Thanks @paulfd for the PR ! 

We'll be able to upgrade to lightning 1.5 and above. But first we'll need to update the recipes for the new Trainer API (strategy, accelerator and devices arguments are different from PL<1.3.0).

mpariente:
I changed the base branch of the PR, because as soon as we merge this, the recipes won't work anymore (because of changes in the Trainer arguments), so I prefer to make a dev branch out of this. 

Thanks again, merging this !

PR #645, opened 2022-10-22
Upgrade deprecated huggingface_hub.cached_download

### LINKED ISSUE #644: Warning cached_download deprecated
Hi, this is my first issue here so I hope I report correctly enough.
Before starting, I would like to give you my congrats for the amazing work done here.

The issue is just a warning, but it may be better to get rid of this stuff before it becomes problematic

### Before reporting a bug:
First, please search [previous issues](https://github.com/asteroid-team/asteroid/issues)
and [the FAQ](https://asteroid-team.github.io/asteroid/faq.html) and be sure this hasn't
been answered elsewhere.

## 🐛 Bug

using `cached_download` raises a warning: 
```
huggingface_hub/file_download.py:588: FutureWarning: `cached_download` is the legacy way to download files from the HF hub, please consider upgrading to `hf_hub_download`
```

### To Reproduce
```
BaseModel.from_pretrained(model_path)
```
where `model_path` is a path that doesn't exist locally.

### Expected behavior

Warning message: 
```
huggingface_hub/file_download.py:588: FutureWarning: `cached_download` is the legacy way to download files from the HF hub, please consider upgrading to `hf_hub_download`
```

#### Package versions

I have 0.5.3 but the same function is still used in the main branch

mattiadg:
Fix #644 
Removes the warning by following the upgrade instructions

### DIFF
--- a/asteroid/utils/hub_utils.py
+++ b/asteroid/utils/hub_utils.py
@@ -65,15 +65,14 @@ def cached_download(filename_or_url):
         else:
             model_id = filename_or_url
             revision = None
-        url = huggingface_hub.hf_hub_url(
-            model_id, filename=huggingface_hub.PYTORCH_WEIGHTS_NAME, revision=revision
-        )
-        return huggingface_hub.cached_download(
-            url,
-            cache_dir=get_cache_dir(),
+        return huggingface_hub.hf_hub_download(
+            repo_id=model_id,
+            filename=huggingface_hub.PYTORCH_WEIGHTS_NAME,
+            revision=revision,
             library_name="asteroid",
             library_version=asteroid_version,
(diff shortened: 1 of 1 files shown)


mpariente:
Thank you very much for the PR. 

Have you tested it ? 

In which version has the `huggingface_hub.hf_hub_download` been introduced ? Won't upgrading to it break asteroid's install for current users ?

mattiadg:
I tried it manually and all the tests for cached_download continue to pass.
I haven't yet checked in which version it was introduced, but it's present
in the version you get from requirements.txt. We can investigate further
though.

mpariente:
Then it's perfect, thanks !

Issue #648, opened 2022-11-04
Cumulative normalization, wrong variance computation


zmolikova:
Hi,
I suspect that the computation of variance in `CumLN` is missing a division by the length of the sequence. That is:

https://github.com/asteroid-team/asteroid/blob/140733bbaafb5940e1b542c7522d4e3b0ba875bb/asteroid/masknn/norms.py#L101

should be
`cum_var = cum_pow_sum / cnt - cum_mean.pow(2)`
as in $var(x) = E[x^2] - E[x]^2$.

mpariente:
You're fully right, thank you ! 

Could you please submit a PR ?

PR #649, opened 2022-11-07
Fix cumulative normalization

zmolikova:
Fixes computation of variance in cumulative normalization (#648).


mpariente:
That's perfect, thanks !

Issue #628, opened 2022-07-02
KeyError: Options in training stage for recipes


joeoct93:
## 🐛 Bug
When I tried to run a recipe (SuDORMRFNET and SuDORMRFImprovedNet), I always end up running into this problem in stage 2 like the following.

Stage 2: Training
Traceback (most recent call last):
  File "/mnt/c/Project/asteroid/egs/librimix/SuDORMRFNet/train.py", line 140, in <module>
    arg_dic, plain_args = parse_args_as_dict(parser, return_plain_args=True)
  File "/home/cchoi1022/cd/envs/asteroid/lib/python3.10/site-packages/asteroid/utils/parser_utils.py", line 135, in parse_args_as_dict
    args_dic["main_args"] = args_dic["optional arguments"]
KeyError: 'optional arguments'

### To Reproduce
cd /asteroid/egs/librimix/SuDORMRFImprovedNet
. ./run.sh

or 
python train.py

### Expected behavior
The training commences

#### Package versions

Run `asteroid-versions` and paste the output here:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 584, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 901, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 792, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (PyYAML 5.3.1 (/usr/lib/python3/dist-packages), Requirement.parse('PyYAML>=5.4'), {'pytorch-lightning'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/cchoi1022/.local/bin/asteroid-versions", line 33, in <module>
    sys.exit(load_entry_point('asteroid', 'console_scripts', 'asteroid-versions')())
  File "/home/cchoi1022/.local/bin/asteroid-versions", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
[...]

PR #657, opened 2023-02-22
[src] Fix argparse key based on Python version (#628)

jbartolewska:


mpariente:
That looks perfect, thanks !

Closes #628

PR #665, opened 2023-05-21
Add a new pre-trained model (X-UMXL)

r-sawata:
We newly trained a model, named X-UMXL, and released its weights on Zenodo. We would like to add it to the list of Asteroid.

Thank you in advance!


mpariente:
Thank you very much for the update ! 

I think you should take a look at HuggingFace Hub instead of Zenodo

r-sawata:
Thank you for your beneficial comment, @mpariente! I'll check it for next time:)

PR #675, opened 2023-09-03
Update overlap_add.py, auto casting device for window

popcornell:
allows to use LambdaOverlapAdd also on GPU without calling before .to on the instantiated module.


asteroid/dsp/overlap_add.py
```diff
@@ -113,7 +113,7 @@ def ola_forward(self, x):
                 frame = _reorder_sources(frame, out[-1], n_src, self.window_size, self.hop_size)
 
             if self.use_window:
-                frame = frame * self.window
+                frame = frame * self.window.to(frame)
```
mpariente, line 116: I didn't know you could pass a tensor directly.

popcornell: Yeah or call frame.device. 
If you pass a tensor it also casts to the same type

Issue #662, opened 2023-04-05
RuntimeError: istft requires a complex-valued input tensor matching the output from stft with return_complex=True.


lucyanddarlin:
I ran `/asteroid-master/egs/musdb18/X-UMX/run.sh`, but got the error: 'RuntimeError: istft requires a complex-valued input tensor matching the output from stft with return_complex=True.'

I try to set the `return_complex=True` in x_umx.py :
```python
 stft_f = torch.stft(
            x,
            n_fft=self.n_fft,
            hop_length=self.n_hop,
            window=self.window,
            center=self.center,
            normalized=False,
            onesided=True,
            pad_mode="reflect",
            return_complex=True,
        )
```
but it didn't work...could someone tell me how to solve it? Thank u so much!

DavidDiazGuerra:
It seems like the newer versions of Pytorch have made some changes to the `torch.stft` and `torch.istft` functions. I've just run through the same issue and I think I could fix it by doing 'x = torch.view_as_complex(x)' just before calling `torch.istft` in the [line that is raising the error](https://github.com/asteroid-team/asteroid/blob/bd3caa3963cf2f8756da17801958073001aaa0f2/asteroid/models/x_umx.py#L401).

Btw, you can also get rid of the deprecation warning you're getting by changing `return_complex` to `True` in the [call to `torch.stft`](https://github.com/asteroid-team/asteroid/blob/bd3caa3963cf2f8756da17801958073001aaa0f2/asteroid/models/x_umx.py#L348) and then doing `stft_f = torch.view_as_real(stft_f)` just after it.

r-sawata:
Thank you so much, @DavidDiazGuerra!

As he said, this was caused by the mismatch between old and new pytorch versions. If the PR #684 will be accepted, then this problem should be resolved.

jrzzzz:
Thank you so much!!! @DavidDiazGuerra

PR #684, opened 2023-11-07
Quick fix for XUMX in torch 2.0

DavidDiazGuerra:
Hello!

I know we should ideally rewrite the XUMX model to work with the new torch's complex type, but it is possible to workaround this by just converting the new type to the old format just after calling `torch.stft` using `torch.view_as_real` and then converting it back to the new type with `torch.view_as_complex` before calling to `torch.istft` (#662) .  This is not the ideal solution but I think it would be useful to allow people to keep using XUMX till the model is fully rewrited.

Best,
David

### DIFF
--- a/asteroid/models/x_umx.py
+++ b/asteroid/models/x_umx.py
@@ -7,13 +7,6 @@ from .base_models import BaseModel
 
 
 class XUMX(BaseModel):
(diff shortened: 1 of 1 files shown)


mpariente:
Thank you very much for the PR ! 

@r-sawata, do you agree ?

PR #688, opened 2024-01-23
_reorder_sources() in OLA compares scaled/windowed frame with un-scaled/un-windowed frame

hendriks73:
Hey there,

`_reorder_sources()` is called with the *current* and the *previous* frames as arguments. The way it had been coded, the previous frame might have been scaled or windowed already. But to get best results, we probably want to compare to an unscaled, un-windowed version. That's what this change fixes.

Cheers.

### DIFF
--- a/asteroid/dsp/overlap_add.py
+++ b/asteroid/dsp/overlap_add.py
@@ -114,7 +114,7 @@ class LambdaOverlapAdd(torch.nn.Module):
 
             out.append(frame)
 
-        # apply windowing/scaling *after* _reorder_sources has been called.
+        # apply windowing/scaling *after* _reorder_sources has been called, inplace.
         for frame in out:
             if self.use_window:
                 frame *= self.window.to(frame)
--- a/asteroid/dsp/overlap_add.py
+++ b/asteroid/dsp/overlap_add.py
@@ -112,11 +112,14 @@ class LambdaOverlapAdd(torch.nn.Module):
                 # we determine best perm based on xcorr with previous sources
                 frame = _reorder_sources(frame, out[-1], n_src, self.window_size, self.hop_size)
 
-            if self.use_window:
-                frame = frame * self.window.to(frame)
-            else:
-                frame = frame / (self.window_size / self.hop_size)
             out.append(frame)
+
+        # apply windowing/scaling *after* _reorder_sources has been called.
+        for frame in out:
+            if self.use_window:
+                frame *= self.window.to(frame)
+            else:
+                frame /= self.window_size / self.hop_size
 
         out = torch.stack(out).reshape(n_chunks, batch * n_src, self.window_size)
         out = out.permute(1, 2, 0)


mpariente:
Hey, 
Thanks for the contribution. 

Look at the diff, because you changed other things (formating), the diff is much less clear. Can you change only the functional part of the code you intend to change please ?

hendriks73:
Looked at the diff—I don't think I have changed the formatting. 🤷‍♂️

asteroid/dsp/overlap_add.py
```diff
@@ -112,11 +112,14 @@ def ola_forward(self, x):
                 # we determine best perm based on xcorr with previous sources
                 frame = _reorder_sources(frame, out[-1], n_src, self.window_size, self.hop_size)
 
+            out.append(frame)
+
+        # apply windowing/scaling *after* _reorder_sources has been called.
+        for frame in out:
             if self.use_window:
-                frame = frame * self.window.to(frame)
+                frame *= self.window.to(frame)
```
mpariente, line 120: This is not a functional change, only cosmetic.

asteroid/dsp/overlap_add.py
```diff
@@ -112,11 +112,14 @@ def ola_forward(self, x):
...
             else:
-                frame = frame / (self.window_size / self.hop_size)
-            out.append(frame)
+                frame /= self.window_size / self.hop_size
```
mpariente, line 122: This too right ?

mpariente (changes requested):

asteroid/dsp/overlap_add.py
hendriks73: If I am not mistaken, `*=` acts in-place, so the `frame` tensor itself gets modified, meaning that the `frame` in the `out` list is also modified.

`frame = frame * self.window.to(frame)` would not accomplish this, would it?

asteroid/dsp/overlap_add.py
hendriks73: See above.

asteroid/dsp/overlap_add.py
mpariente: Indeed, I misread the logic at first, thanks.

asteroid/dsp/overlap_add.py
```diff
@@ -112,11 +112,14 @@ def ola_forward(self, x):
                 # we determine best perm based on xcorr with previous sources
                 frame = _reorder_sources(frame, out[-1], n_src, self.window_size, self.hop_size)
 
+            out.append(frame)
+
+        # apply windowing/scaling *after* _reorder_sources has been called.
```
mpariente, line 117:
```suggestion
        # apply windowing/scaling *after* _reorder_sources has been called, inplace.
```

mpariente:
You are right, it was functional because of the inplace modification I didn't get. 

Merging.

PR #696, opened 2024-04-05
Fixed a bug caused by a type error

Xinyi-Cheng:
This commit addresses a specific `TypeError` encountered in the `LibriMix.__getitem__` method when using the `random.randint` function. The error was triggered by passing a `numpy.float64` type argument to `random.randint`, which expects integer arguments to generate a random integer between two specified integer values. The problematic code snippet was:

```python
start = random.randint(0, row["length"] - self.seg_len)
```

Here, the operation between `row["length"]` and `self.seg_len` could result in a floating-point number, while `random.randint` requires integer inputs. To resolve this issue, the code has been modified to explicitly convert the result of `row["length"] - self.seg_len` to an integer using the `int()` function, as shown below:

```python
start = random.randint(0, int(row["length"] - self.seg_len))
```

This modification ensures that the input to `random.randint` is always an integer, irrespective of the original data types involved, thus preventing the `TypeError`.


mpariente:
Thank you very much for your contribution and clear explanation, it's perfect. 
Please remove all the `.DS_Store` file from the PR, and this is ready to be merged.

Xinyi-Cheng:
Hi, sorry for `.DS_Store` files. I have corrected them in the latest commit.

PR #700, opened 2024-07-19
Fix dns_challenge recipe issue forcing INTERSPEECH 2020 branch

Rodolfo-S:
The current master of the dns_challenge repo does not work with the dns_challenge asteroid recipe. Many scripts have either changed, moved, or have been removed. The version of the dns_challenge repo which was intended to be used was from INTERSPEECH 2020. 

This PR does two things
1. Clones the `interspeech2020/master` branch rather than the `master`, in the `download_data.sh` script
2. Changes the `dns_challenge` recipe directory name to `dns_challenge_INTERSPEECH2020` to indicate that the recipe is meant to use a specific version of the dataset (and dataset repo).


mpariente:
This is perfect, thank you very much !

PR #707, opened 2025-01-08
Fix a missing super-init in the GEVD beamformer class

ovistetom:
Seems like there was a missing `super().__init__()` in the GEV-D beamformer class definition, causing improper inheritance of attributes from the parent `Beamformer` class. 
Notably, this is not the case for the SDW-WMF beamformer class definition.
Caused an error when calling a beamformer instance.

```
>>> import asteroid.dsp.beamforming as bf
>>> beamformer = bf.GEVDBeamformer()
>>> stft_enhanced = beamformer(mix=signal_noisy, target_scm=psd_clean, noise_scm=psd_noise)
AttributeError: 'GEVDBeamformer' object has no attribute '_backward_hooks'
```

### DIFF
--- a/asteroid/dsp/beamforming.py
+++ b/asteroid/dsp/beamforming.py
@@ -268,6 +268,7 @@ class GEVDBeamformer(Beamformer):
     """
 
     def __init__(self, mu: float = 1.0, rank: int = 1):
+        super().__init__()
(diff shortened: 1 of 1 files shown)


mpariente:
Thank you !

PR #719, opened 2026-09-23
[install] Declare requests in install_requires

Kayvan-Zahiri:
A fresh `pip install asteroid` (0.7.0, and main) can't be imported:

```
  File ".../asteroid/utils/hub_utils.py", line 6, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

`hub_utils.py` and `models/zenodo.py` import requests, and `requirements/torchhub.txt` lists it, but `setup.py` doesn't. It used to come in through huggingface_hub, which dropped requests in 1.0 (it uses httpx now), so new installs don't get it.

This adds `requests` to `install_requires`, next to the other torchhub requirements. I checked it by installing this branch into a clean Python 3.12 venv. `import asteroid` works, and `ConvTasNet(n_src=2)` separates a random 8000-sample input into shape `(1, 2, 8000)`.


mpariente:
Thanks a lot !

